PR #2, opened 2020-07-30
Add initial debug enhancements doc

eriknelson:
Working on TODOs, but submitting now for comment


enhancements/debug/README.md
```diff
@@ -0,0 +1,334 @@
...
+so we'll need to go with the controller applied label selector.
+
+
+[3] `Backup` and `Restore` objects have the label `migmigration=<uid>` on it.
+It was very confusing to me that this value was actually the uid from the
+migmigration and not the name of the `MigMigration`, because they are BOTH UUIDs.
+I didn't understand why they did not match until I realized it was the uid field.
+I would propose we 1) make that label very explicit which field it's referencing,
+and 2) add a label that contains the name of the related `MigMigration` so it
+can be queried via label selector.
```
pranavgaikwad, line 234: perhaps, we do not need the `migmigration` label. Looking at the migration plugin code, this label is not used anywhere, it's just informational. So we could just have the `migmigration-name` label and not worry about the UID at all.

enhancements/debug/README.md
```diff
@@ -0,0 +1,334 @@
...
+resources right now are [name and namespace](https://github.com/kubernetes/kubernetes/issues/53459),
+so we'll need to go with the controller applied label selector.
+
+
+[3] `Backup` and `Restore` objects have the label `migmigration=<uid>` on it.
+It was very confusing to me that this value was actually the uid from the
+migmigration and not the name of the `MigMigration`, because they are BOTH UUIDs.
+I didn't understand why they did not match until I realized it was the uid field.
+I would propose we 1) make that label very explicit which field it's referencing,
+and 2) add a label that contains the name of the related `MigMigration` so it
```
pranavgaikwad, line 233: @djwhatle @alaypatel07  I would like to know your thoughts on (1) here. The correlation labels use `Resource.kind` as a key to reference their UID in the value field. Do you think this is a right expectation?

pranavgaikwad: We're not changing the correlation labels at the moment. We only want to introduce new labels

PR #4, opened 2020-09-16
Progress reporting enhancements

### LINKED ISSUE #3: MTC Progress Reporting Enhancements
Create an enhancement doc for Progress Reporting improvements in MTC

pranavgaikwad:
Fixes #3 

Proposing enhancements in MTC progress reporting


djwhatle:
@pranavgaikwad can you add Jeff as a reviewer in the enhancement?

enhancements/progress-reporting/README.md
```diff
@@ -0,0 +1,214 @@
...
+  conditions:
+  - category: Advisory
+    lastTransitionTime: "2020-09-15T14:54:46Z"
+    message: 'Step: 7/33'
+    reason: InitialBackupCreated
+    status: "True"
+    type: Running
+  itenerary: Final
+  observedDigest: 9834d071975562d5e2c2eb855bca6950711ded8a0e45af5307fa56cd0f5ba3c7
+  step: Backup
```
jortel, line 162: The `Reason` in the Running condition communicates the current _step_ to the user.  Currently it is 1-1 with the phase but was designed with the mapping in mind for decoupling.  How about using it instead of a new `step` field?

pranavgaikwad: @jortel Okay. The idea behind introducing this field is to let UI know which broader step in the Migration the current phase belongs to. The word `step` might not be appropriate given the current usage of the word in the context of progress.

enhancements/progress-reporting/README.md
```diff
@@ -0,0 +1,214 @@
...
+see-also:
+  - "N/A" 
+replaces:
+  - "N/A"
+superseded-by:
+  - "N/A"
+---
+
+# MTC Progress Reporting Improvements
+
```
jwmatthews, line 24: @pranavgaikwad I saw the below mentioned deeper in the document, I think this captures the most important goal for this effort, please add this to the first paragraph to reemphasize

"user will have enough information in their hands to understand what exactly is happening in the background"

Thinking something like:

The goal of this document is to highlight user experience issues in the current implementation of progress reporting in MTC, and propose improvements in some of the areas for a better user experience.  

**The driving goal of this work is to ensure the user has enough information in their hands to understand what exactly is happening in the background.**

enhancements/progress-reporting/README.md
```diff
@@ -0,0 +1,220 @@
...
+  step: Backup
+```
+
+#### Phase to Step mapping
+
+Each phase in the migration will belong to some _Step_ based on the the _Itinerary_. The below sections propose a possible mapping for different phases to their correspoding steps.
+
+##### Final Itinerary 
+
+- Prepare: Created, Started, Prepare, EnsureCloudSecretPropagated
```
djwhatle, line 177: @jortel phase to step mapping.

@pranavgaikwad, Jeff had the idea to change the itinerary struct to allow it to be broken up into ordered `Steps`

This would allow us to avoid maintaining a separate mapping. Would also allow `Steps` to contain different phases per itinerary if it makes sense for us.
```
Itinerary
    Step Backup
        Phase InitialBackup
        Phase WaitForBackup
    Step Restore
       ...

File: enhancements/progress-reporting/README.md [hunk 4]
```diff
@@ -0,0 +1,220 @@
...
+
+```yml
+  status:
+    conditions:
+    - category: Advisory
+      lastTransitionTime: "2020-09-16T15:39:44Z"
+      reason: StageBackupCreated
+      status: "True"
+      type: Running
+      progress: 6/9 items backed up
```
djwhatle, line 104: +1 to adding a progress field.

enhancements/progress-reporting/README.md
```diff
@@ -0,0 +1,220 @@
...
+Here are some examples of proposed change:
+
+1. _StageBackupCreated_ Phase:
+
+```yml
+  status:
+    conditions:
+    - category: Advisory
+      lastTransitionTime: "2020-09-16T15:39:44Z"
+      reason: StageBackupCreated
```
djwhatle, line 101: @jortel and I were talking about putting  the new `Step` into this `Running` condition `Reason` slot. mig-ui is supposed to consume all progress info from this condition.

enhancements/progress-reporting/README.md
```diff
@@ -0,0 +1,220 @@
...
+With the progress information relayed in the _MigMigration_ CR, we believe that the user will have enough information in their hands to understand what exactly is happening in the background. We can then simplify the migration by dividing existing phases into broader steps of migration.  
+
+All _Phases_ in a migration can be grouped into following high level steps that abstract out the details from the end user:
+
+* Prepare
+* VolumeBackup
+* Backup
+* VolumeRestore
+* Restore
+* Final
```
djwhatle, line 151: I like the names you've chosen here, but `Final` is unclear to me. Maybe `Cleanup`? Open to other ideas.

enhancements/progress-reporting/README.md
```diff
@@ -0,0 +1,220 @@
...
+##### Stage Itinerary
+
+- Prepare: Created, Started, Prepare, EnsureCloudSecretPropagated
+- VolumeBackup: EnsureStagePodsFromRunning, EnsureStagePodsFromTemplates, EnsureStagePodsFromOrphanedPVCs, StagePodsCreated, AnnotateResources, RestartRestic, ResticRestarted, QuiesceApplications, EnsureQuiesced, EnsureStageBackup, StageBackupCreated, EnsureStageBackupReplicated
+- VolumeRestore: EnsureStageRestore, StageRestoreCreated, EnsureStagePodsDeleted, EnsureStagePodsTerminated
+- Final: EnsureAnnotationsDeleted, EnsureLabelsDeleted, Completed
+
+##### Failed Itinerary
+
+- Final: MigrationFailed, EnsureStagePodsDeleted, EnsureAnnotationsDeleted, DeleteMigrated, EnsureMigratedDeleted, UnQuiesceApplications, Completed
```
djwhatle, line 193: Just FYI, the work @aufi is doing will change the Failed and Cancel itinerary substantially.

enhancements/progress-reporting/README.md
```diff
@@ -0,0 +1,220 @@
...
+All _Phases_ in a migration can be grouped into following high level steps that abstract out the details from the end user:
+
+* Prepare
+* VolumeBackup
+* Backup
+* VolumeRestore
+* Restore
+* Final
+
+Having a fixed number of steps will make it possible for us to implement a pipeline type progress view in the UI.
```
djwhatle, line 153: +1 to thinking about future pipeline support.

enhancements/progress-reporting/README.md
pranavgaikwad: @djwhatle I was unsure about `Final` too. We want something that doesn't just capture the cleanup phases but also Verification and Completed phases

enhancements/progress-reporting/README.md
pranavgaikwad: @djwhatle +1 makes sense

enhancements/progress-reporting/README.md
pranavgaikwad: @djwhatle +1 to updating struct.

djwhatle (changes requested):
@pranavgaikwad I think we can merge this as "implementable" after a few changes:

1. Needs update to intro section re: John's comment
2. Needs to incorporate mention plan to modify the itinerary struct (instead of keeping a totally separate mapping of [phases] -> [step])

I think we could stick with "Final" as the working name for now. It's easy enough to change later, but don't want to block implementation on this.

enhancements/progress-reporting/README.md
```diff
@@ -0,0 +1,220 @@
...
+  startTimestamp: "2020-09-15T14:53:15Z"
+```
+
+The `message` field shows how many steps/phases have been completed so far in the migration. The total number of steps change based on whether it's a stage migration or a full migration. The MTC UI uses this field to derive progress in percentages and depicts it in the form a progress bar:
+
+![MTC Progress Bar](./mtc-progress-bar.png)
+
+### Issues with the current implementation
+
+* In a full migration, there are over 30 steps/phases involved. The migration transitions from one phase to another until completion. All these phases are posted to the Status field of the _MigMigration_ CR as and when encountered. We cannot expect an end user to know the meaning of each of these phases. For instance, one of the migration phases is `EnsureCloudSecretPropagated`. While, from a debugging point of view, knowing that the migration is attempting to propagate cloud secret is an important piece of information, it is questionable how much value it adds to the end user experience.
```
alaypatel07, line 77: @pranavgaikwad I wonder if we should add
1. naming the phases appropriately and 
2. adding description to the phase either in documentation or mig cr

as solution solving the same problem, i.e. improving visibility to user on whats happening

enhancements/progress-reporting/README.md
```diff
@@ -0,0 +1,220 @@
...
+
+* Some phases in migration run for a longer period of time than other phases. For such long running phases, MTC currently does not have a way to inform the user of the progress of the ongoing phase itself. For instance, during an initial Backup, the migration controller enters `EnsureInitialBackup` phase where it creates a Velero Backup object and transitions to `InitialBackupCreated` phase upon successful creation of Backup. It then waits until the underlying Backup object has a Completed / PartiallyFailed / Failed condition. While the backup runs in the background, the end user only sees `InitialBackupCreated` in the progress bar. The problem aggrevates when the Backup contains a huge number of resources. The progress bar is stuck in `InitialBackupCreated` phase with a certain percentage value for a very long period of time. The user is left with no reason to believe that there's actually something happening in the background.
+
+## Proposal
+
+To address the problems discussed in the previous section, we propose to make two modifications:
+
+1. Provide detailed progress information of ongoing migration phase in _MigMigration_ CR
```
alaypatel07, line 85: please see the above [comment](https://github.com/konveyor/enhancements/pull/4/files#r492438380) if we can add this a second goal below this

enhancements/progress-reporting/README.md
alaypatel07: wonder if we should make this field `message` so that for phases that do not have progress to report, we can display an appropriate message. Also, how about formatting the value of this field as: `backup myAwesomeBackup is in progress, 6/9 items completed`. The idea is I would want the name/namespace from this CR so I can look up detailed yaml on source cluster if need be

enhancements/progress-reporting/README.md
```diff
@@ -0,0 +1,220 @@
...
+      lastTransitionTime: "2020-09-16T15:41:41Z"
+      reason: StageRestoreCreated
+      status: "True"
+      type: Running
+      progress: InProgress
+```
+
+Detailed progress of _Restore_ objects is currently unavailable in Velero.  
+
+Apart from the above phases, we propose to incorporate similar progress reporting for other phases of migration, wherever possible. For example, in Stage Pod phases, We can relay the status of the pods to _MigMigration_ CR. This is useful since we have been repeatedly observing issues in Stage pod phases.
```
alaypatel07, line 137: In general, I think the convention to adopt is if a single resource is being created/updated in the phase, we should have name/namespace of that resource so it can be input to other debug steps. If there are multiple resources being updated/created, we should display `x/y updated, 1 failed, 1 in progress` etc. This is in accordance to convention being followed in openshift look here [[1]](https://github.com/openshift/cluster-etcd-operator/pull/290#discussion_r407666816)

enhancements/progress-reporting/README.md
```diff
@@ -0,0 +1,220 @@
...
+
+ 
+### Enhancement 2: Grouping migration phases
+
+With the progress information relayed in the _MigMigration_ CR, we believe that the user will have enough information in their hands to understand what exactly is happening in the background. We can then simplify the migration by dividing existing phases into broader steps of migration.
```
alaypatel07, line 142: IMHO I'm in a grey area if this helpful or misleading user. As a user will I like if there is a smaller number of steps(or phases, using the term to denote actual steps of migration)? Definitely yes. But because the actual process of migration is quite involved, will a user like to have a hierarchy in mig CR, `top level step -> involves multiple lower level phases`? I am not quite sure. It would be great for CI and pipeline view but honestly not too good for CLI. This hierarchy will make it hard(er) for CLI users using jsonpath (and alike feature tools) to query migmigration with `complex` and conditions. Also, we would have to come up with a way of portraying this so that hierarchy is evident. Steps and phases are not good terms for hierarchy, they might lead to confusion, will add the terms here if I can think of more apt ones.
 
In general, I think this is an add-on feature and would only like to see it being implemented if time permits after implementing other efforts in 1.4.0

/cc @jwmatthews @djwhatle

enhancements/progress-reporting/README.md
```diff
@@ -0,0 +1,220 @@
...
+- Backup: PreBackupHooks, EnsureInitialBackup, InitialBackupCreated
+- VolumeBackup: EnsureStagePodsFromRunning, EnsureStagePodsFromTemplates, EnsureStagePodsFromOrphanedPVCs, StagePodsCreated, AnnotateResources, RestartRestic, ResticRestarted, QuiesceApplications, EnsureQuiesced, EnsureStageBackup, StageBackupCreated, EnsureStageBackupReplicated
+- VolumeRestore: EnsureStageRestore, StageRestoreCreated, EnsureStagePodsDeleted, EnsureStagePodsTerminated
+- ResourceRestore: EnsureAnnotationsDeleted, EnsureInitialBackupReplicated, PostBackupHooks, PreRestoreHooks, EnsureFinalRestore, FinalRestoreCreated, EnsureLabelsDeleted, PostRestoreHooks
+- Final: Verification, Completed
+
+##### Stage Itinerary
+
+- Prepare: Created, Started, Prepare, EnsureCloudSecretPropagated
+- VolumeBackup: EnsureStagePodsFromRunning, EnsureStagePodsFromTemplates, EnsureStagePodsFromOrphanedPVCs, StagePodsCreated, AnnotateResources, RestartRestic, ResticRestarted, QuiesceApplications, EnsureQuiesced, EnsureStageBackup, StageBackupCreated, EnsureStageBackupReplicated
```
alaypatel07, line 187: technically this is not accurate

Stage backups now include image stream resources as well. So may rename this to `StageBackup` instead of `VolumeBackup`?

enhancements/progress-reporting/README.md
```diff
@@ -0,0 +1,220 @@
...
+- VolumeRestore: EnsureStageRestore, StageRestoreCreated, EnsureStagePodsDeleted, EnsureStagePodsTerminated
```
alaypatel07, line 188: Same as above

alaypatel07 (commented):
Added a few questions.... +1 to reporting progress

enhancements/progress-reporting/README.md
pranavgaikwad: @alaypatel07 I don't like the idea of adding documentation for each phase as its just not something user likes to do. I want to find the answer to the question of - how can we make it more implicit? How can we make the user know without having to explain them anything?

enhancements/progress-reporting/README.md
pranavgaikwad: @alaypatel07 There won't be hierarchy in the Status field itself. The hierarchy is logical. The status field would look like:

```yml
  conditions:
  - category: Advisory
    lastTransitionTime: "2020-09-15T14:54:46Z"
    message: 'Step: 7/33'
    reason: Backup
    status: "True"
    type: Running
  - category: Required
    lastTransitionTime: "2020-09-15T14:53:15Z"
    message: The migration is ready.
    status: "True"
    type: Ready
  itenerary: Final
  phase: InitialBackupCreated
``` 
The original idea of adding a new field `step` definitely would add confusion. But as @jortel suggested, we can use the `reason` field instead to do that. Do you think this will still confuse the user?

enhancements/progress-reporting/README.md
alaypatel07: `adding documentation` @pranavgaikwad I did not mean to say this description is as exhaustive as documentation. but a little more verbose the one phrase. If you look at pod conditions, it has the following field which I would like to model

```
   message      <string>
     Human-readable message indicating details about last transition.

   reason       <string>
     Unique, one-word, CamelCase reason for the condition's last transition.
```

In our case the reason is the phase and message can be human readable string

enhancements/progress-reporting/README.md
djwhatle: @alaypatel07 want to understand what you are saying.

Right now we use message for `Step 7/33`, which is a phase progress indicator:

```
  - category: Advisory
    lastTransitionTime: "2020-09-15T14:54:46Z"
    message: 'Step: 7/33'
    reason: InitialBackupCreated
    status: "True"
    type: Running
```

Are you proposing we should wipe out the phase based progress info from the status (since we'd be using the field it exists in now for something else)?

enhancements/progress-reporting/README.md
alaypatel07: @pranavgaikwad in the status field you proposed, I wonder if users would be able to immediately connect Step 7/33 corresponds to `InitialBackupCreated`.  In the message field, we are calling it Step 7 of 33 but later when displaying the name we are calling it phase

enhancements/progress-reporting/README.md
djwhatle: @pranavgaikwad in case you weren't aware, I've been looking at adding `detailed phase descriptions` to controller since before this sprint. I'm not sure how it will fit in with all of these proposed changes but I'm suspecting it will need to adapt. https://github.com/konveyor/mig-controller/pull/656

I'm also planning to do a round of updating the phase names to be more indicative of what is _currently_ being worked on.

djwhatle: So to clarify, the user wouldn't need to look up detailed phase info in docs, it would be reflected on Mig status.

enhancements/progress-reporting/README.md
pranavgaikwad: @alaypatel07 this problem is there with the current phase implementation too. Step and Phase are kind of interchangeable in their meanings.

enhancements/progress-reporting/README.md
pranavgaikwad: @djwhatle @alaypatel07 I get the point now. Having a `message` field with detailed description, I think makes the Enhancement 2 (grouping phases) invalid. I wouldn't call it an enhancement if we can already convey the same information through the message field.

enhancements/progress-reporting/README.md
alaypatel07: @pranavgaikwad If they are interchangeable I wonder if we can replace `message: 'Step: 7/33'` with `message: 'Phase: 7/33'`?

enhancements/progress-reporting/README.md
alaypatel07: @djwhatle multiple options:

options 1:

```
  - category: Advisory
    lastTransitionTime: "2020-09-15T14:54:46Z"
    message: 'phase: 7/33, backup <backup-name> in progress, 6/9 completed'
    reason: InitialBackupCreated
    status: "True"
    type: Running
```

2.
```
  - category: Advisory
    lastTransitionTime: "2020-09-15T14:54:46Z"
    message: 'Phase: 7/33, backup <backup-name> in progress'
    reason: InitialBackupCreated
    status: "True"
    type: Running
     veleroBackupProgress: <struct from velero API>
``` 
and similar for restore. 

I like option one because we can make it generic for other phases, `message: Phase: x/y; a/z resources updated`

alaypatel07:
```
status:
   phaseStatuses:
   - name: "FOO"
     startTime: "today 4pm"
     lastUpdateTime: "today 6pm"
     progress: "8/10 completed, working on 9th"
```

I wonder if adding a new status field `phaseStatuses` instead of reusing the conditions makes sense. This would serve multiple objectives:
1. a good place to update progress
2. a place to record the start time and update time for each phase status. This will help in bubbling up phases that take longer than expected. Since we have update time, what we are reporting in progress is tied to a time, so technically we are as accurate as we can be
3. ordered list of phaseStatuses will allow us to retain the information even after migrations for retrospecting. This will make the migmigration yaml larger but all the information is useful in debugging

djwhatle:
@pranavgaikwad  would you prefer to leave this open as a living doc until we have finalized changes, or should we merge as `implementable`?

pranavgaikwad:
> @pranavgaikwad would you prefer to leave this open as a living doc until we have finalized changes, or should we merge as `implementable`?

I am just addressing Alay's feedback of having a structured progress field instead of strings. I would like to get that in probably tomorrow and then the PR would be good to merge.

djwhatle:
Merging as `implementable`, @pranavgaikwad is planning to make another pass to reconcile this with whatever drift occurred during the actual implementation process.

PR #6, opened 2020-09-23
add enhancement for logging improvement

alaypatel07:


enhancements/logging/controller-logging.md
```diff
@@ -0,0 +1,137 @@
...
+## Design Details
+
+1. Use the `log.V(x)` to get the verbosity of the log.
+2. Replace all `log.Trace(err)` to include some context of:
+    1. what the controller was trying to do when error occured
+    2. the resource and NamespacedName on which error occured (if applicable)
+    3. if this is source cluster or destination (if applicable)
+   Example: `log.Trace(fmt.Errorf("unable to update foo: fooNS/fooName in %s cluster", clusterURL))`
+3. For any status condition change, generate a kube event,
+   for all logs that are `warning`, generate a kube event
```
alaypatel07, line 101: I am still working on coming up with a generic method of classifying if a kube event is important or not but mostly will go based on gut feel for now. Didnt want this to stop me for raising the PR

enhancements/logging/controller-logging.md
```diff
@@ -0,0 +1,137 @@
+---
+title: cam-non-admin-design
```
jwmatthews, line 2: is this the right title?

alaypatel07: fixed!

enhancements/logging/controller-logging.md
```diff
@@ -0,0 +1,139 @@
...
+ [[2]](https://github.com/kubernetes/kubernetes/blob/0fd10997dfee0f9ec2c413004bf93d310d76c84b/pkg/controller/deployment/deployment_controller.go#L572)
+and OpenShift
+[[3]](https://github.com/openshift/cluster-etcd-operator/blob/0806334d716f294ebd22b0cf38be4ebce1e64b30/pkg/operator/clustermembercontroller/clustermembercontroller.go#L92) 
+[[4]](https://github.com/openshift/cluster-etcd-operator/blob/0806334d716f294ebd22b0cf38be4ebce1e64b30/pkg/operator/clustermembercontroller/clustermembercontroller.go#L170). 
+
+All our controller as essentially driving towards the desired state. Because
+of this, most of the time a controller checks the state, there is no action required
+(steady state). When there is no action required, the controller should still log 
+important info but at level v(4) or more verbose. When controller does have to take
+an "ensure" action, it should log at v(2). By default, the pod starts with  v(2), but
```
djwhatle, line 72: @alaypatel07 could we include a section with definition of the log levels you mention?

We previously talked about this set, not sure if it still holds:

```
0 – Errors and warnings only
2 – Normal information
4 – Debugging information
6 – API- debugging information (request / response)
8 – Body API debugging information
```

enhancements/logging/controller-logging.md
```diff
@@ -0,0 +1,139 @@
...
+kubernetes[[1]](https://github.com/kubernetes/kubernetes/blob/0fd10997dfee0f9ec2c413004bf93d310d76c84b/pkg/controller/deployment/deployment_controller.go#L581)
+ [[2]](https://github.com/kubernetes/kubernetes/blob/0fd10997dfee0f9ec2c413004bf93d310d76c84b/pkg/controller/deployment/deployment_controller.go#L572)
+and OpenShift
+[[3]](https://github.com/openshift/cluster-etcd-operator/blob/0806334d716f294ebd22b0cf38be4ebce1e64b30/pkg/operator/clustermembercontroller/clustermembercontroller.go#L92) 
+[[4]](https://github.com/openshift/cluster-etcd-operator/blob/0806334d716f294ebd22b0cf38be4ebce1e64b30/pkg/operator/clustermembercontroller/clustermembercontroller.go#L170). 
+
+All our controller as essentially driving towards the desired state. Because
+of this, most of the time a controller checks the state, there is no action required
+(steady state). When there is no action required, the controller should still log 
+important info but at level v(4) or more verbose. When controller does have to take
```
djwhatle, line 71: @alaypatel07 Can you provide some examples of what constitutes `important info` that also requires no action on behalf of the controller?

enhancements/logging/controller-logging.md
```diff
@@ -0,0 +1,139 @@
...
+what actions were performed by mig-controller against the source and destination clusters.
+
+### Goals
+
+- Improving debugging experience by adding more logging information to mig-controller
+- Alleviate pains and improve turnaround time in chasing bugs
+
+### Non-Goals
+
+- Aggregate logs from other pods in MTC architecture for better visibility
```
djwhatle, line 57: I'm looking at this separately in case anyone is wondering.

enhancements/logging/controller-logging.md
djwhatle, line 72: @alaypatel07 the way you summarized this for me before:

- `v2` - used to log places where mig-controller changes the state of the world
- `v4` - used to log places where mig-controller is issuing API requests without changing world state

Is this in line with what you're thinking? If not, what differs?

enhancements/logging/controller-logging.md
```diff
@@ -0,0 +1,139 @@
...
+### Implementation Details/Notes/Constraints [optional]
+
+### Risks and Mitigations
+
+NA
+
+## Design Details
+
+1. Use the `log.V(x)` to get the verbosity of the log.
+1. Replace all `log.Trace(err)` to include some context of:
```
djwhatle, line 95: Shouldn't a log trace include this ns/name info already? Or is this not what you've observed in practice? I thought we would be logging additional places, not replacing log.Trace which I think does a good job for logging critical errors.

enhancements/logging/controller-logging.md
```diff
@@ -0,0 +1,139 @@
...
+    1. what the controller was trying to do when error occured
+    1. the resource and NamespacedName on which error occured (if applicable)
+    1. if this is source cluster or destination (if applicable)
+   Example: `log.Trace(fmt.Errorf("unable to update foo: fooNS/fooName in %s cluster", clusterURL))`
+1. Go through all ensure actions and log after making create/update/delete call 
+   to kube apiserver. If the action is important enough, generate an event
+1. For any status condition change, generate a kube event,
+1. For all logs that are `warning`, generate a kube event
```
djwhatle, line 103: What are some examples of things in mig-controller where we would generate a warning but not an error?

djwhatle (commented):
Thanks for putting this together @alaypatel07 , added some questions

enhancements/logging/controller-logging.md
sseago: I imagine any time we add a Warning condition on a mig CR we'd also want a warning level log.

enhancements/logging/controller-logging.md
djwhatle: `log.Trace()` is used for stack tracing when a critical error occurs. Would we continue using `log.Trace()` for this purpose exclusively?

enhancements/logging/controller-logging.md
djwhatle: That makes sense to me

enhancements/logging/controller-logging.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Implementation Details/Notes/Constraints [optional]
+
+### Risks and Mitigations
+
+NA
+
+## Design Details
+
+1. Use the `log.V(x)` to get the verbosity of the log.
```
djwhatle, line 94: Based on the way this is worded, `log.V(x)` just returns `v4` or `v2` or whatever the set verbosity level is? Or is this how we actually write logs of a particular level.

enhancements/logging/controller-logging.md
```diff
@@ -0,0 +1,139 @@
...
+[ ] update logs in migplan-controller
+[ ] update logs in migstorage-controller
+[ ] update logs in migcluster-controller
+[ ] update logs in mighook-controller
+[ ] update logs in mig-analytic-controller
+[ ] update logs in migmigration
+
+## Drawbacks
+
+This will make logs little more verbose and devs will have to find effective tools
```
djwhatle, line 130: It sounds like this will make the logs a lot more verbose rather than a little, but I agree it could be very useful for piecing together the sequence of events that happened in a user env.

enhancements/logging/controller-logging.md
alaypatel07: sure sounds like good info to add

alaypatel07: > v2 - used to log places where mig-controller changes the state of the world
v4 - used to log places where mig-controller is issuing API requests without changing world state

v4 - used where the controller finds out information that is not critical to break the reconcile function and requeue but the information would be important during debugging

enhancements/logging/controller-logging.md
alaypatel07: @djwhatle I have linked v4 information from kube and openshift controllers. I havent gotten around to finding appropriate examples in mig-controller but will work on that soon

enhancements/logging/controller-logging.md
alaypatel07: yeah so we would use log.V(2) to logs at v2 and log.V(4) to log at v4 just use a generic x for logging level

enhancements/logging/controller-logging.md
alaypatel07: I don't recall seeing ns/name information in log.Trace(), hence I propose to add it exclusively. 

> log.Trace() is used for stack tracing when a critical error occurs. Would we continue using log.Trace() for this purpose exclusively?

I think we will continue to use log.Trace() calls for this purpose exclusively, I am looking for way to pretty-print the stack trace

enhancements/logging/controller-logging.md
alaypatel07: yes, we'll be careful in not getting it over verbose and hiding some verbosity under v4

enhancements/logging/controller-logging.md
alaypatel07: @djwhatle sorry for the lag, but I have an example where I think v4 log will be very helpful:

```
There are phases in the controller where we list all the resources and the label/annotate[1] 
them. Since kube api does not support batch update, we need to iterate over the list for update 
operation. For all such places, I would like to see a v4 long entry that include the resource
being update, number of items being updated, and the total amount it took for all of the list
to be updated. 
```

1. https://github.com/konveyor/mig-controller/blob/5c15c6d933dd7d7bfc5d9bcdfa2c91500b00b348/pkg/controller/migmigration/annotation.go#L368

enhancements/logging/controller-logging.md
```diff
@@ -0,0 +1,139 @@
...
+Major milestones in the life cycle of a proposal should be tracked in `Implementation
+History`.
+
+[ ] update logs in migmigration-controller
+[ ] update logs in migplan-controller
+[ ] update logs in migstorage-controller
+[ ] update logs in migcluster-controller
+[ ] update logs in mighook-controller
+[ ] update logs in mig-analytic-controller
+[ ] update logs in migmigration
```
alaypatel07, line 126: need to be removed

djwhatle:
@alaypatel07 would you prefer to leave this open as a living doc until we have finalized changes, or should we merge as `implementable`?

djwhatle:
Alay agreed that we should scrap this doc due to heavy differences in implementation after feedback received.

PR #9, opened 2020-11-03
Breakdown long running phase to add requeue logic

JaydipGabani:
Proposing to add requeue logic in `AnnotateResource` phase to report progress during the long-running annotations.

### DIFF
--- a/enhancements/annotation-requeue/README.md
+++ b/enhancements/annotation-requeue/README.md
@@ -51,7 +51,7 @@ We propose the following modifications in the function responsible for annotatin
 
 ```
 total := len(allResource)
-annotation := 0
+now := time.Now()
 for i, ns := range allResource() {
     retrieve single resource
     .
@@ -66,9 +66,8 @@ for i, ns := range allResource() {
     .
     .
     .
-    annotation ++
-    if total >= 5 && annotation == total / 5 {
-        t.Requeue = PollReQ
+    if time.Now().Sub(now) == time.Second * 3 {
+        t.Requeue = FastReQ
         return nil
     }
     t.Progress = []string{fmt.Sprintf("%v/%v Namespace labeled", i, total)}    
@@ -78,7 +77,7 @@ for i, ns := range allResource() {
 Changes in task.go file in `AnnotateResource` switch case:
 
 ```
-if t.Requeue != PollReQ {
+if t.Requeue != FastReQ {
     if err = t.next(); err != nil {
         return liberr.Wrap(err)
     }
--- a/enhancements/annotation-requeue/README.md
+++ b/enhancements/annotation-requeue/README.md
@@ -51,7 +51,6 @@ We propose the following modifications in the function responsible for annotatin
 
 ```
 total := len(allResource)
-now := time.Now()
 for i, ns := range allResource() {
     retrieve single resource
     .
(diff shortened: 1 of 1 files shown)


djwhatle:
Hey @JaydipGabani , thanks for the PR. I have a question about how many annotations we'll make before exiting and requeuing. I see this:

```
   if total >= 5 && annotation == total / 5 {
        t.Requeue = PollReQ
        return nil
    }
```

**Question**: _Is this saying that we exit for PollReQ when:_
1. We've annotated at least 5 resources
2. AND we have annotated 1/5 of the total resources

My thought is that instead of letting the size of the migration determine how long each reconcile is, we should instead choose a reasonable fixed number of resources to annotate before exiting the reconcile.

We can use known burst/QPS figures in our client to think about how long each reconcile should last. Our QPS (queries per second) is 100, and our burst is 1000 in our current client config.

From [migcluster_types.go](https://github.com/konveyor/mig-controller/blob/5733c65e32c71d712024d6d7b54cdc6dc519b755/pkg/apis/migration/v1alpha1/migcluster_types.go#L193-L199)

```
	restConfig := &rest.Config{
		Host:            m.Spec.URL,
		BearerToken:     string(secret.Data[SaToken]),
		TLSClientConfig: tlsClientConfig,
		Burst:           1000,
		QPS:             100,
	}
```

---

Another idea would be to record a timestamp at the start of the annotateResources method and then gracefully exit the method to requeue after some fixed amount of time has passed, maybe ~10 seconds or so.

---

Last piece of feedback: I think we probably want to use `FastReQ` instead of `PollReQ` here to avoid dead time between reconciles.

JaydipGabani:
@djwhatle  the idea was to exit after each 1/5th of the resources are annotated here
```
if total >= 5 && annotation == total / 5 {
        t.Requeue = PollReQ
        return nil
    }
```
But instead, I like the idea of logging a timestamp and then exit after a fixed amount of time.

enhancements/annotation-requeue/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+    if _, exist := <resource>.GetLabels()[IncludedInStageBackupLabel]; exist {
+			continue
+	}
+
+    set annotations
+    .
+    .
+    .
+    if time.Now().Sub(now) == time.Second * 3 {
```
djwhatle, line 69: This looks good to me, want to call out one more thing.

There is a tradeoff here between how quick the user gets updates vs. how much time we have to spend doing re-validation of the MigMigration after each requeue at the beginning of reconciles.

3 seconds works as a good placeholder for now. When we implement, I think we'll be able to find a good balance by experimentation. I think we could run a few tests to see how much slower the overall annotation process runs with a few different exit times. We could compare performance between exiting after 3 seconds, ~10 seconds, and perhaps ~30 seconds.

djwhatle (approved):

enhancements/annotation-requeue/README.md
djwhatle: I am also concerned that on a sluggish cluster, we may not do any work in 3 seconds.

JaydipGabani: +1 on deciding the optimal time for reporting, I was concerned about the same.

PR #10, opened 2021-01-08
Add direct migration filesystem ownership enhancement

eriknelson:
I wanted to capture our understanding of the filesystem ownership problem and proposed solutions in a more permanent location than a Bugzilla comment, so submitting here for review for posterity.

### DIFF
--- a/enhancements/direct-migration-filesystem-ownership/README.md
+++ b/enhancements/direct-migration-filesystem-ownership/README.md
@@ -1,0 +1,120 @@
+# Filesystem ownership reconciliation with direct volume migrations
+
+While working the [direct volume migration] feature, it became apparent that
+is can be complex to manage the owner and group permissions of volumes between
+the source, replication pod (rsync pod), and target destination pods. This is
+especially true when considering the many different provisioners that potentially
+behave differently with regard to security contexts (some respect the fsGroup,
+while some unusual ones do not).
+
+## Related issues:
+
+Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1908214
+
+## Observed issue
+
+With the original implementation of DVM, the rsync transfer pod was run
+as a privilieged pod, meaning it executed as root. By launching the pod with
+these properties, it mutated the source volumes's filesystem permission bits
+to **no longer match those of the original** (in this case, obviously it would
+not be able to access root permissioned files). This caused the source workload
+to crashloop with permission errors while trying to access its volume's data.
+
+## BZ Root Cause Analysis
+
+In order to root cause this bug the following test set up was used:
+
+1. A dummy pod was setup to replace the rsync client pod. This pod just sleeps
+2. On gp2 even with a dummy pod, the application broke. This strongly suggests that mounting gp2 volumes twice is the problem, nothing to do with rsync.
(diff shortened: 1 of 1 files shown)


enhancements/direct-migration-filesystem-ownership/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+The group, in this case, is the same as the application user, hence the application pod does not break.)
+
+## Proposed solution
+
+Before starting the rsync client pod, we check if the pvc’s are attached to
+any pods in the namespace.
+
+* If any pvc is attached to one or more pods:
+  * Attached to one pod, get its fsGroup and use the same fsGroup for the rsync
```
shawn-hurley, line 95: We should supplemental groups here as well IMO

enhancements/direct-migration-filesystem-ownership/README.md
```diff
@@ -0,0 +1,120 @@
...
+  for the rsync client pod while explicitly setting the node to match the picked pod’s.
+  * If pvc’s are attached to pods and none of them have a scc with fsgroup,
+  don't use any fsgroup. This will adopt the namespaces default security context.
+
+* If not attached to any pod:
+  * Run without explicit fsgroup, adopting default security context.
+
+Along with the fsgroup, we will also have to work with supplemental group ids,
+which is the same concept as fsgroup but for shared storages like NFS and glusterfs.
+
```
shawn-hurley, line 109: There is more here than @alaypatel07 is working on in regards to the destination pod and how to make sure that it is run as none root and keep the same permissions suck that the application pod when migrated will still be able to read/write.

I think we should add that info here as well.

enhancements/direct-migration-filesystem-ownership/README.md
```diff
@@ -0,0 +1,120 @@
...
+## Edge cases
```
shawn-hurley, line 110: @alaypatel07 are there potential issues with the destination rysnc server being root with the escape hatch and the application pod having new GIDs/UID

shawn-hurley (changes requested):
We have found more edge cases with the destination that we should continue to add to this doc. All the source cluster bits look right to me, though.

enhancements/direct-migration-filesystem-ownership/README.md
eriknelson: Do we need the selinux context value as well?

eriknelson: Actually, have we audited the rest of the `PodSecurityContext` to determine all of the fields we actually care about? (https://docs.openshift.com/container-platform/4.6/rest_api/objects/index.html#podsecuritycontext-core-v1). I'd expect that we only care about those that will grant us the correct filesystem privileges.

I'm seeing the `SELinuxOptions` option as a struct itself, I'm assuming we would need to clone this full object?

enhancements/direct-migration-filesystem-ownership/README.md
eriknelson: @alaypatel07 Can you elaborate on the above?

enhancements/direct-migration-filesystem-ownership/README.md
```diff
@@ -0,0 +1,144 @@
...
+
+I'm going to use `securityContext` in the following as a shorthand for
+`securityContext.{fsGroup,supplementalGroups,seLinuxOptions}`
+
+* If any pvc is attached to one or more pods:
+  * Attached to one pod, get its `securityContext` and use the same `securityContext` for the rsync
+  client pods and explicitly set the node to match the current pod that the
+  volume is attached to.
+  * Attached to more than one pod, raise a warning because this could lead to
+  indeterminate behavior. Pick any of the pods’s fsGroup at random and use it
```
shawn-hurley, line 109: nit: s/fsGroup/securityContext/ I think

shawn-hurley (approved):

PR #12, opened 2021-01-19
Intelligent PV resizing

pranavgaikwad:


enhancements/pv-resizing/README.md
```diff
@@ -0,0 +1,181 @@
...
+
+1. Gathering disk capacity and actual usage data from the source cluster.
+2. Computing resulting disk capacity in the target cluster
+3. Creating the volume in the target cluster
+
+Steps 1 & 2 will be common for both direct and indirect migrations, while Step 3 will need to be implemented separately for both.
+
+#### Gathering disk capacity and actual usage data from the source cluster.
+
+To determine the usage information, the _MigPlan_ controller will run `df` command in one of the Restic Pods which is mounting the source volume. The information for each volume will be stored in the _MigPlan_ status. When the controller fails to gather this information, it will raise a _Warning_ condition to inform the user that the volume capacity cannot be adjusted automatically. However, this will not be considered a _fatal_ failure and migrations can still continue despite this missing information.
```
shawn-hurley, line 57: Is there a world where restic pods are not deployed?

enhancements/pv-resizing/README.md
```diff
@@ -0,0 +1,181 @@
...
+
+- Actual used volume size + threshold value in percentage
+- Actual provisioned volume capacity
+- Requested volume capacity in the Persistent Volume Claim
+
+The "threshold" value specified above will be set to 3% by default and will be configureable through an exposed variable in _MigrationController_ or _MigPlan_ CR. In the Mig UI, the new adjusted size will be displayed in the MigPlan Wizard. An editeable field (disabled by default) for adjusted volume size will allow user to either confirm the size or propose a new size by enabling the editeable field with the help of a check box. The Mig UI will run validation on the field such that the user cannot propose a smaller value than the original requested size of the volume. Additionally, similar validation will be implemented in the _MigPlan_ controller to allow similar experience for CLI users. The _MigMigration_ controller will add an annotation on each Persistent Volume Claim to indicate the actual capacity of the to-be-restored volume.
```
shawn-hurley, line 67: Don't know how I feel about updating the PVC on the source cluster, is there any other option that we have? 

Is the problem that we use velero to restore and we have no way to pass extra information into our velero plugin?

enhancements/pv-resizing/README.md
```diff
@@ -0,0 +1,181 @@
...
+  srcVolumes:
+  - name: pvc-27c0bd90-5b3a-11eb-9ab3-0a06fd6a26f8    <---
+    actualCapacity: 4Gi                               <---
+    requestedCapacity: 3Gi                            <---
+    usage: 98%                                        <---
+    adjustedCapacity: 4200Mi                          <---
+    reason: The source volume is almost full.         <---
+```
+
+Currently, I am assuming the new changes will be fully backward-compatible.
```
shawn-hurley, line 173: What happens on restore in the plugin if the annotation is not found?

shawn-hurley:
I think the annotation approach is something that we have used often, I am wondering if we have a second to think of an alternative.

Another question is there something that we can do for specifically DVM only as a carrot for moving folks forward, is it a strict requirement to  use with indirect?

If we remove that requirement would it be simplier or more complicated?

pranavgaikwad:
> I think the annotation approach is something that we have used often, I am wondering if we have a second to think of an alternative.
> 
> Another question is there something that we can do for specifically DVM only as a carrot for moving folks forward, is it a strict requirement to use with indirect?
> 
> If we remove that requirement would it be simplier or more complicated?

There is no hard requirement to support indirect. However, if we have to let users give a chance to confirm the adjusted value, we must have to leverage MigPlan capabilities. Otherwise, there is no way for the user to actually intervene. And if we choose to use MigPlan to achieve that, indirect can naturally make use of the added annotation.

enhancements/pv-resizing/README.md
pranavgaikwad: The expected behaviour will be that the plugin takes no action.

enhancements/pv-resizing/README.md
pranavgaikwad: That has been a default pattern in our controller to pass information to Velero plugins post backup

enhancements/pv-resizing/README.md
pranavgaikwad: @shawn-hurley not right now. could be in future.

enhancements/pv-resizing/README.md
shawn-hurley: So, I don't know how all the objects look as well as you so if this is not even close to possible just let me know. 

I was thinking that during the Mig Plan selection we would show the user:
1. How big the PVC claim is
2. The actual usage
3. Editable field that says the size we will make it.

A user will set on the MigPlan the volumes and the sizes.

When mig controller runs it will, for each PV/PVC in the mig plan it will create the DVM for that PVC and set the updated value as a spec field to DVM.

This allows us to have DVM have this functionality not by fiat but by spec value, allowing folks to use the same mechanism if they ever use DVM by itself.

The reason that I think this pattern is "better" is that
1. For a given DVM we always know what it is attempting
2. On rollback and migrate we don't have to edit a PVC
3. If the PVC is ruled by an operator, it won't conflict w/ the operator
4.  A carrot to move folks to direct migration

enhancements/pv-resizing/README.md
alaypatel07: I wonder if we Migplan controller is the right place to do this. There are other choices available like MigAnalytics controller or creating a new CR for this I would like to see arguments/thoughts put into those choices.

alaypatel07: > When the controller fails to gather this information, it will raise a _Warning_ condition to inform the user that the volume capacity cannot be adjusted automatically. However, this will not be considered a _fatal_ failure and migrations can still continue despite this missing information. 

+1 like this to be the goal, this feature is an enhancement, failure of this process in the feature is non-fatal.

enhancements/pv-resizing/README.md
```diff
@@ -0,0 +1,181 @@
...
+
+#### Gathering disk capacity and actual usage data from the source cluster.
+
+To determine the usage information, the _MigPlan_ controller will run `df` command in one of the Restic Pods which is mounting the source volume. The information for each volume will be stored in the _MigPlan_ status. When the controller fails to gather this information, it will raise a _Warning_ condition to inform the user that the volume capacity cannot be adjusted automatically. However, this will not be considered a _fatal_ failure and migrations can still continue despite this missing information. 
+
+#### Computing resulting disk capacity in the target cluster
+
+Once the capacity and the usage information is successfully collected in the previous step, the _MigPlan_ controller will compute the new adjusted capacity. The controller will use the maximum of the following values as the resulting capacity for the to-be-restored volume:
+
+- Actual used volume size + threshold value in percentage
```
alaypatel07, line 63: is the threshold a hardcoded value or would it be coming in from mig-operator CR?

enhancements/pv-resizing/README.md
alaypatel07: Also, I think I would like to see thoughts laid out in this document as to why using Restic vs deploying a separate pod?

IMO, Restic is an easy way to get the feature going but not might now be a good eventual goal. Reasons:

Restic pods are "owned" by the mig-operator and if we want to do this right, the controller responsible for this feature should create its dependency pods to manage it well. The obvious benefit of the responsible controller owning these pods is that any future improvements we make to the way we watch/reconcile remote resources will be covered here and we won't have to make an exception for Restic. Any future tests we plan to add will have one less external dependency for this controller. In short, we are trading containing dependencies of the controller with development time and it would be a good exercise to think, talk and document those.

enhancements/pv-resizing/README.md
alaypatel07: it is documented below, ignore this I guess

enhancements/pv-resizing/README.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Motivation
+
+When working with one of our upstream tooling - [pvc-migrate](https://github.com/konveyor/pvc-migrate) in a production environment, we ran into a scenario where the users had resized several GlusterFS volumes using Gluster CLI. The volumes were originally created by dynamic provisioner through PVCs. The size of the volumes were no longer equal to their respective requested sizes in the claims. This shows that it is not uncommon to find examples of volumes in production where their actual capacities and apparent capacities differ. 
+
+### Goals
+
+1. Accurately determine and adjust the size of the volume to be restored such that the data copy operations don't result into failure caused due to insufficient disk space
+2. Design a common solution so that it can be leveraged by both direct and indirect migration workflows.
```
alaypatel07, line 38: Given that there are questions raised below on implementing this with indirect migration, I am wondering if this is the right goal?

I am curious if we can approach this as, we will solve it for direct migrations and then if we do it right we will have a way of piping it into velero plugin as well.

Also, if we remove this as a goal, it means this is not a deliverable for 1.4.1, which is an added advantage given the short timeframe of 1.4.1 dev cycle

enhancements/pv-resizing/README.md
alaypatel07, line 67: The usage of a PVC is not a static quantity. Given that a migplan can stale, i.e. it can be created long before it is attempted to migrate, there is a chance that the size of the PVC has changed. Is this a problem worth considering?

enhancements/pv-resizing/README.md
```diff
@@ -0,0 +1,181 @@
...
+
+The Velero Restore plugin will look for the annotation on each claim and create a new PVC accordingly.
+
+---
+
+### User Stories [optional]
+
+#### Story 1
+
+As a user, I want to be able to see the *proposed* destination volume size (along with the reasons) in an editable field on MigPlan wizard at the time of selecting Storage Class for each volume so that I can confirm the automatically adjusted size or suggest my own value.
```
alaypatel07, line 85: again, what if the proposed change is not accurate during the time of migration but accurate during the time of migplan creation

enhancements/pv-resizing/README.md
```diff
@@ -0,0 +1,181 @@
...
+
+### Implementation Details/Notes/Constraints [optional]
+
+#### Using Restic pod to run df command
+
+Since the application pod which is using the volume may or may not have `df` command installed, Restic pods are a good choice to run the `df` command. However, this makes the solution completely rely on the fact that Restic will always be available in the source cluster. Another approach would be to launch a temporary pod to gather this information. Some storage classes also report volume usage information as metrics. We can leverage this information for known storage classes.
+
+#### Size conversions 
+
+The requested capacity in a Persistent Volume Claim either follows standard SI notation or their power-2 equivalents as specified [here](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/scheduling/resources.md#resource-quantities). The `df` command on the other hand only follows the SI notation. It is important for the controller to accurately convert all sizes from-to / to-from Kubernetes and `df` units. The conversion code should be absolutely bug free and tested for all posssible sizes.
```
alaypatel07, line 95: +1

enhancements/pv-resizing/README.md
pranavgaikwad: I don't agree with this. If this is a problem, then we should never allow migrations on a stale migplan anyway. Because PVs may change and if MigPlan gets reconciled after new PVs are added to source namespace, this will still work.

alaypatel07 (changes requested):
Added few questions to think about, it might be that those questions dont change the implementation in which case we should document why? and if they do change anything that we need to update the doc. PTAL, thanks!

enhancements/pv-resizing/README.md
djwhatle: For what it's worth, the MigPlan does one last reconcile when the migration starts running. The user doesn't really have a chance to respond and make new selections on any changes since they last went through the wizard.

```
		{Name: Created, Step: StepPrepare},
		{Name: Started, Step: StepPrepare},
		{Name: StartRefresh, Step: StepPrepare},
		{Name: WaitForRefresh, Step: StepPrepare},
```

enhancements/pv-resizing/README.md
djwhatle: I think the last minute refresh might catch this. See other comment https://github.com/konveyor/enhancements/pull/12#discussion_r568900029

djwhatle: @shawn-hurley are you suggesting that we only include PV resizing as a feature in DVM?

enhancements/pv-resizing/README.md
djwhatle: Great idea `The [...] code should be absolutely bug free` +1

enhancements/pv-resizing/README.md
shawn-hurley: what happens on rollback for 1 and what happens on rollback for 2? 

What happens if the PVC is owned by an operator for 1 and for 2?

the additional surface area to me looks like a huge problem to debug when something goes wrong. Especially if the value is erased on the PVC because of a rollback. I don't see it as a win that the info we need is not on our objects but rather on a user's objects. We can control when our objects are updated/changed not a user's objects. 

because of our step approach if I delete the annotation on the PVC, then when back up runs it won't get the correct value. I am just seeing ways that this can fail that we would/should guard against that make implementing option 2 give us a much larger tail of support/reliability concerns IMO.

enhancements/pv-resizing/README.md
alaypatel07: I guess the part where I am confused it, assume the user has adjusted the size to 5 gigs from the UI, and the last-minute refresh suggests that the PV has grown to 7gig, how do we expect the system to react? 

The possible options are:

1. we make the migplan unready, the user will have to go through the wizard again and find this
2. we keep the migplan as is and don't migrate the PV, complete the migmigraiton with a warning. I don't see how this is forward progress from where we are now.
3. try the migration and fail. This is what we do now and so no forward progress here as well.

Curious to hear more thoughts on this

enhancements/pv-resizing/README.md
pranavgaikwad: @alaypatel07 @shawn-hurley Deploying a new Pod comes with costs, it will add significant time to migration. Observing the state of the fresh pods is an added task to think about. We need to make sure the new pod is in desired state, if it's not, we need to inform the user. Same thing goes with Restic pod as well. But it doesn't add the cost of creation. In general, the idea of depending on mig-operator for these pods is attractive for two reasons: 
1. operator keeps reconciling and making sure pods are always launched on all nodes even when new nodes get added etc
2. startup time is saved. 

If not Restic pods, we should depend on our own daemonset launched by the operator. If the pods are deemed unhealthy, we degrade the service gracefully.

enhancements/pv-resizing/README.md
pranavgaikwad: @alaypatel07 Option 1.

enhancements/pv-resizing/README.md
pranavgaikwad: @shawn-hurley that's a good point. Since annotations is the default way we use to instruct Velero plugins, I don't see any other option for indirect migration. DVM on the other hand can just read from MigPlan. In the first run, we can probably not focus on indirect migration.

enhancements/pv-resizing/README.md
shawn-hurley: I think this may be ok for now, but I don't like this as a UX that we always use IMO. 

I would like to explore ways that a user, during a migration, can mitigate these concerns to move the migration forward rather than starting over.

enhancements/pv-resizing/README.md
shawn-hurley: During a velero restore, do we have anyway of telling the user that X was from an later version of the backup plugins? 

This might be a concern for OADP in the future, but if there is an annotation, and the older version of the plugin can not handle, we probably want to alert the user right?

shawn-hurley (approved):
For 1.4.1 I think the functionality as described will be ok with the change to how DVM gets the information.

I still personally would like to see us not implement this for indirect migration @jwmatthews if you would like to weigh in on the customer use cases and if that makes sense?

enhancements/pv-resizing/README.md
alaypatel07: sure, both the patterns have pros and cons, just wanted to point which one is more forward-looking. Either way works for me, just wanted to see it documented in alternatives that were rejected so we can revisit it later.

enhancements/pv-resizing/README.md
alaypatel07: @pranavgaikwad works with me. Can you please capture the appropriate details for implementing this in the doc?

> @alaypatel07 Option 1.

alaypatel07 (approved):
I have two requests for additions to the document, apart from that LGTM

While it is ideal to do it for both direct and indirect, but if forced to choose in the interest of dev time it makes sense to proritise DVM first and revisit velero later when the time is right.

djwhatle:
@shawn-hurley @jwmatthews I have 2 concerns with only implementing PV resizing for _direct migration_ and skipping _indirect migration_ implementation:

1. My impression is that  _Indirect migration_ is still the only viable option for some users. Are we at the point where we are generally scaling down investment in indirect migration?

2. UI will need to be restructured, as right now user doesn't get to select "direct/indirect" until late in the wizard, and the user has already made PV selections by that point.

enhancements/pv-resizing/README.md
djwhatle: Seems likely that for a large category of apps (any app that writes to the disk) the PV will _always_ change in size by some small amount (kilobytes maybe) between plan creation and execution. I don't think this should cause the migration to lock up. 

I think it might make more sense to extrapolate whatever user selections were made (e.g. user wanted 5% headroom in disk space) to the new detected size of PVs.

pranavgaikwad: @djwhatle I agree that the PV will always change. I like the idea of using a default headroom in case something like that happends. Also note that, although PV changes, how many times will it actually happen that the capacity of PV changed? 
For instance, 10 Gigs of originally provisioned PV, although apps read/write to it, how often do you think the capacity of 10Gigs will be changed to..say...11 Gigs? It might happen that one branch in the logic (checking whether PV is full) might get triggered more often than others.

enhancements/pv-resizing/README.md
```diff
@@ -0,0 +1,181 @@
...
+approvers:
+  - "@sseago"
+creation-date: 2021-01-19
+last-updated: 2021-01-19
+status: provisional
+---
+
+# Intelligent PV resizing
+
+When using filesytem copy mode in MTC to migrate Persistent Volume data, MTC uses Velero's default capabilities to restore the associated Persistent Volume Claim in the target cluster. The capacity of the new Persistent Volume in the target cluster is determined by the requested capacity in its Persistent Volume Claim. While this approach works well in most scenarios, there are some cases in which the capacity of Persistent Volume in the source cluster may actually be different than the requested capacity in its Claim. In such cases, the filesystem copy mode in MTC will fail if the capacity of the volume is more than requested. This document proposes a common solution to address this problem in MTC's direct and indirect migration workflows both.
```
jwmatthews, line 18: Do we want to remove we are addressing this in 'indirect'?

enhancements/pv-resizing/README.md
```diff
@@ -0,0 +1,181 @@
...
...in the target cluster for both direct and indirect migrations. As stated in the opening section, the actual capacity of a provisioned volume isn't necessarily same as the requested capacity in its claim. Both direct and indirect migration workflows in MTC don't take into account such a scenario. For instance, GlusterFS storage in OpenShift 3.11 cannot handle requested capacities in the granularity of Megabytes or below. As a result, when a user requests a volume of 1100Mi, the provisioned volume will actually be of 2Gi capacity. However, when Velero restores such a volume in a cluster with AWS EBS backed storage, the resulting volume will match its requested size i.e. 1100Mi. This will make a migration fail if the volume in the source contains more than 1100Mi of data. Additionally, when volume usage in the source cluster is close to 100%, it is likely that the same data may require more than 100% of capacity in the target cluster due to differences in underlying storage classes. MTC doesn't handle this scenario either. This document proposes a solution to accurately determine the resulting volume size in a way that it can be leveraged by the direct and the indirect migration both.
```
jwmatthews, line 29: +1 great example

enhancements/pv-resizing/README.md
jwmatthews: +1,  I think it make sense to remove solving this for indirect.  Focus on Direct first.

enhancements/pv-resizing/README.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Proposal
+
+The proposed solution will work in three steps:
+
+1. Gathering disk capacity and actual usage data from the source cluster.
+2. Computing resulting disk capacity in the target cluster
+3. Creating the volume in the target cluster
+
+Steps 1 & 2 will be common for both direct and indirect migrations, while Step 3 will need to be implemented separately for both.
```
jwmatthews, line 53: Remove indirect?

enhancements/pv-resizing/README.md
jwmatthews: I see value in limiting scope, remove indirect support and only implement for Direct.   If we see customers _need_ indirect we can come back to revisit, I don't see enough to suggest indirect will be needed for our customers.

jwmatthews:
> @shawn-hurley @jwmatthews I have 2 concerns with only implementing PV resizing for _direct migration_ and skipping _indirect migration_ implementation:
> 
> 1. My impression is that  _Indirect migration_ is still the only viable option for some users. Are we at the point where we are generally scaling down investment in indirect migration?

Yes, desire is to invest efforts in Direct migration primarily.  We want to keep indirect functional, yet we are not trying to maintain feature parity.  It is OK to make Direct migration 'better' while not doing same for indirect.

I *think* we will see almost all of our users can and will use Direct.  I may be wrong :) 

> 2. UI will need to be restructured, as right now user doesn't get to select "direct/indirect" until late in the wizard, and the user has already made PV selections by that point.

enhancements/pv-resizing/README.md
```diff
@@ -0,0 +1,191 @@
...
+
+The requested capacity in a Persistent Volume Claim either follows standard SI notation or their power-2 equivalents as specified [here](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/scheduling/resources.md#resource-quantities). The `df` command on the other hand only follows the SI notation. It is important for the controller to accurately convert all sizes from-to / to-from Kubernetes and `df` units. The conversion code should be absolutely bug free and tested for all posssible sizes.  
+
+#### Handling updates to volumes post creation of MigPlan
+
+It is possible that the volumes may be resized after the migplan was created. Since _MigPlan_ is reconciled once before running a migration, it is possible to add a warning condition on _MigPlan_ whenever a change is detected in the volume capacity after user made the selection. 
+
+The change can be handled in two ways: 
+1. MigPlan can be marked "Not Ready" and we can force user to make the selection again
+2. Migration can proceed by choosing a maximum of the following values:
```
pranavgaikwad, line 102: @djwhatle @alaypatel07 Let me know what you guys think about Option 2

PR #14, opened 2021-01-26
Configurable namespace + Finalizers

shawn-hurley:


enhancements/configurable-namespace/configurable-namespace.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Motivation
+
+As we look forward to some features that we may want to implement in the future, it is becomes apparent that the namespace that the operator and controller are deployed to should be different. One of the issues is that we are not currently using finalizers on resources, even when we need them to stick around. The DVM resource for instance can be deleted, throwing the mig migration into an state where it is waiting for a resource that has been deleted. What we would want is a finalizer to prevent the deletion so that the mig controller can gracefully handle this. The primary reason for not using finalizers, is because removing the namespace that has Mig* resources is not possible when the controller and those resources are in the same namespace. In addition to allowing for the finalizers to be used by the controller, it also will start to enable Migration controller watching a configurable namespace. This will help in our future plans of non admin work.
```
djwhatle, line 34: I think the attitude on the team is generally pretty anti-finalizer after using them and ripping them back out previously. 

From the past experience this team has had, finalizers seem to be  _very difficult_ to use correctly and often seem to prevent namespace deletion and frustrate even our own devs. I think @alaypatel07 was the most recent to try to implement finalizers in a more nuanced way that would avoid problems, but I'm not sure what came of that.

To handle the situation where a DVM is deleted and the migmigration is waiting for it to complete, it seems a **really simple solution** would be to just detect that the DVM resource we're looking for is `NotFound` and fail the migration.  

Are there other use-cases besides the one you mentioned that need finalizers?

djwhatle: @jortel @dymurray @sseago not sure if you guys want to weigh in on finalizers, I don't remember all of the details.

djwhatle: @shawn-hurley are we thinking our future setup would always have the namespace for mig-controller and mig-operator seprated? To implement finalizers as a tool in our codebase it seems we'd always have to keep the namespaces separate

enhancements/configurable-namespace/configurable-namespace.md
```diff
@@ -0,0 +1,105 @@
...
+
+This pattern is used by many controllers in openshift already, most notably the core operators for api-server, DNS, and networking.
+
+### Goals
+
+List the specific goals of the proposal. How will we know that this has succeeded?
+
+1. mig controller can use finalizers to prevent the deletion of resources, and gurantee saftey without the user being unable to delete the namespace that contains the resources they create.
```
djwhatle, line 42: I have to ask, is it really desirable to prevent deletion of resources? A user isn't going to accidentally type out an oc delete command, if they do such a thing they had to think about it. They _will_ probably be confused about why the resource isn't deleting like they asked it to, if history is a good indicator.

enhancements/configurable-namespace/configurable-namespace.md
```diff
@@ -0,0 +1,105 @@
...
+1. mig controller cr can take a configurable namespace, that the migration controller will use to watch for MigMigration resources.
+
+### Non-Goals
+
+1. Allow for the mig controller to watch a subset of namespaces
+1. Allow for the mig controller to safetly watch the entire cluster and tenant the operations
+1. Determining all thge saftefy requirements for finalizers and implementation.
```
djwhatle, line 49: Spellcheck :)

enhancements/configurable-namespace/configurable-namespace.md
```diff
@@ -0,0 +1,105 @@
...
+// Create a new Cmd to provide shared dependencies and start components
+log.Info("setting up manager")
+namespace := os.Getenv("CONTROLLER_WATCH_NAMESPACE")
+mgr, err := manager.New(cfg, manager.Options{Namespace:namespace})
+```
+
+1. MigrationController will have a new field `controller_watch_namespace` which be defaulted to `openshift-migration` to keep backwards compatability.
+1. The migration UI will have to be deployed in the same namespace as the migration controller and have the "CONTROLLER_WATCH_NAMESPACE" added to it's permissions.
+1. If there is an active MigMigration, we should protect the MigPlan with a finalizer.
+1. Must Gather must be updated to handle the configurablity of the controller to find the Mig* and Direct* resources.
```
djwhatle, line 66: So I guess here we'd just traverse all MigrationController resources and gather from all of those namespaces. Would Velero also be installed in each of these namespaces?

enhancements/configurable-namespace/configurable-namespace.md
```diff
@@ -0,0 +1,105 @@
...
+#### Story 1
+As a cluster admin I would like to configure the location of the openshift-migration resources to be created/edited and watched.
+
+#### Story 2
+As a cluster admin I would like the resources that should not be deletable be protected with a finalizer so they can be removed saftly.
+
+### Implementation Details/Notes/Constraints
+
+The use of finalizers is to make sure that the controller is alerted to the deletion action and can take the correct actions. In the case of a Plan deletion, this means the clean up of the MigMigration resources and when that is deleted any resources on the source or destination cluster. 
+The namespace seperation is also a step in the direction for non-admin because today the controller only watches a single namespace name and this moves us into the direction that this can be configurable or there could be multiple controllers watching multiple namespaces. This is not something to be solved in this proposal but does start us down the direction that this is possible. I also think it is imprtant that the operator and the controller are in the same namespace.
```
djwhatle, line 85: @shawn-hurley to make sure I understand:

 - mig-operator and mig-controller (possibly multiple instances) will be in same NS
 - our CRs will be in separate namespace

djwhatle (commented):
This enhancement seems to cover
 - Configurability of namespaces
 - Addition of Finalizers to mig-controller/mig-operator

It's unclear to me whether the configuration of namespaces refers to where the controller will be, or where the MTC CRs will be. 

I changed the PR title as such to make it clearer.

eriknelson:
@shawn-hurley What do you think are the next steps with this one? Do we want to open a story in Jira to track this? Is it still of enough interest to polish and merge?

shawn-hurley:
I think that this idea is still very necessary but in regards to descoping conversations, I think this should be closed and rolled into that conversation.

PR #16, opened 2021-02-04
Dynamic pipeline proposal

pranavgaikwad:
This PR:
- updates Progress Reporting doc to match the implementation
- adds a new proposal for dynamically built pipeline

### DIFF
--- a/enhancements/dynamic-pipeline/README.md
+++ b/enhancements/dynamic-pipeline/README.md
@@ -1,0 +1,167 @@
+---
+title: dynamic-pipeline
+authors:
+  - "@pranavgaikwad"
+reviewers:
+  - "@djwhatle"
+  - "@alpatel"
+approvers:
+  - "@ernelson"
+  - "@djwhatle"
+  - "@alpatel"
+creation-date: 2021-02-02
+last-updated: 2021-02-04
+status: implementable
+see-also:
+  - "../progress-reporting/README.md"
+replaces:
+  - "N/A"
+superseded-by:
+  - "N/A"
+---
+
+# Dynamic Pipeline
+
+Progress reporting [enhancement](../progress-reporting/README.md) introduced the idea of a Migration Pipeline in MTC. While the pipeline greatly improves observability and the overall experience, there is still room for improvement in the pipeline implementation. This enhancement document highlights some of the areas of improvement in the pipeline. 
+
+## Release Signoff Checklist
+
+- [x] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] User-facing documentation is created
+
+## Open Questions
+
+> Does flag evaluation depend on outcome of previous migration phases? 
+
+## Summary
+
+This enhancement addresses following problems in the pipeline:
+
+1. Some steps have confusing names. For instance, StageBackup actually involves PV and Image backup, it runs even during a final migration. Since `Stage` has a special meaning in the context of migrations, the terminology of stage steps creates confusion. 
+
+2. Some steps in the pipeline are marked `Skipped` when they are not executed which may add to confusion.
+
+## Motivation
+
+### Goals
+
+1. Propose new names for some of the pipeline steps.
+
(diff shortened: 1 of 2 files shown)


djwhatle (approved):
Thanks for bringing this up to date with what was implemented @pranavgaikwad. Merging.

enhancements/dynamic-pipeline/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+2. Do not include steps in the pipeline which will be eventually skipped.
+
+### Non-Goals [omitted]
+
+## Proposal
+
+### Goal 1: Propose new names for some of the pipeline steps
+
+`StageBackup` can be renamed to `DataBackup` or `StateBackup` as Stage backup is responsible for migrating both images and persistent volume data. `StageRestore` can be renamed to `DataRestore` or `StateRestore`. This will make sure that we dont have conflicting meaning to the term `stage`.
```
djwhatle, line 60: Would "PVBackup" be more suitable? Seems like all of our state is contained within PVs. I defer to @vconzola on naming

enhancements/dynamic-pipeline/README.md
```diff
@@ -0,0 +1,167 @@
...
+```
+
+The above logic can be updated to:
+
+```
+1. Read current itinerary, step and phase from MigMigration
+
+2. Iterate over all phases in the current itinerary 
+
+  - Apply allFlags() and anyFlags() on the phase to determine whether this phase is actually part of migration
```
djwhatle, line 84: I think there may be 1 or 2 special cases where we'll have to do some modification for pre-evaluation of flags to work. 

Example:
```
	if phase.all&HasStagePods != 0 && !t.Owner.Status.HasCondition(StagePodsCreated) {
		return false, nil
	}
```

I am currently not understanding why we need the additional check for `stagePodsCreated` condition here, but this seems like something that would only get added to the migration after running for a bit.

https://github.com/konveyor/mig-controller/blob/f33cc9e2b4b1dfbe2ca1e86b5ab00a8a932f60e7/pkg/controller/migmigration/task.go#L1160-L1162

djwhatle: `StagePodsCreated` condition is added here: https://github.com/konveyor/mig-controller/blob/f33cc9e2b4b1dfbe2ca1e86b5ab00a8a932f60e7/pkg/controller/migmigration/stage.go#L255-L265

And is used here:
```
		{Name: StagePodsCreated, Step: StepStageBackup, all: HasStagePods},
		{Name: EnsureStagePodsDeleted, Step: StepCleanup, all: HasStagePods},
		{Name: EnsureStagePodsTerminated, Step: StepCleanup, all: HasStagePods},
```
It seems like it would be easy to modify the logic such that the check for the the `StagePodsCreated` condition could be moved inside of these phases so that it's not tied up in the flags.

enhancements/dynamic-pipeline/README.md
```diff
@@ -0,0 +1,167 @@
...
+  if prevItinerary != currentItinerary {
+    for _, phase := range t.Itinerary.Phases {
+	  	[...]
+    }
+  }
+  ...
+}
+```
+
+Above will make sure that the pipeline is built _only once_ per itinerary even though the function is called in every reconcile.
```
djwhatle, line 125: +1, slick

djwhatle (approved):
@pranavgaikwad plan sounds good to me. I think we will need to fix the flag `HasStagePods` but other than that seems like it will work well.

enhancements/dynamic-pipeline/README.md
pranavgaikwad: @djwhatle I investigated what happens in this scenario. When a Step is added to the pipeline due to some phase, we never evaluate flags for any other phases present in the same step because we need at most one phase to determine whether a step is part of a migration. 

In this specific scenario, `StageBackup` step is added to the pipeline whenever `EnsureStagePodsFromTemplates` is determined to be part of a migration. Therefore, there never arises a case where we will ever evaluate flags for all subsequent phases including `StagePodsCreated`. I would therefore consider this a non-issue.

PR #17, opened 2021-02-11
adding why privileged SCC and configuring it for DVM

alaypatel07:


enhancements/direct-migration-filesystem-ownership/README.md
```diff
@@ -172,3 +172,39 @@ We'd like to build an "escape hatch" into the operator
 that allows admins to **explicitly opt in** to running the transfer pod as
 privileged. Proper caution should be taken in this case by the admin to ensure
 unauthorized access to the transfer pod is not possible while the transfer is underway.
+
+## Why would a user require Privileged SCC
```
alaypatel07, line 176: I did not have a solid use case where we need the privileged SCC. IIRC we had this as a backdoor in case it is needed. 

Given the above, I need a little help in forming this section considering this will likely be the source of downstream documentation

alaypatel07: cc @shawn-hurley @eriknelson ^

apinnick: +1  The use cases are not clear. 

I am thinking of putting this into the "Troubleshooting" section because it does not look like an admin would change the rysnc SCC to privileged unless either DVM failed or the file permission bits were changed by DVM. Is my understanding correct?

eriknelson: I'm concerned about documenting this in any user facing docs due to the security implications of escalating to privilege; I don't want someone hitting an error and attempting to get around it by using this flag when it's not absolutely needed. It should only be used in a specific circumstance where it's absolutely required, and with the cluster (or at the very least, access to this namespace/pod) completely locked down to non-privileged users.

I would consider this to be an engineering backdoor at this time, but I'm open to arguments to the contrary.

apinnick: @eriknelson  
If Alay has not found a solid use case, I think we should be very careful about documenting a procedure that creates a security vulnerability. This information is not necessary for normal MTC use cases. If DVM problems occur, I think it is better to wait for an escalation. I agree that there is a risk someone might try running rsync as privileged to solve an unrelated DVM problem.

alaypatel07:
@eriknelson @shawn-hurley @apinnick 

We discussed that this does not need to be documented upstream or downstream in order to avoid any user confusion. We will revisit this whenever we find the need to document this with valid use cases in the future. Closing the PR with that. Thanks

PR #19, opened 2021-03-08
DVM Retry Mechanism

pranavgaikwad:
This PR adds proposal for a retry mechanism in Direct Volume Migration. 

Handy link to see this PR if you prefer rich text instead of a crappy diff interface: https://github.com/pranavgaikwad/enhancements/tree/dvmJob/enhancements/dvm-retry-mechanism

### DIFF
--- a/enhancements/dvm-retry-mechanism/README.md
+++ b/enhancements/dvm-retry-mechanism/README.md
@@ -11,7 +11,7 @@ approvers:
   - "@djwhatle"
   - "@shawn-hurley"
 creation-date: 2021-03-08
-last-updated: 2021-03-11
+last-updated: 2021-03-15
 status: implementable
 see-also:
   - "N/A"  
@@ -54,17 +54,17 @@ DVM controller creates an OpenShift Route in the target cluster to establish dir
 
 ## Proposal
 
-As of MTC 1.4.2, DVM controller creates a Pod to run Rsync. It creates as many Pods as there are volumes in the source cluster. Rsync command is injected as startup command in the Pod. When Rsync fails, the Rsync client container in the Pod never restarts as the _restartPolicy_ is set to _Never_ by default. It is possible to simply change the _restartPolicy_ of the Rsync client Pod to _OnFailure_ which will cause Kubelet to restart the Rsync container whenever failure takes place. However, it is not desired for our use case because Kube uses exponential backoff to restart the container. In case of multiple failures, the wait time between two consecutive retries can go upto 5 minutes. The time between consecutive launchs of a container cannot be configured. For our use-case, we want the pods to be re-launched immediately after a failure. 
+The proposed retry mechanism will be implemented in DVM controller. DVM CR will create a Rsync Pod per PVC as previously done. However, instead of marking the migration as failed upon first failure, it will re-create the Rsync Pods until the specified number of retries is met or the most recently created Rsync Pod exits with 0 return code. The number of failures will be kept track of in the DVM CR's _Status_.
 
-[Kubernetes Jobs](https://kubernetes.io/docs/concepts/workloads/controllers/job/) are ideal for our use case and provide the retry feature we need. A Job uses `.spec.template` to specify a Pod template. The field `.spec.backoffLimit` allows setting maximum number of retries on the Job. The _Status_ fields `.status.succeeded` and `.status.failed` tell the number of successful and failed attempts respectively. The Rsync client Pods can be replaced with Jobs to leverage their inherent retry mechanism. The controller can watch the status of the Job instead of watching individual Pods.
-
-Throughout this document, a single _Rsync Pod_ launched by the Job will be referred to as an _Rsync Attempt_, while the collection of all attempts will be referred to as _Rsync Operation_, interchangeably. 
+This approach will work on all supported versions of MTC. It also leaves room for enhancements in the retry logic itself. See [alternative approaches](#alternatives) considered and the problems with each of them. 
 
 ### Implementation Details/Notes/Constraints
 
-#### Creating Job 
+Throughout this document, a single _Rsync Pod_ launched by the DVM controller will be referred to as an _Rsync Attempt_, while the collection of all attempts will be referred to as _Rsync Operation_, interchangeably.
 
-In _VolumeMigration_ itinerary of the DVM controller, _[CreateRsyncClientPods](https://github.com/konveyor/mig-controller/blob/523d3a8ef70e89296535d2d0b94173669e92007c/pkg/controller/directvolumemigration/task.go#L126)_ phase is responsible for creating Rsync Pod. This phase will be renamed to _CreateRsyncClientJobs_. In that, new jobs will be created for each _Persistent Volume_. The same Pod template will be used in the Job's `.spec.template`. By default, `.spec.backOffLimit` will be set to `10`. Optionally, `backOffLimit` value will also be available for overriding through _DirectVolumeMigration_ CR:
+#### Specifying retries
+
+The number of retries Rsync performs will be configured through _DirectVolumeMigration_ CR. A sane default value will be chosen in case the user doesn't specify their own retry value:
 
 ```golang
 type DirectVolumeMigrationSpec struct {
@@ -73,7 +73,9 @@ type DirectVolumeMigrationSpec struct {
 }
 ```
 
-Each Job will add unique labels on the Pods it creates to make sure the Pods associated with a Job can be found easily.
+#### Creating Pods
+
+In _VolumeMigration_ itinerary of the DVM controller, _[CreateRsyncClientPods](https://github.com/konveyor/mig-controller/blob/523d3a8ef70e89296535d2d0b94173669e92007c/pkg/controller/directvolumemigration/task.go#L126)_ phase is responsible for creating Rsync Pod. This phase will create pods as usual and add a unique label on each pod indicating which PVC it belongs to. 
 
 #### Direct Volume Migration Progress
 
@@ -86,17 +88,19 @@ type DirectVolumeMigrationProgressSpec struct {
 }
 ```
 
-A new reference to the Rsync Job will be added in the spec:
+DVMP will be modified to watch all attempts of Rsync instead of watching just one Pod. A new field `PodSelector` will be introduced to input unique label added by the DVM controller on the Pods of each Rsync attempt.
 
 ```golang
 type DirectVolumeMigrationProgressSpec struct {
 	[...]
-	// JobRef is a reference to the Rsync Job
-	JobRef     *kapi.ObjectReference `json:"jobRef,omitempty"`
+	// PodSelector selector to find all attempts of Rsync
+	PodSelector     map[string]string `json:"podSelector,omitempty"`
+	// BackOffLimit retries limit set by DVM
+	BackOffLimit    int               `json:"backOffLimit,omitempty"`
 }
 ```
 
-A new struct `RsyncPodStatus` will be introduced which will encapsulate the _Status_ information of an individual Rsync attempt. `RsyncPodStatuses` in the DVMP _Status_ will hold the historical information of all Rsync attempts. The `RsyncPodStatus` of the most recent Rsync attempt will also be included as top level field in DVMP status to avoid repeated searches through the list of attempts. Field `JobElapsedTime` will record the elapsed time of the ongoing Job. `Succeded`, `Failed` and `Completed` fields in the _Status_ will be introduced to hold the derived status of the Rsync operation. This simplifies logic wherever DVMP is being used in other places.
+A new struct `RsyncPodStatus` will be introduced which will encapsulate the _Status_ information of an individual Rsync attempt. `RsyncPodStatuses` in the DVMP _Status_ will hold the historical information of all Rsync attempts. The `RsyncPodStatus` of the most recent Rsync attempt will also be included as top level field in DVMP status to avoid repeated searches through the list of attempts. Field `RsyncElapsedTime` will record the elapsed time of the ongoing Rsync operation as a whole (including all retries). `Succeded`, `Failed` and `Completed` fields in the _Status_ will be introduced to hold the derived status of the Rsync operation. `BackOffLimit` will be used to verify whether the derived status of the Rsync operation. This simplifies logic wherever DVMP is being used in other places.
 
 ```golang
 type DirectVolumeMigrationProgressStatus struct {
@@ -105,15 +109,15 @@ type DirectVolumeMigrationProgressStatus struct {
 	RsyncPodStatus
 	// RsyncPodStatuses is a list of status information of all Rsync attempts
 	RsyncPodStatuses            []RsyncPodStatus `json:"rsyncPodStatuses,omitempty"`
-	// JobElapsedTime is the total time spent in Rsync operation including all retries
-	JobElapsedTime              *metav1.Duration `json:"jobElapsedTime,omitempty"`
+	// RsyncElapsedTime is the total time spent in Rsync operation including all retries
+	RsyncElapsedTime            *metav1.Duration `json:"rsyncElapsedTime,omitempty"`
 	// TotalProgressPercentage is the sum of progress percentage of all Rsync attempts
 	TotalProgressPercentage     int              `json:"totalProgressPercentage,omitempty"`
 	// Succeded whether the Rsync operation succeded
 	Succeded                    bool             `json:"succeded,omitempty"`
 	// Failed whether the Rsync operation failed
 	Failed                      bool             `json:"failed,omitempty"`
-	// Completed whether the Job has finished - failed or succeded
+	// Completed whether the Rsync operation has finished - failed or succeded
 	Completed                   bool             `json:"completed,omitempty"`
 }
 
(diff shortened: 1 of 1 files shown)


enhancements/dvm-retry-mechanism/README.md
```diff
@@ -0,0 +1,203 @@
...
+
+[Kubernetes Jobs](https://kubernetes.io/docs/concepts/workloads/controllers/job/) are ideal for our use case and provide the retry feature we need. A Job uses `.spec.template` to specify a Pod template. The field `.spec.backoffLimit` allows setting maximum number of retries on the Job. The _Status_ fields `.status.succeeded` and `.status.failed` tell the number of successful and failed attempts respectively. The Rsync client Pods can be replaced with Jobs to leverage their inherent retry mechanism. The controller can watch the status of the Job instead of watching individual Pods.
```
djwhatle, line 59: I think it's a great idea to use the built in kube resource for this purpose.

enhancements/dvm-retry-mechanism/README.md
```diff
@@ -0,0 +1,203 @@
...
+if (dvmp.Status.PodPhase == corev1.PodPending) {
+  ...
+}
+```
+
+Similarly, the methods `hasAllRsyncClientPodsTimedOut()` and `isAllRsyncClientPodsNoRouteToHost()` will work by looking at the most recent Rsync Pod's status. The only difference is that these methods will only be triggered when the Job is in a terminal state.
+
+#### Deriving Rsync progress
+
+If the Rsync process fails at 10% in the first attempt, the second attempt starts from 0% and goes upto 90%. If the second attempt fails at 30%, the third attempt starts at 0% and goes upto 60% as rest of the 40% is already transferred in the previous attempts. The total progress percentage can be calculated by simply addding `lastObservedProgressPercent` of all Rsync attempts. The field `totalProgressPercentage` in DVMP _Status_ stores the calculated total percentage of all attempts. If a Job is determined to be successful, this value will be set to 100 without looking at individual pod percentages. In all other cases, the sum of all progress percentages will be used.
```
djwhatle, line 173: Want to make sure I understand this. 

1. Second attempt doesn't have to re-do any copies that are already completed, right?
2. How can second attempt fail at 30%? Is this 30% through verifying all files that were already transferred and then network cuts out? **OR** are the goalposts changed? Is 30% now actually 93% (30% through the last 10 percent)?

enhancements/dvm-retry-mechanism/README.md
```diff
@@ -0,0 +1,203 @@
...
+
+The _Status_ field in DVMP CR contains history of all Rsync attempts. The `LogMessage` field in each `PodStatus` only contains last 5 lines of the logs from Rsync Pod. In case of a failure even after all Rsync attempts, this information may not be sufficient to determine the cause. On the other hand, searching through logs of Pods of all Rsync attempts can be a daunting task. Therefore, the Rsync pod logs will be integrated with `mig-log-reader`. Currently, `mig-log-reader` only watches `openshift-migration` namespace. But there is a way to supply `--all-namespaces` along with a `--selector` argument which will be used to filter the Pods we need. The exact details of this implementation are not certain yet. But after discussing with [@djwhatle](https://github.com/djwhatle), it is clear that the `mig-log-reader` is best suited for this use case. New labels will be introduced across all the migration pods to make it easier to find pods across namespaces. Upon complete failure, users will simply be asked to look at logs collected by Stern in the source namespace. The relevant logs can be found by grepping for Rsync Pod names present in the DVMP CR _Status_.
```
djwhatle, line 177: Want to make sure I understand, will DVMP status still contain any log lines? Or will it have a list of Pods that were part of past rsync attempts?

enhancements/dvm-retry-mechanism/README.md
```diff
@@ -0,0 +1,203 @@
...
+
+#### Cleaning up Pods
+
+DVM creates one Rsync Pod for every _Persistent Volume_. In case of multiple failures, the number of Rsync Pods launched by the Job in a namespace can grow significantly possibly causing nuisance for users. The only purpose we leave the Rsync Pods behind is because we want to let users give a chance to debug issues in case of failures. However, if we are able to integrate the logs with `mig-log-reader`, we no longer need to keep the Rsync Pods. Therefore, all Jobs will deleted even when they fail. The history of Pods can still be found in DVMP _Status_. All pods associated with a Job are deleted as soon as the Job is deleted.
```
djwhatle, line 181: This could be nice. Have heard some murmers from users recently that leftover rsync pods are interfering with limited quota in user namespaces and causing apps to malfunction.

enhancements/dvm-retry-mechanism/README.md
```diff
@@ -0,0 +1,203 @@
...
+
+### Risks and Mitigations
+
+#### Job API on OpenShift 3.7, 3.8 and 3.9
+
+Kubernetes Job API introduced `.spec.backOffLimit` field in Kubernetes 1.11. The field was then backported to 1.10. It doesn't exist in 1.7, 1.8 and 1.9 which are supported versions of MTC. For these versions, the DVM controller will not use the `.spec.backOffLimit` field at all. Instead, it will implement it's own logic to determine whether the number of retries has reached the maximum specified number. 
+
+The _Status_ field of a Job has `.status.failed` field which shows the number of Pods that failed. Whenever the `.status.failed` value crosses the specified number of maximum attempts, the Job will be automatically deleted. As described in the previous section, we no longer need to keep the Rsync Pods behind for debug purposes.
```
djwhatle, line 189: > we no longer need to keep the Rsync Pods behind for debug purposes.

I would like to warn you about one potential problem with stern that **I think** I have fixed but I wonder if it will ever crop up again on misbehaving clusters. It seems that log tails can die after they are running for a long time (multiple hours, days) so I made a change so that stern reboots every 10 minutes with the message "Refreshing log tail". 

From what I can tell this fixed the issue, but I am wondering if we should give ourselves some kind of escape hatch to get log details directly from individual Pods in case stern goes completely haywire. Maybe some config switch to decide if these pods are preserved?

enhancements/dvm-retry-mechanism/README.md
```diff
@@ -0,0 +1,203 @@
...
+
+#### Cleaning up Pods
+
+DVM creates one Rsync Pod for every _Persistent Volume_. In case of multiple failures, the number of Rsync Pods launched by the Job in a namespace can grow significantly possibly causing nuisance for users. The only purpose we leave the Rsync Pods behind is because we want to let users give a chance to debug issues in case of failures. However, if we are able to integrate the logs with `mig-log-reader`, we no longer need to keep the Rsync Pods. Therefore, all Jobs will deleted even when they fail. The history of Pods can still be found in DVMP _Status_. All pods associated with a Job are deleted as soon as the Job is deleted.
+
+### Risks and Mitigations
+
+#### Job API on OpenShift 3.7, 3.8 and 3.9
+
+Kubernetes Job API introduced `.spec.backOffLimit` field in Kubernetes 1.11. The field was then backported to 1.10. It doesn't exist in 1.7, 1.8 and 1.9 which are supported versions of MTC. For these versions, the DVM controller will not use the `.spec.backOffLimit` field at all. Instead, it will implement it's own logic to determine whether the number of retries has reached the maximum specified number.
```
djwhatle, line 187: > Instead, it will implement it's own logic to determine whether the number of retries has reached

So for 3.7 and 3.9 we're going to implement our own version of the Jobs controller essentially? Want to make sure I'm understanding.

enhancements/dvm-retry-mechanism/README.md
```diff
@@ -0,0 +1,203 @@
...
+
+The _Status_ field of a Job has `.status.failed` field which shows the number of Pods that failed. Whenever the `.status.failed` value crosses the specified number of maximum attempts, the Job will be automatically deleted. As described in the previous section, we no longer need to keep the Rsync Pods behind for debug purposes.
+
+## Design Details
+
+### Test Plan
+
+An automation will be written to simulate transient Network failures by creating and deleting _Network Policy_ in source namespace intermittently. It will be verified that the Rsync Job works as expected in such scenario. Various aspects of progress reporting system will be verified. Mainly, it will be verified whether the progress bar picks up from where it left in the previous failed attempt. Stern integration will be verified.
```
djwhatle, line 195: Sweet

djwhatle (approved):
This proposal looks great. I added some questions.

enhancements/dvm-retry-mechanism/README.md
pranavgaikwad: @djwhatle The Job API still exists on these versions. It has the ability to retry but it doesn't have the ability to stop retries after the specified number of attempts are reached. It will keep retrying infinitely. We will only write logic to check the no of retries and delete the Job altogether when BackOffLimit is reached. There will be some amount of inaccuracy in actual number of attempts depending on when our controller reconciles.

enhancements/dvm-retry-mechanism/README.md
pranavgaikwad: @djwhatle Yes, DVMP status will still contain last 5 log lines of every Rsync attempt along with some additional information we care about.

enhancements/dvm-retry-mechanism/README.md
pranavgaikwad: @djwhatle Yes, that's correct. The second attempt doesn't actually copy the already transferred bits. However, the progress reporting system of Rsync doesn't start from where it left in the previous attempt. It starts from 0 and goes upto 90 if it transferred 10% in the previous attempt.

enhancements/dvm-retry-mechanism/README.md
pranavgaikwad: @djwhatle +1, agreed.

shawn-hurley: Pod logs are kept for some period of time on the node and are retrievable. The only issue will be if we don't know the pod name? Just a thought and wondering if that is a large enough escape hatch?

shawn-hurley (approved):
I think this as a whole makes a ton of sense.

Two things I think to discuss are: 
1. Pod vs Job and if that makes sense
2. What type of escape hatch do we need

I think that this is fine as is, and if we want to merge and continue the discussion during implementation?

enhancements/dvm-retry-mechanism/README.md
```diff
@@ -0,0 +1,203 @@
...
+Each Job will add unique labels on the Pods it creates to make sure the Pods associated with a Job can be found easily.
+
+#### Direct Volume Migration Progress
+
+Most of the logic in determining the state of the ongoing Rsync operation depends on _Status_ field of _DirectVolumeMigrationProgress_ CR which is created per Rsync client Pod. It takes a reference to the MigCluster on which Rsync Pods are running and the reference to the Pod itself: 
+
+```golang
+type DirectVolumeMigrationProgressSpec struct {
+	ClusterRef *kapi.ObjectReference `json:"clusterRef,omitempty"`
+	PodRef     *kapi.ObjectReference `json:"podRef,omitempty"`
```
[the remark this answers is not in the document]
pranavgaikwad: @shawn-hurley moving forward, PodRef will be ignored. I am open for removing this field. However, our existing DVMP CR doesn't have a _Completed_ condition which is a pattern we have been using in other CRs. I thought it would be better to keep this field in this version, add _Completed_ condition all existing DVMP CRs which have this field set and then modify controller to not reconcile _Completed_ CRs and then remove this field later on.

enhancements/dvm-retry-mechanism/README.md
```diff
@@ -0,0 +1,203 @@
...
+	[...]
+	// JobRef is a reference to the Rsync Job
+	JobRef     *kapi.ObjectReference `json:"jobRef,omitempty"`
+}
+```
+
+A new struct `RsyncPodStatus` will be introduced which will encapsulate the _Status_ information of an individual Rsync attempt. `RsyncPodStatuses` in the DVMP _Status_ will hold the historical information of all Rsync attempts. The `RsyncPodStatus` of the most recent Rsync attempt will also be included as top level field in DVMP status to avoid repeated searches through the list of attempts. Field `JobElapsedTime` will record the elapsed time of the ongoing Job. `Succeded`, `Failed` and `Completed` fields in the _Status_ will be introduced to hold the derived status of the Rsync operation. This simplifies logic wherever DVMP is being used in other places.
+
+```golang
+type DirectVolumeMigrationProgressStatus struct {
```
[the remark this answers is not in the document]
pranavgaikwad: @shawn-hurley I have added comment further down in the upgrade/downgrade section. Essentially, since all fields are optional, I don't see any issues with OLM validation we saw in 1.4.2. I will make sure the controller is designed in a way that existing DVMP CRs dont behave in unexpected ways causing breakage. Part of the solution is adding _Completed_ condition as described in the last section.

enhancements/dvm-retry-mechanism/README.md
```diff
@@ -0,0 +1,203 @@
...
+
+### Non-Goals
+
+1. Guarantee successful transfer when non-transient errors make the Rsync operation fail repeatedly.
+
+2. Allow users to manually retry Rsync transfer.
+
+## Proposal
+
+As of MTC 1.4.2, DVM controller creates a Pod to run Rsync. It creates as many Pods as there are volumes in the source cluster. Rsync command is injected as startup command in the Pod. When Rsync fails, the Rsync client container in the Pod never restarts as the _restartPolicy_ is set to _Never_ by default. It is possible to simply change the _restartPolicy_ of the Rsync client Pod to _OnFailure_ which will cause Kubelet to restart the Rsync container whenever failure takes place. However, it is not desired for our use case because Kube uses exponential backoff to restart the container. In case of multiple failures, the wait time between two consecutive retries can go upto 5 minutes. The time between consecutive launchs of a container cannot be configured. For our use-case, we want the pods to be re-launched immediately after a failure.
```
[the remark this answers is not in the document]
pranavgaikwad: @shawn-hurley That's correct. The primary reason I could think of was this backoff time. However, there are other reasons like 
1. We cant keep history
2. There would be some problems with progress reporting
3. The validations we have right now are based on container statuses, we may have to re-design those as well

I am open to considering both options and discuss pros and cons.

enhancements/dvm-retry-mechanism/README.md
```diff
@@ -0,0 +1,203 @@
...
+
+Throughout this document, a single _Rsync Pod_ launched by the Job will be referred to as an _Rsync Attempt_, while the collection of all attempts will be referred to as _Rsync Operation_, interchangeably. 
+
+### Implementation Details/Notes/Constraints
+
+#### Creating Job 
+
+In _VolumeMigration_ itinerary of the DVM controller, _[CreateRsyncClientPods](https://github.com/konveyor/mig-controller/blob/523d3a8ef70e89296535d2d0b94173669e92007c/pkg/controller/directvolumemigration/task.go#L126)_ phase is responsible for creating Rsync Pod. This phase will be renamed to _CreateRsyncClientJobs_. In that, new jobs will be created for each _Persistent Volume_. The same Pod template will be used in the Job's `.spec.template`. By default, `.spec.backOffLimit` will be set to `10`. Optionally, `backOffLimit` value will also be available for overriding through _DirectVolumeMigration_ CR:
```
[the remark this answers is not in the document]
pranavgaikwad: @shawn-hurley open to changing that. I don't have concrete reasons to find the right number.

enhancements/dvm-retry-mechanism/README.md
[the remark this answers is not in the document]
shawn-hurley: If it will require significantly more work and custom logic to use pods I think it doesn't make sense to do it if it is achievable with the same complexity it might be worth it.

enhancements/dvm-retry-mechanism/README.md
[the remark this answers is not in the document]
shawn-hurley: What happens on/if a rollback?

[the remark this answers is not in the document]
pranavgaikwad: @shawn-hurley I will have to discuss this with @alaypatel07 But AFAIK, currently DVM is out of rollback. Also, DVMP is just an informational CR. maybe we will just delete the existing DVMP CRs on rollback.

enhancements/dvm-retry-mechanism/README.md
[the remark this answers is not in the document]
pranavgaikwad: @alaypatel07 brought up this point as well. Will be happy to discuss this with you guys.

enhancements/dvm-retry-mechanism/README.md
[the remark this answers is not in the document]
shawn-hurley: Not our rollback a rollback in the controller going from 1.3.2 -> 1.4.3 -> 1.3.2 type of thing.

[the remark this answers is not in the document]
pranavgaikwad: @shawn-hurley Ohh I see. Haven't thought about that scenario yet. I will take some time to think about those scenarios and update the upgrade/downgrade section with my thoughts.

enhancements/dvm-retry-mechanism/README.md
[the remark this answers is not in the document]
djwhatle: > This will only happen based on the time it takes to fail right? Also on top of that, you need a bunch of retries to get to that point as well right?

I wasn't aware that the backoff eventually resets so I looked into this a bit. 

```
After containers in a Pod exit, the kubelet restarts them with an exponential back-off delay 
(10s, 20s, 40s, …), that is capped at five minutes. Once a container has executed for 10 minutes 
without any problems, the kubelet resets the restart backoff timer for that container.
```
I don't think using an exponential backoff makes sense in this context. Having rsync idling because it failed doesn't seem like it will increase our chance of success, but will slow down the overall migration process.

Source: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/

enhancements/dvm-retry-mechanism/README.md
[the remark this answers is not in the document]
djwhatle: What kind of behavior do we typically see here? Does rsync tend to fail after running for a while, or just for a little bit?

enhancements/dvm-retry-mechanism/README.md
djwhatle: @pranavgaikwad do we record all of the Pod names somewhere?

pranavgaikwad: @djwhatle yes, in DVMP.Status.PodStatuses[] field

enhancements/dvm-retry-mechanism/README.md
djwhatle: Not sure if this is convincing to you Shawn, but I do expect 3.7 and 3.9 will be a very minor portion of users.

enhancements/dvm-retry-mechanism/README.md
djwhatle: Ok, so in that case @shawn-hurley I could see pulling logs off the node as perhaps being sufficient. We would probably want to tweak must-gather to grab these logs from the dead Pods using any Pod names we find in DVMP status.

enhancements/dvm-retry-mechanism/README.md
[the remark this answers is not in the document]
pranavgaikwad: @djwhatle we have typically seen Rsync fail after some time (not immediately), but there's no pattern to the timing of it.

enhancements/dvm-retry-mechanism/README.md
pranavgaikwad: @djwhatle thanks for pointing that out. we should update must-gather as part of the implemention of this enhancement

enhancements/dvm-retry-mechanism/README.md
```diff
@@ -73,7 +73,9 @@ type DirectVolumeMigrationSpec struct {
 }
 ```
 
-Each Job will add unique labels on the Pods it creates to make sure the Pods associated with a Job can be found easily.
+#### Creating Pods
+
+In _VolumeMigration_ itinerary of the DVM controller, _[CreateRsyncClientPods](https://github.com/konveyor/mig-controller/blob/523d3a8ef70e89296535d2d0b94173669e92007c/pkg/controller/directvolumemigration/task.go#L126)_ phase is responsible for creating Rsync Pod. This phase will create pods as usual and add a unique label on each pod indicating which PVC it belongs to.
```
shawn-hurley, line 78: Can we also make sure that the name of the pod contains info for what it is doing like rsync-<vol-name>- as the generate-name?

PR #22, opened 2021-05-04
add crane-2.0 discovery enhancement

alaypatel07:


enhancements/crane-2.0-discovery/README.md
```diff
@@ -0,0 +1,127 @@
...
+ |  |   |- kind-a_namespace-1_name.yaml
+ |  |   |- kind-a_namespace-1_name-1.yaml
+ |  |   |- kind-b_namespace-1_name.yaml
+ |  |- namespace-2
+ |  |   |- kind-a_namespace-2_name.yaml
+ |  |   |- kind-a_namespace-2_name-1.yaml
+ |  |   |- kind-b_namespace-2_name.yaml
+ |- failures
+ |  |- namespace-1
+ |  |   |- failed-kind.yaml
```
shawn-hurley, line 83: this is failed kind because the list request failed right?

Should we add failed-kind-list or something to make that clear at a glance?

enhancements/crane-2.0-discovery/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Proposal
+
+There are three key areas for this tool to work on, first being the discovery
+ API to find the namespace scoped resources. Upstream Velero also uses the
+Kubernetes discovery API. This tool will import the discovery package of
+Velero to implement getting the resources. This enhancement will be followed
+by and indepth dive into what challenges are involved in using the velero
+package for Crane 2.0. During the extraction of resources, this tool will
+ skip "virtual resources" (resources that do not support "create" verb).
```
shawn-hurley, line 65: Do we also skip resources that do not have a list verb?

shawn-hurley (changes requested):
Some suggestions and some questions. Good start though IMO

enhancements/crane-2.0-discovery/README.md
```diff
@@ -0,0 +1,127 @@
...
+## Release Signoff Checklist
+
+- [x] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+
+## Summary 
+
+At a high level this tool will 
+- be used by the namespace admin (not cluster-admin)
```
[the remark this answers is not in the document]
alaypatel07: +1 makes sense, that captures it better I think

enhancements/crane-2.0-discovery/README.md
```diff
@@ -0,0 +1,127 @@
...
+discover and extract the GVK's available in the cluster that needs to be
+backed up. A conscious decision is made here to make sure to design this
+tool with a restricted privileges of non-admin user. It will be more
+empowering if the namespace owners backup the applications that they want
+/need to, instead of requiring cluster-wide admin privilege to backup
+the namespace. 
+
+## Non-goals
+
+- This tool will not be intended for a cluster admin, hence it will not backup
```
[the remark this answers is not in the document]
alaypatel07: well I was not clear what plans we have right now to handle cluster configuration, can we simply state here that cluster configuration is not the goal of this tool, we will create a separate enhancement that handles this case in the future (whenever we get to it). WDYT?

enhancements/crane-2.0-discovery/README.md
```diff
@@ -0,0 +1,127 @@
...
+```
+ export-dir
+ |- resources
+ |  |- namespace-1
+ |  |   |- kind-a_namespace-1_name.yaml
+ |  |   |- kind-a_namespace-1_name-1.yaml
+ |  |   |- kind-b_namespace-1_name.yaml
+ |  |- namespace-2
+ |  |   |- kind-a_namespace-2_name.yaml
+ |  |   |- kind-a_namespace-2_name-1.yaml
```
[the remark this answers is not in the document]
alaypatel07: +1

enhancements/crane-2.0-discovery/README.md
alaypatel07: >  this is failed kind because the list request failed right?

Yes but Im wonderng if that technical detail needed? I am not 100% sure if "foo-list.yaml" will be interpreted as "foo failed because of list method" or "list of all the foo kind resources that failed"

But I like the idea of providing "at a glance information of what failed"

enhancements/crane-2.0-discovery/README.md
alaypatel07: +1

enhancements/crane-2.0-discovery/README.md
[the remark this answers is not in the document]
alaypatel07: Updated @shawn-hurley PTAL

enhancements/crane-2.0-discovery/README.md
```diff
@@ -0,0 +1,127 @@
...
+- [x] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+
+## Summary 
+
+At a high level this tool will 
+- be used by the namespace admin (not cluster-admin)
+- use the Kubernetes discovery API to find the GVK's available in a cluster
+- will create a YAML dump of all the namespaced resources
```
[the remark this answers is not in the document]
alaypatel07: refined the language here PTAL

enhancements/crane-2.0-discovery/README.md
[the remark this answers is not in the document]
alaypatel07: removed

enhancements/crane-2.0-discovery/README.md
alaypatel07: updated

enhancements/crane-2.0-discovery/README.md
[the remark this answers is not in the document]
shawn-hurley: Yeag agree

enhancements/crane-2.0-discovery/README.md
shawn-hurley: Yes, I see, the difference, is it obvious when you look at the file that it was the list method?

shawn-hurley:
Let's add a quick section on the inputs and outputs to the library. 

is the library responsible for writing the files?

alaypatel07:
I am wondering how would a crane 2.0 user handle the difference in APIs? For example, during backup, a preferred API version is available, but upon restore it is not available? The solution in velero is to backup all the versions of the resource that increases the chance of one of the versions being available during restore. We also can handle this out of band in transform. I am just trying to nail down if we need to expose a flag for backing up all the version of the same object?

shawn-hurley:
I think conversion should be a seperate tool WDYT?

alaypatel07:
> I think conversion should be a seperate tool WDYT?

make sense, +1

> Let's add a quick section on the inputs and outputs to the library.


Updated.

enhancements/crane-2.0-discovery/README.md
```diff
@@ -0,0 +1,145 @@
...
+The third part is providing a UX like kubectl. This tool will try to read the
+default kube context, along with providing a `--namespace` flag to override
+the local context.
+
+## Packages
+
+### Discovery helper
+
+- the input to this library will be an authenticated discovery client
+ interface, and the namespace which needs to be backed up
```
shawn-hurley, line 138: Could this be a list of namespaces?

enhancements/crane-2.0-discovery/README.md
```diff
@@ -0,0 +1,145 @@
...
+- the output will be a
```
shawn-hurley, line 139: I think you probably need a list of "namespace output" and then the maps are in the namespaced outputs WDYT?

enhancements/crane-2.0-discovery/README.md
alaypatel07: My thought was to start with a single namespace, if we need a list of the namespace, make a decision then once we have a rough outline of the code.

enhancements/crane-2.0-discovery/README.md
alaypatel07: Same, start with a single namespace, I am debating with myself if returning a namespaced map would be better or calling this function with multiple namespaces in a loop would be better. Can we defer this decision?

shawn-hurley (approved):
Yep I agree defer decisions or stick with this and make the caller choose to make it a list or what not +1

shawn-hurley (approved):
Yep I agree defer decissions or stick with this and make the caller choose to make it a list or what not +1

PR #24, opened 2021-05-06
Adding transformations enhancement proposal for the library

shawn-hurley:


enhancements/crane-2.0/transformations.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Open Questions [optional]
+
+
+## Summary
+
+Transformations are needed when moving kubernetes manifests between clusters,
+between distributions or even to move particular objects to be reconciled by
+new controllers. 
+These transformations need to be sharable, testable, and idempotent.
```
alaypatel07, line 32: +1

enhancements/crane-2.0/transformations.md
```diff
@@ -0,0 +1,107 @@
...
+The transformations are essential for users who want the choice of distribution, 
+need to move applications between clusters, or want to move to a new deployment
+ mechansim such as [ArgoCD](https://argoproj.github.io/argo-cd/).
+
+### Goals
+
+- Ability to run transformations on a set of kubernetes manifests
+- Ability to codify the transformations and run them
+- Ability to run transformations such that they can be applied at one time
+- Transformations should be able to be determined if they conflict.
```
alaypatel07, line 45: I am wondering if this goal is valid here? Can we make a separate tool/subcommand that checks for transformation sanity? It feels to me that this falls slightly in the realms of checking/testing the transformations, and having this one point here will not cater justice to the problem space

enhancements/crane-2.0/transformations.md
```diff
@@ -0,0 +1,107 @@
...
+## Design Details
+
+### Implementation
+
+#### Packages
+
+##### pkg/cmd
+
+A library to implement a command that will conform to the defined
+CLI interface for a plugin.
```
alaypatel07, line 82: s/plugin/kubectl plugin?

enhancements/crane-2.0/transformations.md
```diff
@@ -0,0 +1,107 @@
...
+A library to implement a command that will conform to the defined
+CLI interface for a plugin. 
+The output will be a set of CLI flags and an options struct that can be used by
+the users as well as used by further library functions and packages
+
+##### pkg/transfrom
+
+A library that will be responsible for helpers for creating and updating the
+defined files for a transform directory.
+This will work on file byte arrays to make all the implementation details of
```
alaypatel07, line 90: can you please define the term "file byte arrays"?

enhancements/crane-2.0/transformations.md
```diff
@@ -0,0 +1,107 @@
...
+### Implementation
+
+#### Packages
+
+##### pkg/cmd
+
+A library to implement a command that will conform to the defined
+CLI interface for a plugin. 
+The output will be a set of CLI flags and an options struct that can be used by
+the users as well as used by further library functions and packages
```
alaypatel07, line 84: NIT: can you please break this down into an enum of the possible outputs. the two I read constructs to as:

```
The output will be a set of CLI flags and an options struct that can be used in the following manner
1. to generate transformations in a specified directory with the default transformation library
2. to apply the transformations from the transform directory and create YAML manifests in a specified directory
3. to generate a new library of transformation functions that can be added to extend the default transformation library

File: enhancements/crane-2.0/transformations.md (hunk 2 above)
Comment (shawn-hurley) [reply enhancements/crane-2.0/transformations.md:45]: I think I see the confusion, I think that a goal is not that we determine them and try and figure out the right thing to do, but instead that we focus on alerting/warning the user that there was a conflict. At the end of the day, for a given set of transformations we will always use the latest applied but giving the warning I think is important.

File: enhancements/crane-2.0/transformations.md (hunk 3 above)
Comment (shawn-hurley) [reply enhancements/crane-2.0/transformations.md:82]: I believe that we may want to define our own plugin interface defintion here. 

For instance, I can see us saying that any plugin must implement the "transform" subcommand, and if that exists it must have XXXXX flags. WDYT?

Comment (alaypatel07) [reply enhancements/crane-2.0/transformations.md:82]: As discussed on slack may be worth adding, "ideally we would like to support a container mode where users can provide the image they want to run and we will create a container with that image, pass all the flags to it and mount the transform directory from host into the container so that the container can write the transforms. This will provide a way to plug in images to run from CI/CD system for full extensibility"

File: enhancements/crane-2.0/transformations.md [hunk 6]
```diff
@@ -0,0 +1,109 @@
...
+#### Packages
+
+##### pkg/cmd
+
+A library to implement a command that will conform to the defined
+CLI interface for a plugin. 
+The output will be a set of CLI flags and a corresponding options struct.
+The output should be used by a new binary to:
+1. Generate transformations in a specified directory
+2. Create a new library/command to generate custom transformations
```
alaypatel07, line 86: I am a little confused, is the only purpose of the `cmd` package to help in creating new transforming binaries? Would we have a set of default out-of-the-box OpenShift transforms that can be invoked from here?   I am fine if they are in a separate `cmd` directory one level above the same level as `pkg` but just making sure if that is covered.

enhancements/crane-2.0/transformations.md
```diff
@@ -0,0 +1,109 @@
...
+
+##### pkg/transfrom
+
+A library that will be responsible for helpers for creating and updating the
+defined files for a transform directory.
+This will work on file byte arrays to make all the implementation details of
+the file, internal to the package.
+We will add a versioning system for this file and internally to this package
+
```
alaypatel07, line 95: Would it be helpful if we have a milestone for the helpers to start with? for instance, say that the first set of helpers will help in migration from OpenShift 3 to 4 or EKS to ROSA, or something that is more concrete and aligned to the MVP goal?

alaypatel07 (commented):
I have some clarifying questions added them inline, apart from this looks in good shape

enhancements/crane-2.0/transformations.md
shawn-hurley: I think that what we are talking about here, are helpers for a command to get well-defined names that the library can codify? Maybe this doesn't make sense here but is the UX of the cmd. Removing this section as I don't think it should be in this proposal you are correct.

enhancements/crane-2.0/transformations.md
shawn-hurley: I think this is an interesting question. Initially, the library to me feels like a runner, and then we need to implement specific transformation plugins for these things. 

Maybe this should be a separate library package? Shoudl this be in this PR or a different enhancement WDYT? @alaypatel07 @jmontleon

shawn-hurley: Added a `transform/kubernetes` section. I think it probably makes sense to start with this particular library and we can add openshift specifics later.

enhancements/crane-2.0/transformations.md
```diff
@@ -0,0 +1,104 @@
...
+##### transform/kubernetes
+
+Create the reusable plugin for the Kubernetes changes that must be applied to
+enable a transformation from one Kubernetes to another Kubernetes cluster.
+The plugin/library should be focused only on changes to manifests that will most likely have
+to run on all distributions/versions of Kubernetes.
```
alaypatel07, line 89: +1 this makes sense, so essentially this library should pass the k8s conformance tests.

enhancements/crane-2.0/transformations.md
alaypatel07: separate packages makes a lot of sense, +1

enhancements/crane-2.0/transformations.md
shawn-hurley: Yes, that is true. Do you want that added to the enhancement, or is it good as is?

alaypatel07 (approved):
ACK, had a question on the `pkg/transform/kubernetes` but will discuss it with Shawn on slack/call.

Everything else looks good to me, thanks

jmontleon (approved):
ACK

PR #26, opened 2021-06-16
State Only Migrations

pranavgaikwad:


enhancements/state-only-migrations/README.md
```diff
@@ -0,0 +1,156 @@
...
+### Non-Goals
+
+- Allow incremental migration of Kubernetes resources.
+
+## Proposal
+
+A new itinerary 'StateTransfer' will be introduced in MigMigration controller. This itinerary will address the migration of Persistent Volumes and Kubernetes resources. The _Persistent Volume_ migration bits of it will be exactly same as _Stage_ itinerary in that the users will have an option of chosing either _Direct_ or _Indirect_ migration, _Direct_ being the     default choice. For the Kubernetes resources, the users can provide a list of GVKs and a custom label selector per Migration. Unlike _Stage_ itinerary, image resources will not be a part by default unless explicitely included by the user using the GVK inclusion list.
+ 
+It is possible to simply change _Stage_ itinerary to include Kubernetes resources with some minor additions to PV migration flow. However, introducing new itinerary has some advantages. Since _Stage_ and _Final_ itineraries are not affected by the change, the existing MTC users can continue using them as they did before. Any new bugs will be localized to the new itinerary. A separate itinerary also makes it easier to itegrate Scribe in the future.
+
```
pranavgaikwad, line 61: @shawn-hurley Shawn commented on the other PR I opened which was merged by mistake:

```
I would like to see an alternative added, around not using a new itinerary and instead of moving things in the direction that uses sub-controllers for a "phase" with events being the mechanism for stating progress and status for reporting what sub-controller or "phase" it is in.
```

pranavgaikwad: @eriknelson Erik asked for an alternative approach as well.

enhancements/state-only-migrations/README.md
```diff
@@ -0,0 +1,156 @@
...
+#### Mapping PV names
+
+It is possible that when the users deployed their applications in the target cluster, the names of the resources changed due to usage of generateName or alike. In such scenarios, we need the ability to map a source PV to a target PV with a different name. To achieve that, we take an optional `targetName` for each PVC. The optional name will be provided by the user in the wizard and saved in the PV object on the MigPlan. It will be further propagated to the DVM CR:
+
+_migplan_types.go_
+
+```golang
+// PVC
+type PVC struct {
```
shawn-hurley, line 88: Is this a list of PVC objects?

enhancements/state-only-migrations/README.md
```diff
@@ -0,0 +1,156 @@
...
+```
+### Migrating Kubernetes resources
+
+Apart from Persistent Volumes, _StateTransfer_ will also allow users to include a subset of Kubernetes resources from their source namespaces. 
+
+The users will be able to provide a list of resources to include in the Velero Backup along with a label selector. The list of GVKs and the label selector will be used when creating a Velero Backup.
+
+#### Resource inclusion/exclusion
+
+By default, _StateTransfer_ will include following resources:
```
shawn-hurley, line 111: Is there an alternative here to just exposing these as fields on the migplan only, and keeping stage and final migration the same?

enhancements/state-only-migrations/README.md
```diff
@@ -0,0 +1,156 @@
...
+```
+The users will also be able to provide their own list of resource in addition to the above defaults. The list will be stored on _MigMigration_ CR:
+
+_migmigration_types.go_
+
+```golang
+type MigMigrationSpec struct {
+	[...]
+	// Comma separated list of included Kubernetes resources in Velero backup
+	IncludedResources string `json:"includedResources,omitempty"`
```
shawn-hurley, line 129: When this field is empty what does that mean? it will use the above list? how do I tell it no resources for only state PVC migration? How does this field get used or ignored when used in stage or final migrations.

enhancements/state-only-migrations/README.md
```diff
@@ -0,0 +1,156 @@
...
+}
+```
+
+Currently, MTC uses a list of `EXCLUDED_RESOURCES` which is provided in the MigrationController CR. That list will not be respected in the _StateTransfer_. 
+
+#### Resource list and label selector
+
+State only migration does not aim to include all the resources from source namespaces in a migration. Instead, it will give the users a label they can add on their resources.
```
shawn-hurley, line 137: Does this then work with final migrations?

pranavgaikwad: Please see [compatibility matrix](https://github.com/pranavgaikwad/enhancements/blob/master/enhancements/state-only-migrations/README.md#final-migration-compatibility) in the updated doc @shawn-hurley

enhancements/state-only-migrations/README.md
pranavgaikwad: Please see compatibility matrix in the updated doc

enhancements/state-only-migrations/README.md
pranavgaikwad: Maybe answered in the updated doc

enhancements/state-only-migrations/README.md
pranavgaikwad: Yes, this is a list of PVC objects in the MigPlan API.

enhancements/state-only-migrations/README.md
```diff
@@ -0,0 +1,180 @@
...
+
+It is assumed that the most Kubernetes resources will be migrated externally. There are two possible scenarios in this case - either the PVC objects will not be migrated externally or the PVC objects will be migrated externally and the PVs will be already provisioned in the target cluster. In the former case, MTC already has the ability to provision new PVs and no additional change is needed. In the latter case, MTC cannot migrate to PVs which are already provisioned in the target cluster. We will modify the DVM API to allow providing an optional target PVC name for each source PVC. Furthermore, the PVs can also be present in different namespace than the source. This is already handled in DVM. The combination of namespace and name mapping will facilitate State Only Migration. This also enables migrations within the same cluster in that users can leverage DVM where they wish to change the underlying storage.
```
alaypatel07, line 62: some questions regarding this section?

1.  `the PVs will be already provisioned` it says that PVs will be already provisioned, does that mean that the PVC for it will exist? 
2. if yes to the above answer, should we call out that the PVC name on the destination can be different from source ?
3. If yes to the above answer, what happens if data is already present on the PVC on the destination? will it be wiped out and the data will look an exact mirror of the source? or would we merge the data? if merge, how would we handle conflicts?

enhancements/state-only-migrations/README.md
```diff
@@ -0,0 +1,180 @@
...
+
+In the Final migration, MTC assumes that the users will be migrating the entire source namespace. In State Migration, the users will be interested in selecting a subset of resources from the namespaces. The Final migration will be updated to allow passing a list of GVKs to select from the source namespaces along with an optional label selector which will act as an additional filter for the included resources. Both of these features are offered by Velero. The final migration will simply propagate them as additional Backup options to Velero.
+
+In an effort to eventually integrate Scribe in DVM, we will also begin replacing some of the internals in DVM with Crane 2 state transfer library. Hence, it is possible that the actual implementation of the API changes in this doc will take place in crane library. They will be covered in their own docs and referenced here.
+
+
+### Implementation Details/Notes/Constraints [optional]
+
+#### Selecting PVs
```
alaypatel07, line 71: Do you mean selecting PVCs?

enhancements/state-only-migrations/README.md
```diff
@@ -0,0 +1,180 @@
...
+
+Optionally, the users will be able to provide a label selector to further filter the included resources. The label selector will also be accepted through MigPlan:
+
+_[migplan_types.go](https://github.com/konveyor/mig-controller/blob/ee9670d4054b76f9b63fbb12c2395302a14d0d4c/pkg/apis/migration/v1alpha1/migplan_types.go#L65-L93)_
+
+```golang
+type MigPlanSpec struct {
+	[...]
+
+	LabelSelector *metav1.LabelSelector `json:"labelSelector,omitempty"`
```
alaypatel07, line 127: what would happen if the user configures the LabelSelector but does not configure the `IncludedResources`?

Are both the `IncludedResources` and `LebalSelector` separate features that could work irrespective of the other or will label selector only be applied if IncludedResources is not empty?

enhancements/state-only-migrations/README.md
```diff
@@ -0,0 +1,180 @@
...
+
+> Note: The complete resource inclusion hierarchy is explained in detail [in this section](#final-migration-compatibility).
+
+
+
+### Risks and Mitigations
+
+#### Migrating into an existing namespace
+
+If a namespace was created as a result of GitOps/Pipeline deployment in the target cluster, it is possible that the UID range of the namespace in the target cluster differ from that of the source. In this case, migrating Persistent Volumes from the source to the target cluster will result in a mismatch in file ownership and the PVs won't be usable as-is. There are various ways to solve this problem. One of the safer solutions is to patch the namespace before migrating PVs such that the annotations are equal to that of source ns. If there are any workloads running in that ns already, the workloads need to be queisced and unquisced once post update.
```
alaypatel07, line 141: > One of the safer solutions is to patch the namespace before migrating PVs

So this is safer solutions from a migration perspective, but not really safe for the user right? for example, assume that the there is another namespace in the destination cluster with the same annotations as the source and a user in that namespace has access to hostPath. They could have access to the data of migrated namespace right?

Instead can we call out here that "when migrating to different namespace/PVC there will be permissions issues will will be mentioned in a separate enhancement in the future, for now we will document the various options that user has and they could decide which one is safe for them to move forward" ?

alaypatel07 (changes requested):
The API looks good to me. Had some questions on other sections of the implementation, PTAL

enhancements/state-only-migrations/README.md
pranavgaikwad: @alaypatel07 
1. Yes, it is assumed that the PVs will be provisioned as a result of pre-existing PVC in the target namespace. The PVC will be a part of application YAML definition stored in Git. 

2. Yes, mapping of names assumes that the PVC name will be different, we don't care about PV name.

3. The default behavior would be to wipe out existing data and replace it with source data making it a clone of source. This can be adjusted at Rsync level using different options. However, we only consider default use case of overwriting.

enhancements/state-only-migrations/README.md
pranavgaikwad: Yes, I am into habit of using PV <-> PVC interchangeably, but in general, this entire document assumes that for every PV, there is an associated PVC.

enhancements/state-only-migrations/README.md
pranavgaikwad: @alaypatel07 all of the different scenarios are explained in the compatibility matrix in the Upgrade/Downgrade section. But to answer your question, yes! Both the settings can be used independently, meaning, empty list of included resources with a non-empty label selector implies that all the resources that match the label will be included.

enhancements/state-only-migrations/README.md
pranavgaikwad: @alaypatel07 I just wanted to include this section to note down the "risk". The proposed solution is one of the different solutions we will be considering. I believe that the problem requires its own enhancement doc and we keep this section as-is just to call out the possibility of failures.

enhancements/state-only-migrations/README.md
sseago: I was slightly confused by the wording as well, leading to my own version of 1) in my head. Maybe a minor wording change above -- from "and the PVs will be already provisioned in the target cluster" to "and the PVs will be already provisioned in the target cluster bound to PVCs" -- later sentences make this clearer already, but if you make it clear from the first mention of that use case that we'll be copying into already-existing (and possibly already mounted) PVCs, this will make it clearer.

sseago: Also, note that if the applications have already been deployed to the target/dest cluster, then we may run into similar RWO-oriented challenges as we have on the source cluster when not quiescing. If it's not already in the design, you may need to offer a similar quiesce option on the target cluster here -- quiesce apps, dvm/rsync to the volumes, un-quiesce.

enhancements/state-only-migrations/README.md
```diff
@@ -0,0 +1,180 @@
...
+
+In an effort to eventually integrate Scribe in DVM, we will also begin replacing some of the internals in DVM with Crane 2 state transfer library. Hence, it is possible that the actual implementation of the API changes in this doc will take place in crane library. They will be covered in their own docs and referenced here.
+
+
+### Implementation Details/Notes/Constraints [optional]
+
+#### Selecting PVs
+
+The users will be able to select PVs to be included in the migration in the UI. This will be achieved by setting `PvSkipAction` on any PVs excluded from the migration. This is already possible in our current API and does not need any change. Most changes around this will be addressed in the UI.
```
sseago, line 73: You may need some minor changes to the PvSkipAction. I think right now we only allow skipping PVCs which are *not* connected to source cluster applications. The thinking was that since MTC doesn't allow the user to skip particular deployments, we don't allow users to skip that deployment's PVCs either. However, if there are PVCs that are currently unconnected to applications, those *are* skippable. We may need to relax this restriction to enable state only migrations.

enhancements/state-only-migrations/README.md
sseago: So the *namespace* will be constrained by the migplan's namespace-to-namespace mapping. This means that within a namespace, any PVC can be mapped to any new name, but all PVCs in a particular source namespace must be mapped to other PVCs in the same target/dest namespace (same as each other, not necessarily the same as src). In other words, if src has ns1/pvc1, ns1/pvc2, ns2/pvc3, ns2/pvc4, then you could map those first two ns1 PVCs to ns5/pvc10 and ns5/pvc11 but you couldn't map those to ns5/pvc10 and ns6/pvc11.

sseago: (to be clear, the above applies to migplan-driven migrations. A raw DVM resource created could map any PVC to any other PVC, since in the DVM CR, the only place PVC namespaces are identified is within a single PVCToMigrate struct)

enhancements/state-only-migrations/README.md
sseago: @alaypatel07 Since these are just being passed on to velero, the two act independently. include/exclude are related -- if exclude is given but include is not, then the effect is "everything except exclude", but if include is given, then only those resources are included. Then, in addition to this, if a label selector is given, then once we have a full list of resources to back up from the include/exclude lists, that list is then filtered on a per-item basis using the label selector. This is how we do stage backups today. We have a limited set of included resources (PVCs, ImageStreams, etc.) *and* a label selector (the IncludedInStageBackup label).

sseago: @pranavgaikwad Are these includes/labels only being used for final backup? For stage are we using the existing label and include list?

enhancements/state-only-migrations/README.md
sseago: @alaypatel07 " assume that the there is another namespace in the destination cluster with the same annotations as the source and a user in that namespace has access to hostPath. They could have access to the data of migrated namespace right" -- note that this is also a problem today in current (and prior) versions of MTC, since we also create the namespace with that UID range which may be in use in the target cluster already.

enhancements/state-only-migrations/README.md
alaypatel07: > Yes, mapping of names assumes that the PVC name will be different, we don't care about PV name.

Since PVC is the unit of data transfer, can I ask to refactor this entire doc to remove all usage of PV and replace it with PVC?

> The default behavior would be to wipe out existing data and replace it with source data making it a clone of source. This can be adjusted at Rsync level using different options. However, we only consider default use case of overwriting.

Sounds fair to me as long as this is explicitely documented. Can you please include this in the doc?

enhancements/state-only-migrations/README.md
alaypatel07: @sseago noted, but I believe, with this new work there is an opportunity at solving that shortcoming.

@pranavgaikwad I agree that noting down the "risk" is the right thing to do, and I also agree that this enhancement does not intend to provide the solution, but doesnt the following do exactly that i.e. suggest a solution?

> One of the safer solutions is to patch the namespace before migrating PVs such that the annotations are equal to that of source ns. If there are any workloads running in that ns already, the workloads need to be queisced and unquisced once post update.

enhancements/state-only-migrations/README.md
shawn-hurley: Why can't I pre-create a volume and put data there with this feature? 

Why are we assuming that the PVC would be created and the pipeline would be run before transferring the data?

enhancements/state-only-migrations/README.md
alaypatel07: > I really would like us to consider if PV is the correct thing to consider here.

I don't really think that is possible. I would love to see what I am missing.

enhancements/state-only-migrations/README.md
shawn-hurley: I think I would suggest that we consider that for now, we force a user to take an action and just warn them when this happens? Do we have the ability to do that?

enhancements/state-only-migrations/README.md
```diff
@@ -0,0 +1,180 @@
...
+
+#### Stage migration compatibility
+
+Since the PV mapping will be handled in DVM API, no change is needed in the _Stage_ migration or in any of its itinerary phases. The API changes in MigPlan and DVM both introduce an optional `targetName` variable each. 
+
+#### Final migration compatibility
+
+Since we will be adding optional inclusion list and label selector to the MigPlan which never existed in the API before, we will ensure that the MigPlans created in the previous versions are not adversely affected. Also, the proposed inclusion list differs from the existing `EXCLUDED_RESOURCES` list in that the exlusion list is a global configuration, while the inclusion list is at a MigPlan level. It is important to consider how different configurations will affect each other. Following table lists all possible combinations of the old and the new configurations:
+
+| No 	|   Resources in source ns   	| Global exclusion list 	| MigPlan inclusion list 	| Label Selector 	|           Outcome          	|
```
shawn-hurley, line 168: Are A B and C GVK's?

shawn-hurley (commented):
I believe that Alay is correct regarding the PV thing and that I am wrong.

I think this enhancement from an API sense makes sense I would just make sure two things:

1. That any absence of values means that all values are used to preserve backwards compatibility,
2. That these fields are 100% optional and them not existing is equal to the absence of the value.

enhancements/state-only-migrations/README.md
```diff
@@ -0,0 +1,180 @@
...
+
+In the _Final_ migration, the users will be able to provide an optional list of resources to be included in the migration. It will be converse of MTC's `EXCLUDED_RESOURCES` in that the the only resources present in the inclusion list will be migrated. The inclusion list will be accepted through MigPlan:
+
+_[migplan_types.go](https://github.com/konveyor/mig-controller/blob/ee9670d4054b76f9b63fbb12c2395302a14d0d4c/pkg/apis/migration/v1alpha1/migplan_types.go#L65-L93)_
+
+```golang
+type MigPlanSpec struct {
+	[...]
+
+	IncludedResources []string `json:"includedResources,omitempty"`
```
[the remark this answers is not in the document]
sseago: @shawn-hurley We're using this to set `IncludedResources` on the velero backup spec, so we need to use the same format as velero. For example, for a current MTC stage backup, we'd have something like this:
```
  includedResources:
  - serviceaccount
  - persistentvolumes
  - persistentvolumeclaims
  - imagestreams
  - secrets
  - configmaps
  - pods
```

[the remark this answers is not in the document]
sseago: From the velero BackupSpec struct definition:
```
        // IncludedResources is a slice of resource names to include
        // in the backup. If empty, all resources are included.
        // +optional
        // +nullable
        IncludedResources []string `json:"includedResources,omitempty"`
```

enhancements/state-only-migrations/README.md
pranavgaikwad: Yeah, I should have been more clear on the naming. Yes, alphabets are GVKs and digits are the labels added on instances of them.

enhancements/state-only-migrations/README.md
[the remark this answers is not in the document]
shawn-hurley: I would really prefer this be a GVR or a GR. 

I assume that those work for core types but what about non-core types?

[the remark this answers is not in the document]
sseago: We really can't do anything other than "what velero does" since we're passing them directly over. I haven't actually looked at the internal velero parsing code directly to know the limits of what is parsed out of this, but they certainly do work for non-core types, CRDs, etc.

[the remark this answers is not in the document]
shawn-hurley: We can pass that down, however, velero needs, that doesn't mean that our API has to look at the same if there is a better way?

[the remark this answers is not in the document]
sseago: If we want to create our own custom spec field, parse it, and generate the velero backup.Spec.IncludeResources in a way that works for velero, nothing is stopping us. Whether this is a "better way" is not at all clear to me, though. We'd need to identify what we would gain for this extra effort.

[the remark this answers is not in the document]
sseago: Here's where Velero deals with individual elements of this slice:
https://github.com/vmware-tanzu/velero/blob/main/pkg/util/collections/includes_excludes.go#L202

enhancements/state-only-migrations/README.md
pranavgaikwad: @sseago yes, just the final backup. Stage Backup remains untouched

enhancements/state-only-migrations/README.md
pranavgaikwad: @shawn-hurley we currently don't have that ability. However, it is not hard to implement in DVM. Good point.

enhancements/state-only-migrations/README.md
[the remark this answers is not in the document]
shawn-hurley: If I wanted to exclude deployments today, what string do I need to add?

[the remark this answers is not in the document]
sseago: If IncludeResources is non-empty, you just need to make sure that "deployments" isn't there. A non-empty IncludeResources means that only the listed resources are included. If IncludeResources isn't specified, then everything *except* ExcludeResources is included.

enhancements/state-only-migrations/README.md
pranavgaikwad: @alaypatel07 risks and mitigations :)

alaypatel07: fair @pranavgaikwad

enhancements/state-only-migrations/README.md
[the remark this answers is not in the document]
pranavgaikwad: @shawn-hurley As Scott suggested, having a GVK list in our API means that we have to do reverse of what Velero is doing to get GVK from the short names and then pass that list to Velero. Keeping it a list of strings means that we don't need that other logic.

enhancements/state-only-migrations/README.md
pranavgaikwad: @alaypatel07 updated

enhancements/state-only-migrations/README.md
pranavgaikwad: @alaypatel07 called it out specifically that this problem will be addressed separately.

enhancements/state-only-migrations/README.md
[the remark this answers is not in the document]
pranavgaikwad: @shawn-hurley @sseago As we discussed offline, it appears that everyone is in fact in agreement on changing this from a string slice to a structured type. We will also propose to update the upstream Velero api to make the inclusion lists structured. As for MTC, irrespective of what Velero does, we will stick with the structured types, most likely `metav1.GroupResource`.  Until Velero accepts the change, we will hide the conversion in MTC.

enhancements/state-only-migrations/README.md
pranavgaikwad: @sseago Updated the document to specify this catch.

enhancements/state-only-migrations/README.md
pranavgaikwad: I added a footer to the table explaining what is what. I believe that answers your question. Resolving!

shawn-hurley (approved):

sseago (approved):

pranavgaikwad:
Since we have 2 approvals out of 3, I am going ahead and merging this PR. Thank you all for your inputs.

PR #28, opened 2021-06-24
transform plugin priorities design

sseago:
Fixes https://github.com/konveyor/enhancements/issues/30

### DIFF
--- a/enhancements/crane-2.0/transform-plugin-priority/README.md
+++ b/enhancements/crane-2.0/transform-plugin-priority/README.md
@@ -1,0 +1,205 @@
+---
+title: transform-plugin-priority
+authors:
+  - "@sseago"
+reviewers:
+  - "@shawn_hurley"
+  - "@alaypatel07"
+  - "@jmontleon"
+approvers:
+  - "@shawn_hurley"
+  - "@alaypatel07"
+  - "@jmontleon"
+creation-date: 2021-06-24
+status: implementable
+see-also:
+  - "/enhancements/this-other-neat-thing.md"  
+replaces:
+  - "/enhancements/that-less-than-great-idea.md"
+superseded-by:
+  - "/enhancements/our-past-effort.md"
+---
+
+# Transform plugin priority
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] User-facing documentation is created
+
+## Open Questions
+
+## Summary
+
+Currently, when generating a transformation, the plugins passed into
+the transform runner are executed in slice order, and when there are
+conflicting patches generated, the plugin which executed first in the
+conflict set gets its patch applied. Subsequent plugins whose patches
+conflict with the first get their patches discarded. The patch run
+order isn't necessarily always predictable, and when it is
+predictable, it the ordering has nothing to do with priority or
+importance.
+
+We need to introduce a notion of plugin priority which is associated with
+a transform action, rather than an inherent attribute of each
+plugin. The user should be able to provide a non-exhaustive list of plugin names in
+priority order as a CLI option for the `transform` command. This could
+be a single "must take priority" plugin, or a list of all plugins in
+priority order. Any plugins not prioritized in this way will be
+considered according to the default application order, based on the
+slice passed into the runner.
+
+## Motivation
+
+Being able to choose which plugin actions take priority becomes
+important when third party or custom plugins are used in conjunction with
+included Crane plugins.
+
+### Goals
+
+- Ability to choose, for a single CLI transformation run, which
+  plugins (if any) should take precedence
+- Ability to run without prioritization, if desired
+
+### Non-Goals
+
+- Whiteout behavior will not be affected by prioritization. If any
+plugin whites out a resource, it will be excluded.
+
+## Proposal
+
+### User Stories
+
+#### Story 1
+
+As a library user, I would like to be able to control the priority of
+one or more plugin patches when conflicts arise.
+
+#### Story 2
+
+As a transform CLI user, I would like to be able to control the priority of
+one or more plugin patches when conflicts arise.
+
+### Implementation Details/Notes/Constraints
+
+The `transform.Plugin` interface will need to add a `Name()` func to
+return the name of the plugin, since plugin names will be used to
+identify priorities.
+
+## Design Details
+
+### crane-lib changes
+
+#### transform package
+
+`transform.Plugin` will need a new `Name()` func:
+
+```go
+type Plugin interface {
+	// Determine for a given resources what the plugin is deciding to do with this
+	Run(*unstructured.Unstructured) (PluginResponse, error)
+	// Returns the name of the plugin
+	Name() (string)
+}
+```
+Name uniqueness won't be enforced at this point. If two plugins return
+the same name, then conflicts between them will be handled in the same
+way as conflicts between plugins that are not in the prioritized list
+-- the plugin patch earliest in the patch list will be applied.
+
+For binary plugins, the initial implementation will use the filename
+as the plugin name. A possible future enhancement would be for the
+plugin itself to return its own name in response to a "name" or
+"version" request.
+
+`transform.Runner` will need a `PluginPriorities` map field:
+
+```go
+type Runner struct {
+	// This is where we need to put extra info
+	// This should include generic args to be passed to each Plugin
+	// This also needs to handle the options that it will need.
+	// TODO: Figure out options that the runner will need and implement here.
+	PluginPriorities map[string]int
+	Log *logrus.Logger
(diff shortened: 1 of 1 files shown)


enhancements/crane-2.0/transform-plugin-priority/README.md
```diff
@@ -0,0 +1,218 @@
...
+conflicting patches generated, the plugin which executed first in the
+conflict set gets its patch applied. Subsequent plugins whose patches
+conflict with the first get their patches discarded. The patch run
+order isn't necessarily always predictable, and when it is
+predictable, it the ordering has nothing to do with priority or
+importance.
+
+We need to introduce a notion of plugin priority which is associated with
+a transform action, rather than an inherent attribute of each
+plugin. The user should be able to provide a non-exhaustive list of plugin names in
```
alaypatel07, line 50: does this mean that we expect the user to know what result each plugin is producing in detail?

enhancements/crane-2.0/transform-plugin-priority/README.md
```diff
@@ -0,0 +1,218 @@
...
+priority order as a CLI option for the `transform` command. This could
+be a single "must take priority" plugin, or a list of all plugins in
+priority order. Any plugins not prioritized in this way will be
+considered according to the default application order, based on the
+slice passed into the runner.
+
+The `Summary` section is incredibly important for producing high quality
```
alaypatel07, line 57: needs to be removed?

enhancements/crane-2.0/transform-plugin-priority/README.md
sseago: @alaypatel07 Possibly. For example they have a custom plugin that overrides one of our out-of-the-box ones. Another possibility is that the user did an initial transform run, noticed that there was a conflict, decided that the operation that was thrown out was one they wanted, so they re-ran with the priority applied. That's why priority is a function of a transform operation and not inherently tied to the plugin.

enhancements/crane-2.0/transform-plugin-priority/README.md
sseago: Yes -- I'll pull it out on the next revision.

enhancements/crane-2.0/transform-plugin-priority/README.md
```diff
@@ -0,0 +1,218 @@
...
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] User-facing documentation is created
+
+## Open Questions
+
+What should the CLI flag be for "plugin-priorities", as "-p" is
```
shawn-hurley, line 34: For now, let's just not give it a short name and only make it a long name? WDYT?

enhancements/crane-2.0/transform-plugin-priority/README.md
```diff
@@ -0,0 +1,218 @@
...
+
+## Motivation
+
+Being able to choose which plugin actions take priority becomes
+important when third party or custom plugins are used in conjunction with
+included Crane plugins.
+
+### Goals
+
+- Ability to choose, on a per-transform-action basis, which plugins
```
shawn-hurley, line 73: Really quick point of clarification to make sure I am following:

The goal is for the runner to be able to select on a per transform action which one to choose. 
The user interface is not per transform action, rather is on a plugin in basis I thought?

enhancements/crane-2.0/transform-plugin-priority/README.md
```diff
@@ -0,0 +1,218 @@
...
+-- the plugin patch earliest in the patch list will be applied.
+
+For binary plugins, the initial implementation will use the filename
+as the plugin name. A possible future enhancement would be for the
+plugin itself to return its own name in response to a "name" or
+"version" request.
+
+`transform.Runner` will need a `PluginPriorities` map field:
+
+```
```
shawn-hurley, line 130: nit: add go here for pretty rendering

enhancements/crane-2.0/transform-plugin-priority/README.md
```diff
@@ -0,0 +1,218 @@
...
+	// This is where we need to put extra info
+	// This should include generic args to be passed to each Plugin
+	// This also needs to handle the options that it will need.
+	// TODO: Figure out options that the runner will need and implement here.
+	PluginPriorities map[string]int
+	Log *logrus.Logger
+}
+```
+
+`PluginPriorities` will be a mep of plugin name keys and priority
```
shawn-hurley, line 141: nit: s/mep/map

enhancements/crane-2.0/transform-plugin-priority/README.md
```diff
@@ -0,0 +1,218 @@
...
+`PluginOperation`:
+```
+type PluginOperation struct {
+	PluginName string
+	Operation  jsonpatch.Operation
+}
+```
+Instead of just building a `jsonpatch.Patch` (which is just
+`[]jsonpatch.Operation`), we'll build a `[]PluginOperation`
+When the runner calls SanitizePatchesm this `PluginOperation` list
```
shawn-hurley, line 157: nit: s/SanitizePatchesm/SanitizePataches

enhancements/crane-2.0/transform-plugin-priority/README.md
```diff
@@ -0,0 +1,218 @@
...
+that of the previous operation for this `operationKey`, then we will
+replace the matching entry in `patchMap` with the current operation,
+and put the previous map entry on the ignored list. Otherwise, the
+current operation in the map will be retained.
+
+#### binary_plugin package
+
+`BinaryPlugin` will get a new `name` field:
+```
+type BinaryPlugin struct {
```
shawn-hurley, line 179: nit: Missing name Field

shawn-hurley (changes requested):
Nits and a point of clarification other than that it looks good to me.

enhancements/crane-2.0/transform-plugin-priority/README.md
sseago: Works for me.

enhancements/crane-2.0/transform-plugin-priority/README.md
sseago: I think the original wording here reflects my understanding of how this works (and how I've explained it below), but perhaps it's not clear enough? The priorities passed in are based purely on plugin name, but priorities are only valid for *this* transform. If you run it again without selecting priorities, we're back at using the default "no priorities" and the plugins which appear first in the conflict win. I think the only clear way to describe this is "plugin priorities for this transform command invocation", which re-stated as a goal is what I have above. Is there a better way to phrase it here so that the meaning is clearer?

sseago: If options end up eventually specified via env vars, then in fact a given set of options (including priorities, but also export dir, etc.) will be used across multiple transforms, but that's a side effect of the way environment variables work. The command options are still specified on a per command invocation basis.

sseago: But maybe the confusion here is "transform action" of a single plugin vs. "transform action" -- i.e. the user action of running the transform command which invokes multiple plugins across multiple resources. The latter is what I mean. Perhaps "per-transform-command-invocation" is clearer?

enhancements/crane-2.0/transform-plugin-priority/README.md
sseago: s/SanitizePataches/SanitizePatches, but yes.

enhancements/crane-2.0/transform-plugin-priority/README.md
shawn-hurley: Or make it clear the goal is about the CLI and not the library bits. I would probably say "Ability to choose, for a CLI transformation run, which plugins...." Or something like that. 

I agree what I stated above and what you describe down below are the same.

sseago: How about this (based on your last suggestion):
```
- Ability to choose, for a single CLI transformation run, which plugins (if any) should take precedence
```

sseago:
Feedback received so far has now been addressed.

shawn-hurley (approved):
Looks good to me

I think that we need to add an issue to cover the need for the runner to determine conflicts in operations for the same path, I don't think that sanitizePatches does that today

sseago:
@shawn-hurley That's right. I think different operations on the same path don't register as a conflict. I think the answer there is to just stop using the Kind+Path operation key struct and just use the Path as a string key for patchMap. I'll create a crane-lib issue and we can discuss the fix there, though. That's an issue that's largely orthogonal to this one (although the fixes will modify the same code)

sseago:
@shawn-hurley https://github.com/konveyor/crane-lib/issues/32

PR #31, opened 2021-06-28
DVM Crane State Transfer Feature Parity

pranavgaikwad:
This is a feature parity enhancement document created to propose solutions to different DVM features we want to implement in Crane 2 to maintain feature parity. This is also a follow-up on the previous State Only Migrations doc I proposed. It will address the implementation details for the API changes proposed in the state only mgirations as well.

### DIFF
--- a/enhancements/crane-2.0/dvm-crane-feature-parity/README.md
+++ b/enhancements/crane-2.0/dvm-crane-feature-parity/README.md
@@ -93,36 +93,45 @@ _options.go_
 ```golang
 package rsync
 
-type RsyncTransferOptions struct {
-	RsyncCommandOptions
-	RsyncResourceMetadata
+// TransferOptions defines customizeable options for Rsync Transfer
+type TransferOptions struct {
+	CommandOptions
+	sourceResourceMetadata      transfer.ResourceMetadata
+	destinationResourceMetadata transfer.ResourceMetadata
 }
 
-// RsyncCommandOptions defines options that can be customized in the Rsync command
-type RsyncCommandOptions struct {
-	Recursive            bool
-	SymLinks             bool
-	Permissions          bool
-	ModTimes             bool
-	DeviceFiles          bool
-	SpecialFiles         bool
-	HardLinks            bool
-	Delete               bool
-	Partial              bool
-	BwLimit              int
-	Info                 string
+// TransferOption
+type TransferOption interface {
+	ApplyTo(*TransferOptions) error
 }
 
-// RsyncResourceMetadata defines options that let consumers customize different metadata placed on Rsync resources
-type RsyncResourceMetadata struct {
-	SourcePodLabels                 map[string]string
-	SourcePodAnnotations            map[string]string
-	SourcePodOwnerReferences        []metav1.OwnerReference
-	SourcePodSecurityContext        *corev1.SecurityContext
-	DestinationPodLabels            map[string]string
-	DestinationPodAnnotations       map[string]string
-	DestinationPodOwnerReferences   []metav1.OwnerReference
-	DestinationPodSecurityContext   *corev1.SecurityContext
+// CommandOptions defines options that can be customized in the Rsync command
+type CommandOptions struct {
+	Recursive    bool
+	SymLinks     bool
+	Permissions  bool
+	ModTimes     bool
+	DeviceFiles  bool
+	SpecialFiles bool
+	Groups       bool
+	Owners       bool
+	HardLinks    bool
+	Delete       bool
+	Partial      bool
+	BwLimit      int
+	Info         []string
+	Extras       []string
+}
+```
+Struct _ResourceMetadata_ defines extra customizeable information that can be used while creating any temporary resource used by State Transfer.
+
+```golang
+// ResourceMetadata defines options that let consumers customize different metadata placed on Rsync resources
+type ResourceMetadata struct {
+	Labels            map[string]string
+	Annotations       map[string]string
+	OwnerReferences   []metav1.OwnerReference
+	SecurityContext   *corev1.SecurityContext
 }
 ```
 
@@ -154,16 +163,17 @@ type RsyncTransferOption interface {
 This interface will prove useful in cases where we want to provide users with a set of pre-defined options like below: 
 
 ```golang
-type RsyncCommandArchive bool
+type ArchiveFiles bool
 
-func (rca RsyncCommandArchive) ApplyTo(opts *RsyncTransferOptions) error {
+func (rca ArchiveFiles) ApplyTo(opts *TransferOptions) error {
 	opts.Recursive = bool(rca)
 	opts.SymLinks = bool(rca)
 	opts.Permissions = bool(rca)
 	opts.ModTimes = bool(rca)
-	opts.DeviceFiles = bool(rca)
 	opts.Groups = bool(rca)
-	opts.Users = bool(rca)
+	opts.Owners = bool(rca)
+	opts.DeviceFiles = bool(rca)
+	opts.SpecialFiles = bool(rca)
 	return nil
 }
 ```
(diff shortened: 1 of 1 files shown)


enhancements/crane-2.0/dvm-crane-feature-parity/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+
+#### Allowing custom labels and custom Rsync options for Rsync Transfer
+
+Both of these features are considered together as they are additional options a user may provide for Rsync Transfer. It makes sense to generalize such additional options by creating a new interface `TransferOptions`:
+
+```golang
+type TransferOptions interface {
+	Validate() error
+	SetDefault()
```
alaypatel07, line 92: I am having little trouble understanding this interface. 

So from my understanding of looking at the interface mentioned above, the `CreateClient` method of the interface is supposed to use the Rsync options. When the CreateClient function is called can you help me understand how do you see the usage of the methods of this interface to populate the Client pod?

pranavgaikwad: The usage of validate() doesn't populate the Client Pod. Everytime a transfer uses its options, it calls the `Validate()` method,  if the method returns an error, `SetDefault()` will be called to set a sane default.

alaypatel07: So after calling `SetDefault()` you still need to take those options and convert them into rsync command right? How would that work with the interface?

pranavgaikwad: @alaypatel07 this interface is just impose an agreement on Transfer options that they need to be validated by someone, or at least that is the intention behind the design. The someone in this case will be the Transfer which is using those options. See down below the concrete implementation for RsyncTransferOptions

enhancements/crane-2.0/dvm-crane-feature-parity/README.md
```diff
@@ -0,0 +1,211 @@
...
+    destinationPVC *v1.PersistentVolumeClaim
+}
+```
+
+_Please note that the existing absolute objects are replaced with a memory reference instead to allow 'nil' checks by the consumer_
+
+With the proposed change above, the Rsync Transfer Server will be updated to use Source PVC reference while the Client will be updated to use Destination PVC reference. Same applies to the Transport server & client as well. I may have used server and client interchangeably, but the point is to update in right places.
+
+
+#### Allowing custom labels and custom Rsync options for Rsync Transfer
```
pranavgaikwad, line 87: Please see a sample here: https://github.com/pranavgaikwad/crane-lib/blob/dvmParity/state_transfer/transfer/rsync/options.go

enhancements/crane-2.0/dvm-crane-feature-parity/README.md
```diff
@@ -0,0 +1,211 @@
...
+	SpecialFiles         bool
+	HardLinks            bool
+	Delete               bool
+	Partial              bool
+	BwLimit              int
+	Info                 string
+}
+
+// RsyncResourceMetadata defines options that let consumers customize different metadata placed on Rsync resources
+type RsyncResourceMetadata struct {
```
alaypatel07, line 117: could we reduce the number of fields in this struct to half? 
```
ReourceMetadata struct {
	Labels                 map[string]string
	Annotations            map[string]string
	OwnerReferences        []metav1.OwnerReference
	SecurityContext        *corev1.SecurityContext
}
```

And then we use two different instances for sourcePods and destinationPods?

enhancements/crane-2.0/dvm-crane-feature-parity/README.md
```diff
@@ -0,0 +1,211 @@
...
+The custom labels and the custom command options can be envisioned as additional configuration options passed by the consumer to the library. They can be consolidated in a struct called `RsyncTransferOptions`:
+
+_options.go_
+
+```golang
+package rsync
+
+type RsyncTransferOptions struct {
+	RsyncCommandOptions
+	RsyncResourceMetadata
```
alaypatel07, line 98: I would expect the ResourceMetadata to be a more generic field shared by all the packages and not be hidden in rsync package. For example, even the routes and services we create in the endpoint package would need those right?

enhancements/crane-2.0/dvm-crane-feature-parity/README.md
pranavgaikwad: @alaypatel07 makes sense. I would assume at the transfer level or maybe even one level above for transport, endpoint and transfer all

pranavgaikwad: @alaypatel07 +1

enhancements/crane-2.0/dvm-crane-feature-parity/README.md
pranavgaikwad: @alaypatel07 yeah, makes sense. maybe at the top level for endpoint, transfer and transport all

enhancements/crane-2.0/dvm-crane-feature-parity/README.md
```diff
@@ -0,0 +1,211 @@
...
+```golang
+type RsyncTransfer struct {
+    [...]
+
+    sourcePVC      *v1.PersistentVolumeClaim
+    destinationPVC *v1.PersistentVolumeClaim
+}
+```
+
+_Please note that the existing absolute objects are replaced with a memory reference instead to allow 'nil' checks by the consumer_
```
shawn-hurley, line 82: If these are unexported, are you suggesting that the nil check would happen from a call to `if pvc = t.SourcePVC(); pvc == nil {`  type of thing?

Would we consider a missing source or dest PVC an error that a user should handle?

enhancements/crane-2.0/dvm-crane-feature-parity/README.md
shawn-hurley: I would suggest we consider how we would add something similar to this, but for a different type of transfer. 

I wonder if for now we can hide the rsync bits if they are at the top level, and add validation that a given config value can be used by a concrete transfer type?

enhancements/crane-2.0/dvm-crane-feature-parity/README.md
```diff
@@ -0,0 +1,211 @@
...
+
+In the first phase, we will implement the custom label options discussed in the previous section in crane-lib. Rsync Retry mechanism depends on using custom labels on the Rsync Pods created. Once this phase finishes, we will simply replace DVM code that creates Rsync Pods with crane-lib function calls. We will keep the logic of maintaining count of retries in DVM. This will allow us to integrate with DVM in the early phase and begin testing. 
+
+In the second phase, we will gradually bring the Rsync retry logic in crane-lib. When this phase finishes, DVM will simply pass on all the Retry related options to the library. At this point, DVMP will still work as before. It will use the label selector information passed to crane-lib to observe progress of Rsync Pods.
+
+In the last phase, we will also bring the DVMP's logic in the library. At the end of this phase, we will simply call crane-lib functions to get progress information of Rsync Pods in DVM.
+
+This enhancement will be updated at every phase to include implementation details.
+
+### Implementation Details/Notes/Constraints [optional]
```
shawn-hurley, line 201: If none, can we remove it?

shawn-hurley (changes requested):
Can you make sure that you delete the empty file from the commit as well?

enhancements/crane-2.0/dvm-crane-feature-parity/README.md
pranavgaikwad: @shawn-hurley yeah, not just consumer but even within the system itself, I think memory addresses are better because the objects get fiilled with default values. I also agree that a nil PVC should trigger an error so that the consumer knows something is wrong

enhancements/crane-2.0/dvm-crane-feature-parity/README.md
```diff
@@ -0,0 +1,211 @@
...
+
+_[transfer.go](https://github.com/konveyor/crane-lib/blob/a5ead85cb740c0325f2797c23a5cbb74f8972220/state_transfer/transfer.go#L14-L33)_
+
+```golang
+type Transfer interface {
+	[...]
+
+	SetSourcePVC(*v1.PersistentVolumeClaim)
+	SourcePVC() *v1.PersistentVolumeClaim
+	SetDestinationPVC(*v1.PersistentVolumeClaim)
```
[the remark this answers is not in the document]
pranavgaikwad: @shawn-hurley my objective here was to just split definitions of source and target PVCs, I copied the existing method signatures. I agree to your point though that these things should only be created once at the time of Transport creation and core properties of a transport should be immutable

enhancements/crane-2.0/dvm-crane-feature-parity/README.md
pranavgaikwad: @shawn-hurley would you explain, I don't think I follow. At the time I was writing this, I couldn't imagine something common that could apply to all transfers

alaypatel07 (approved):
ACK

PR #35, opened 2021-07-14
Add enhancement for CRD v1beta1 deprecation

djwhatle:
Alternative to https://github.com/konveyor/enhancements/pull/39


djwhatle:
@rayfordj @jmontleon maybe you guys are interested in this as well from a rel-eng perspective. Feel free to review.

enhancements/crane-1.x/crd-v1beta1-deprecation/README.md
```diff
@@ -0,0 +1,103 @@
...
+- OCP 3.x to 4.x
+  - Use Crane 1.5.z for OCP 3
+  - Use Crane 1.6+ for OCP 4
+- OCP 3.x to 3.x
+  - Use Crane 1.5.z for OCP 3
+
+
+The thought is that since Crane 1 is mostly feature-complete, we shouldn't have to backport major API changes to Crane 1.5.z. 
+
+For users in the traditional OCP 3.x -> 4.x migration scenario, Crane 1.5.z will esentially just serve as a Velero and ConfigMap installer, and these operator managed resources shouldn't need changes often if Crane 1.x is in maintenence mode.
```
alaypatel07, line 65: does this mean we are relying on the fact that whatever backup velero takes on the source works on the destination?

and a follow-up to the above, would it be fair to ask for an e2e test so that we learn about any breakages as soon as we rebase our branches with newer velero versions?

sseago: @alaypatel07 That's exactly what it means. In the last velero community call, when they suggested that for velero 1.7, v1beta1 CRDs would *not* be supported, I pointed out that this increased the importance of testing backups from older versions and restoring with newer velero versions. Dave suggested that it was already important for this to work, but that while there is *some* testing of this, we probably need more.

sseago: Note that Velero 1.6.3 will probably support both v1beta1 and v1 CRDs, but future versions of Velero (1.7 and later) will probably *not* support this. As a result, if we decide that we want to support Crane 1.6 and later on OCP3.x source clusters, this will mean that either on src clusters we'll need to work with legacy velero versions ( same as proposed above) or that we'll need to create our own v1beta1 CRDs for velero, which would put us even further out of support on OCP3.x than we already are. Velero 1.6 is supported upstream for Kube 1.12 and newer, so we're already out of support, but our only modifications right now are removing CRD validation. For Velero 1.7, if we continue to do this, we won't just be removing validation from CRDs but generating our own.

enhancements/crane-1.x/crd-v1beta1-deprecation/README.md
```diff
@@ -0,0 +1,103 @@
...
+creation-date: 2021-07-14
+last-updated: 2021-07-14
+status: provisional
+---
+
+# CRD v1beta1 API deprecation handling
+
+## Summary
+
+This enhancement proposes maintaining a long-term support MTC/Crane 1.5.z version for users with source clusters on OpenShift 3.7 - 4.2 that will remain compatible with `apiextensions/v1beta1` CRDs. This version of MTC/Crane would receive only minor updates:
```
sseago, line 24: Note that we may want to consider long-term support for 3.7-3.11 only, and not 4.1-4.2. If I'm remembering prior comments on this subject correctly, while 3.x clusters still have some current support available, the same is not the case for 4.1 or 4.2, so we may not need to support those. This means that our src cluster needs would be non-OLM/ansible only, with no requirement to maintain OLM support there.

enhancements/crane-1.x/crd-v1beta1-deprecation/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+- CVE patches
+- Bug fixes
+- Velero updates where possible, though Velero may drop v1beta1 CRD support in future
```
sseago, line 28: Current expectation is that Velero 1.6.z will continue to support v1beta1 for Kube 1.12-1.15 clusters (1.11 and older is already out of support upstream), but that 1.7 will *not* include v1beta1 CRDs for support for Kube 1.15 and older.

enhancements/crane-1.x/crd-v1beta1-deprecation/README.md
```diff
@@ -0,0 +1,103 @@
...
+  - Use Crane 1.6+ for OCP 4
+- OCP 3.x to 3.x
+  - Use Crane 1.5.z for OCP 3
+
+
+The thought is that since Crane 1 is mostly feature-complete, we shouldn't have to backport major API changes to Crane 1.5.z. 
+
+For users in the traditional OCP 3.x -> 4.x migration scenario, Crane 1.5.z will esentially just serve as a Velero and ConfigMap installer, and these operator managed resources shouldn't need changes often if Crane 1.x is in maintenence mode.
+
+We will need to make sure Crane 1.5.z receives matching operator updates. This probably means that some PRs to mig-operator for Crane 1.6+ will also get PR'd to the 1.5.z branch.
```
sseago, line 67: Note thate previously we've only supported matching MTC versions on source, destination, and control clusters. WIth this change, we would need to relax this requirement for Source clusters only. The requirement would then be Source clusters requiring at least 1.5.z (will we need to specify a minimum z here, or will any 1.5.z, including the about-to-release 1.5.0, be sufficient to work with 1.6, 1.7, 1.8, etc. control clusters?), while control and destination clusters may still be required to be the same.

enhancements/crane-1.x/crd-v1beta1-deprecation/README.md
```diff
@@ -0,0 +1,103 @@
...
+      # bundle is installed on the cluster
+      "olm.properties": '[{"type": "olm.maxOpenShiftVersion", "value": "4.8"}\]'
+  ```
+
+
+### Risks and Mitigations
+
+See the "non-goals" section above. The main risk I see with this approach is the additional effort it will require to keep mig-operator changes for Crane 1.5.z maintained in addition to Crane 1.6+. This will likely also introduce some additional test burden.
+
+If we limit the number of new features we add to Crane 1, we should be OK. Changes should be limited to bugfixes, CVEs, and perhaps Velero updates to mitigate this additional workload.
```
sseago, line 98: Note that velero updates to 1.5.z may be limited to our own updates, possibly including backports of CVEs and other bugfixes from velero main to our forked 1.6.z velero stream.

enhancements/crane-1.x/crd-v1beta1-deprecation/README.md
djwhatle: Thanks for the clarifying point Scott. I've heard this from multiple sources: while 3.x versions continue to be supported, early 4.x versions are no longer under support.

enhancements/crane-1.x/crd-v1beta1-deprecation/README.md
```diff
@@ -0,0 +1,103 @@
...
+- Ensure Crane 1.6 is available at time of OCP 4.9 release so that users have a way to unblock their upgrade by switching to Crane 1.6 release stream (will this available Crane version upgrade be obvious to users though?)
+  ```
+  # Example of blocking cluster upgrade with CSV annotation
+  apiVersion: operators.coreos.com/v1alpha1
+  kind: ClusterServiceVersion
+  metadata:
+    annotations:
+      # Prevent cluster upgrades to OpenShift Version 4.9 when this
+      # bundle is installed on the cluster
+      "olm.properties": '[{"type": "olm.maxOpenShiftVersion", "value": "4.8"}\]'
```
rayfordj, line 90:
```suggestion
      "olm.properties": '[{"type": "olm.maxOpenShiftVersion", "value": "4.8"}]'
```
The `\` must not be present here.  It's an artifact of source document creation and actually causes a failure during downstream MTC builds.

djwhatle: Thanks @rayfordj , I just copied this over from the engineering doc OLM put out. I guess this one bit you already :)

rayfordj: Yes-sir, discovered that one from experience.  Copy/Paste, **nope** :sweat_smile:

enhancements/crane-1.x/crd-v1beta1-deprecation/README.md
djwhatle: Yeah, we could modify the validation that happens on the MigPlan and MigCluster so that 1.5.z is allowed as a source cluster when it mismatches with the control cluster.

shawn-hurley (commented):
Do we think it would be worthwhile to get an estimate of the amount of work this would take vs the Alternative? 

Are there pros to the alternative that we should explore?

enhancements/crane-1.x/crd-v1beta1-deprecation/README.md
djwhatle: So it sounds like we'd only be able to guarantee 1.5.z supports Velero 1.6. Further Velero upgrades (Velero 1.7+) could either be:
- Ignored, if we already have all the Velero features we need for Crane
- Adopted by patching Velero to generate v1beta1 CRDs. At this point we'd be using untested Velero and may need to take on additional Velero test burden on our own.

enhancements/crane-1.x/crd-v1beta1-deprecation/README.md
```diff
@@ -0,0 +1,103 @@
...
+Common migration scenarios can be handled with this strategy:
+
+- OCP 3.x to 4.x
+  - Use Crane 1.5.z for OCP 3
+  - Use Crane 1.6+ for OCP 4
+- OCP 3.x to 3.x
+  - Use Crane 1.5.z for OCP 3
+
+
+The thought is that since Crane 1 is mostly feature-complete, we shouldn't have to backport major API changes to Crane 1.5.z.
```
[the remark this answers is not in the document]
djwhatle: I guess that depends on whether there's a requirement for us to support stuff like the new state transfer pieces landing in 1.6 on OCP 3.x. @pranavgaikwad I assume that involved API changes?

enhancements/crane-1.x/crd-v1beta1-deprecation/README.md
djwhatle: > Could we handle some of the mismatches with a plugin? would there be some way to tell us what version the source is?

I think there is a backup schema version number included in backups, but @sseago would be able to confirm that for us.

> does this mean we are relying on the fact that whatever backup velero takes on the source, a different version of velero is able to restore that on the destination?
Scott expressed to me that Velero is interested in providing  _forward compatibility_, meaning newer Velero versions would be able to restore from Backups generated in older versions, but not vice-versa. 

> would it be fair to ask for an e2e test so that we learn about any breakages as soon as we rebase our branches with newer velero versions?
Sounds like a great idea to me 👍

enhancements/crane-1.x/crd-v1beta1-deprecation/README.md
djwhatle: @shawn-hurley one example I can think of: we have a configmap specific to each cluster that is installed by the mig-operator running on that cluster. This includes things like the cluster-specific registry image to use and other OCP version dependent items.

If we added additional entries to this cluster-unique configmap we'd need to update 1.5.z mig-operator to be able to provide these values, _or_ we could add more spec fields to MigCluster to allow for these customizations.

enhancements/crane-1.x/crd-v1beta1-deprecation/README.md
```diff
@@ -0,0 +1,103 @@
...
+#### Generating v1 CRDs
+
+For Crane 1.6+ in mig-controller, patch the [controller-gen command](https://github.com/konveyor/mig-controller/blob/master/Makefile#L60) to generate v1 CRDs instead of v1beta1 ([see example in Velero repo](https://github.com/vmware-tanzu/velero/pull/3614/files#diff-a3e5a2ec9e9650645fbeea6cc08d0c17b3fdc0be893979ca180b329267d12a9cR48-R56)). Copy generated YAML over to mig-operator and run this modified operator against a pair of 3.11 and 4.x clusters. Verify that migrations still run as usual.
+
+#### Blocking cluster upgrades until mig-operator can handle v1 CRDs
+
+The OLM team has outlined steps that we can follow to ensure that the cluster upgrade to OpenShift 4.9 won't proceed until our operator can handle removal of the CRD v1beta1 API.
+
+- Ensure released Crane versions have the required annotation to block upgrades to OCP 4.9 when they are installed.
+- Ensure Crane 1.6 is available at time of OCP 4.9 release so that users have a way to unblock their upgrade by switching to Crane 1.6 release stream (will this available Crane version upgrade be obvious to users though?)
```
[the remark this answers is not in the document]
djwhatle: @rayfordj are you aware of how to set this so OLM will respect it?

enhancements/crane-1.x/crd-v1beta1-deprecation/README.md
djwhatle: @sseago what kind of overall maintenance burden change from the Velero side do you estimate if we went forward with this plan?

djwhatle:
> Do we think it would be worthwhile to get an estimate of the amount of work this would take vs the Alternative?
> Are there pros to the alternative that we should explore?

Fair points, I probably didn't delve into the alternative approach enough in the doc. Will get some more detail down on Monday for that section.

enhancements/crane-1.x/crd-v1beta1-deprecation/README.md
shawn-hurley: In this case, if the value is not available, instead of requiring it we should handle it when it is not there. I think that this is important because this I think is expressly backporting features that we don't want to do.

djwhatle: +1, we should avoid backporting wherever possible. I think it's unavoidable that we'll have to submit bug and CVE patches in both places though right? 

Having to dupe bug fixes to this LTS 1.5.z release might be the most expensive part of this proposed idea.

enhancements/crane-1.x/crd-v1beta1-deprecation/README.md
sseago: @djwhatle I'm looking at a backup CR for my current cluster, and I don't see any annotations which identify the velero version which took the backup, although I do see annotations for the source cluster's k8s version. It might be useful to propose adding this information at backup time (currently, `velero version -n openshift-migration` returns `v1.6.0-konveyor` for server version. This could probably be added as an annotation at the same time the k8s annotations are created. That being said, if the backup doesn't work in the restore cluster with a newer velero, that would be a bug in velero. Whether a plugin would help us work around the bug would depend on how the bug manifested. Without knowing what sort of bug we're talking about, I have no idea whether a plugin would help.

sseago: Depending on the nature of the bug identified, the fix might be not to right a plugin but to fix velero so that it didn't fail -- i.e. whatever broke, change that code in the new velero to properly cope with the old backup.

enhancements/crane-1.x/crd-v1beta1-deprecation/README.md
sseago: @djwhatle features in OCP4.3+ not 4.9+, since kube 1.16(OCP4.3) will continue to be supported.

enhancements/crane-1.x/crd-v1beta1-deprecation/README.md
sseago: @djwhatle for example, if the recently-patched gogo/protobuf CVE had happened post-velero-1.7 and we were still using 1.6 for 3.x source clusters, we would need to backport that fix to a new MTC 1.5.z z-stream (or update to use upstream's backported fix, if they ended up releacing a fix for Velero 1.6.z).

enhancements/crane-1.x/crd-v1beta1-deprecation/README.md
sseago: In the absence of critical bugfixes and CVEs, there wouldn't be a bunch. If we get a BZ for something that's broken in velero on the backup side, we may have to backport those fixes to our 1.5.z builds, since upstream will probably only fix the latest velero. LIkewise, for CVEs, we may have to backport those. It probably doesn't represent a huge burden. The number of BZs we get during a normal release cycle that are pure upstream bugs is usually pretty small, but those that come through are generally blockers. For code that hasn't changed much, it may just be a cherry-pick into our 1.5.z release branches. With refactored code, we may need to re-implement the fix.

enhancements/crane-1.x/crd-v1beta1-deprecation/README.md
djwhatle: I think the implication of this would be that we could skip some rel-eng steps, right @rayfordj ? We wouldn't be pushing anything to OLM for 1.5.z versions, just container images for use with non-OLM install.

enhancements/crane-1.x/crd-v1beta1-deprecation/README.md
djwhatle: Thanks for clarification, updated

enhancements/crane-1.x/crd-v1beta1-deprecation/README.md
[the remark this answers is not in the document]
pranavgaikwad: @djwhatle Yes, good point. There would be a requirement of using State Transfer pieces on 3.x. The API changes are required in Migration APIs mostly. But since State Transfer is only a 1.6 feature, I am unsure how that would work out on older clusters. One good thing about the required API changes is that - they will be used by the controller and need not exist in the source cluster. Let's chat about how that will work out in detail.

enhancements/crane-1.x/crd-v1beta1-deprecation/README.md
rayfordj: > Not sure if it would be worthwhile to skip those steps or not.

In the scheme of things, I doubt we'll see any real benefit of eliminating that in terms of MTC build effort at this point.  Plus, I believe it's been indicated we'll need to carry support for migrating _**from**_ 4.2 source clusters.

djwhatle:
Merging this as implementable, targeted for development in Crane 1.6.0 timeframe.

We will re-examine the [PR 39 remote install enhancement](https://github.com/konveyor/enhancements/pull/39) again in Crane 1.7.0 timeframe when we have more runway to develop, then update docs, tests, etc.

PR #37, opened 2021-07-16
Add must-gather from Forklift UI enhancement

aufi:
Proposing enhancement allowing to execute must-gather from Forklift UI.

Related to must-gather backend enhancement proposal https://github.com/konveyor/enhancements/pull/36

### DIFF
--- a/enhancements/forklift-2.2/must-gather-from-ui/README.md
+++ b/enhancements/forklift-2.2/must-gather-from-ui/README.md
@@ -141,12 +141,24 @@ Demo video: https://youtu.be/b5YTxyp6WyY
 
 ### Risks and Mitigations
 
+#### UX/UI
+
 The UI needs to be intuitive for user, will be discussed with UX team.
+
+#### Security
 
 Must-gather itself needs ```cluster-admin``` role. It could provide sensitive information about the cluster so needs to be authorized by OpenShift RBAC.
 
 - HTTP API will require ```oauth``` token and verify if the token provides ```cluster-admin``` access to the cluster
 - backend service needs a service account token/kubeconfig that will be ona PV mounted by Operator (in similar way as for other migration-toolkit components)
+
+The must-gather wrapper service is an administration tool serving users with oauth token for ```cluster-admin``` role, so it does not accept requests from anonymous or unknown clients, so e.g. accepting arbitrary ```image``` as API parameter seems reasonable.
+
+#### Resources consumption and limits
+
+Compute power - the must-gather wrapper service is not a bottle-neck regarding to CPU. The gathering itself runs in a pod started from the must-gather image separate namespace in OpenShift outside the pod where the wrapper runs, so there doesn't seem to be need for amount of must-gather executions running in paralel.
+
+Disk space - result archives are temporary stored directly on the filesystem of the pod where must-gather wrapper service is running. The archive size is unknown before the must-gather archive is downloaded to the must-gather wrapper pod. Reasonable maximum disk limit for the must-gather wraper service pod seems to me to be 10GB or more, however I do believe that the limit should be just space that OpenShift node could provide to the pod with must-gather, no must-gather wrapper service limits.
 
 ## Design Details
 
--- a/enhancements/forklift-2.2/must-gather-from-ui/README.md
+++ b/enhancements/forklift-2.2/must-gather-from-ui/README.md
@@ -1,0 +1,72 @@
+---
+title: must-gather-from-ui
+authors:
+  - "@aufi" / maufart
+reviewers:
+  - TBD
+approvers:
+  - TBD
+creation-date: 2021-07-16
+last-updated: 2021-07-16
+status: provisional
+see-also:
+  - "/enhancements/must-gather-rest-service/README.md"
+  - "https://bugzilla.redhat.com/show_bug.cgi?id=1944402"
+---
+
+# Must-gather from UI
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] User-facing documentation is created
+
+## Summary
+
+It should be possible to use OpenShift must-gather feature directly from Forklift migration UI.
+
+## Motivation
+
+When a VM migration fails, Forklift user needs quick access to migration debugging informations and other details provided by must-gather feature. It should help to identify where is a problem and to make a successful migration once the problem is resolved.
+
+To make it easy for user, it should be supported to trigger must-gather directly from UI and don't have to switch to CLI.
+
+### Goals
+
+User should be able to execute must-gather from Forklift UI and get archive with results.
+
+### Non-Goals
+
+This enhancement doesn't cover backend or must-gather execution itself, just provide features of must-gather-service enhancement in UI.
+
+## Proposal
+
+The Forklift UI should allow user to execute OpenShift must-gather and provide the result archive to user.
+
+The must-gather should be possible to execute for
+- Plan
+- VM
+
+A Command parameter of must-gather to specify targets for gathering (e.g. ```PLAN=plan1 /usr/bin/targeted```) will be prepared on UI side into a string passed to must-gather-service API call.
+
+### Risks and Mitigations
+
+The UI needs to be intuitive for user, will be discussed with UX team.
+
+## Design Details
+
+### Test Plan
+
+After a VM migration from VMware or RHV to OpenShift Virtualization, click on button TBD and wait until an archive is downloaded. It should not take more than 5 minutes. The archive should contain following directory structure:
+
+- TBD
+
+## Drawbacks
+
+There is existing feature in OpenShift command line, so user could use CLI.
+
+## Alternatives
+
+User could switch to command line and do the same with existing set of features. In some projects (Crane) there is quite detailed way to get details on Migration directly in UI.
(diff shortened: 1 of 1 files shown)


jwmatthews:
I like the idea of making it easier to obtain the must-gather for end-user.

@aufi do you see a path forward of implementing this in a reusable approach so other projects could share this and tweak for their integration?

@eriknelson @ibolton336 @djwhatle @shawn-hurley please take a look and consider if this would be useful to consider for Crane.  

Assume we'd need to have:

- sane way to limit the size of the must-gather so the user is not downloading a multi-hundred MB+ archive
- need to account for the generation of must-gather taking several minutes+, I think we've seen 10+ min times to generate in past

aufi:
> @aufi do you see a path forward of implementing this in a reusable approach so other projects could share this and tweak for their integration? (Updated comment: Just saw you already mentioned plans of this being shared with other projects via #36)

Correct, I'm planning to have the same must-gather-service as a backend (just with configuration adjustments) and the UI part specific per-project. The backend can be configured e.g. for DEFAULT_IMAGE, so it should be suitable for everyone, https://github.com/aufi/must-gather-rest-wrapper#available-api-parameters

> Assume we'd need to have:
> 
> * sane way to limit the size of the must-gather so the user is not downloading a multi-hundred MB+ archive

Good point, I think bigger files (hundreds MB) could be quite common (also in OpenShift must-gather outside migration toolkit) and it could be useful. We could add information about size to API, so when must-gather has finished and archive is ready for dowload, the archive size would be part of API respose to let UI or User decide if the archive download make actually sense.

> * need to account for the generation of must-gather taking several minutes+, I think we've seen 10+ min times to generate in past

Right, the must-gather execution usually takes from few minutes to few tens of minutes.

The must-gather execution/archive generation on backend is asynchronous, so UI triggers must-gather with POST request and then periodically queries the API until must-gather execution status gets to "completed", then it can serve the archive file to user. But user needs to wait to ge able download result until it finishes.

enhancements/forklift-2.2/must-gather-from-ui/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Summary
+
+It should be possible to use OpenShift must-gather feature directly from Forklift migration UI.
+
+## Motivation
+
+When a VM migration fails, Forklift user needs quick access to migration debugging informations and other details provided by must-gather feature. It should help to identify where is a problem and to make a successful migration once the problem is resolved.
+
+To make it easy for user, it should be supported to trigger must-gather directly from UI and don't have to switch to CLI.
```
djwhatle, line 34: I like this idea of making it obvious to the user that this is an option from the UI. The approach we've taken for some things like this in Crane has been to provide the CLI command that the user can execute on their own (e.g. provide the logs command we recommend the user run).

Since must-gather was designed for CLI in the first place and provides all output on progress through streaming CLI logs (and given that we have no way to offer progress otherwise on the several-minutes long process) I almost wonder if giving the user a pointer to the CLI command would be a more empowering UX that lets the user have more direct access to info.

As a counter point to my above argument, if Forklift is designed so the user rarely has to drop to CLI and this would be a jarring experience for them, I could see the argument for the UI-driven must-gather button being less scary to the user. My main gripe if I was the user here would be complete lack of progress reporting on a rather long-running process.

shawn-hurley (commented):
I would personally add the backend and this together. I wonder if we can consider some sort of type script module for executing the backend that anyone using typescript and has the re-usable backend implementation could use?

enhancements/forklift-2.2/must-gather-from-ui/README.md
ibolton336: +1 IMO it sounds like the flow here would be something best managed by node or something on the middle/backend since it would involve large streams of data and user specific information that could be authenticated with stored session cookies.

ibolton336:
> I would personally add the backend and this together. I wonder if we can consider some sort of type script module for executing the backend that anyone using typescript and has the re-usable backend implementation could use?

I like this idea! Would be really good candidate for https://github.com/konveyor/lib-ui if we can pull something together.

enhancements/forklift-2.2/must-gather-from-ui/README.md
aufi: > I like this idea of making it obvious to the user that this is an option from the UI. The approach we've taken for some things like this in Crane has been to provide a "call to action" in the UI with a "Click this button to copy command" for the CLI command that the user can execute on their own (e.g. provide the logs command we recommend the user run). ...<snip>... My main gripe if I was the user here would be complete lack of progress reporting on a rather long-running process.

Thanks for the review! This is clearly an alternative to CLI, as mentioned, this should fit well to Forklift. The Crane's approach is little bit different (execute command by user).

To provide updates during must-gather execution process, I'm looking on command streamed output capture and updating it in the must-gather execution JSON, so when API returns execution that have started, there will be the latest update of actual command output in response.

> I would also be concerned about how this info would be stored and accessed (auth?) if it's run by the wrapper service, given that info gathered could contain sensitive info depending on how you wrote your must-gather. Would we require a PV allocated to this wrapper service? How long will must-gather dumps be kept in storage?

The authentication - initial plan was that the pod running the service will have mounted PV with auto token. I'm not sure on oauth part passed from UI. I'm going o investigate more and provide update later.

Must-gather dumps are planned to be in the container (not PV) and be regually cleared (files&sqlite records). This is already implemented and can be configured with CLEANUP_MAX_AGE option https://github.com/aufi/must-gather-rest-wrapper#configuration

djwhatle: > To provide updates during must-gather execution process, I'm looking on command streamed output capture and updating it in the must-gather execution JSON, so when API returns execution that have started, there will be the latest update of actual command output in response.

Oh that sounds very cool! +1

> The authentication - initial plan was that the pod running the service will have mounted PV with auth token. I'm not sure on oauth part passed from UI. I'm going o investigate more and provide update later.

Sounds good,  just need to make sure the must-gather dumps are protected if they contain, for instance, VM management credentials.

> Must-gather dumps are planned to be in the container (not PV) 

Not sure I'm following, I don't think you can just write data "in the container" without some sort of volume mount, but maybe you're just using the filesystem of whichever node the rest-wrapper is running on?

> and be regually cleared (files&sqlite records). This has been implemented already and it can be configured with CLEANUP_MAX_AGE option

Nice, smart usage of sqlite features there  👍

aufi: > > Must-gather dumps are planned to be in the container (not PV)
> 
> Not sure I'm following, I don't think you can just write data "in the container" without some sort of volume mount, but maybe you're just using the filesystem of whichever node the rest-wrapper is running on?
> 

Thanks for feedback, I meant a filesystem where the container runs.

aufi:
Thanks for feedback provided so far! Based on suggestions I merged this backend enhancement with a Forklift UI enhancement #36. The discussion can continue here.

aufi:
Added auth by bearer token provided via API, no serviceaccount or secret is needed to be passed to the must-gather-api backend pod deployment.

enhancements/forklift-2.2/must-gather-from-ui/README.md
```diff
@@ -147,12 +148,13 @@ The UI needs to be intuitive for user, will be discussed with UX team.
...
+Must-gather itself needs ```cluster-admin``` role since it could provide sensitive information about the cluster, that is OpenShift design which needs to followed.
 
-- HTTP API will require ```oauth``` token and verify if the token provides ```cluster-admin``` access to the cluster
-- backend service needs a service account token/kubeconfig that will be ona PV mounted by Operator (in similar way as for other migration-toolkit components)
+- Must-gather API requires ```oauth bearer``` token
+- The oc adm must-gather command execution uses the provided bearer token to autheticate against OpenShift cluster
+- No ServiceAccount or other secret is used
 
-The must-gather wrapper service is an administration tool serving users with oauth token for ```cluster-admin``` role, so it does not accept requests from anonymous or unknown clients, so e.g. accepting arbitrary ```image``` as API parameter seems reasonable.
+To ensure that the must-gather result archive is provided to the right user, a hash of the provided bearer token is stored for each must-gather execution and API filters must-gather results by the bearer token hash.
```
shawn-hurley, line 157: Bearer tokens themselves are not the identity of the user IIUC. Do we have any other alternatives?

aufi: Good point on naming, correct; it is more identification of User's temporary session, than persisted User account. The con is that if user would re-login, the must-gather execution from their previous login became not be visible anymore.

I was thinking about translating bearer token to username (like oc whoami output), but that requires oc login call for each request even for get/list must-gather executions, which I wanted to avoid. Let me check more options by tomorrow.

aufi: It seems I don't have anything better that "translating" bearer token to username that it belongs to and keep that username (or its hash) in the must-gather execution record. The bearer token -> username would be cached most likely to avoid maby login requests to OpenShift API.

Shawn, do you think this is an acceptable solution (or would you recommend something else)? Thanks!

shawn-hurley: If you have access to the user that is requesting a resource could you just do a SubjectAccessReview to the specific resource to see if they have access rather than caching it on the object itself?

aufi: Yes, thanks, I can do something like Jeff did in https://github.com/konveyor/forklift-controller/blob/main/pkg/controller/provider/web/base/auth.go

jwmatthews:
Anyone know if further work is planned on this writeup, is this in a state to merge, etc?

jwmatthews:
@aufi are we good with merging this, are there any other updates you would like to make?

aufi:
I'm good merging this enhancement spec!

PR #39, opened 2021-07-21
Add alternate (remote install) enhancement for CRD v1beta1 removal

djwhatle:
Alternative to https://github.com/konveyor/enhancements/pull/35


enhancements/crane-1.x/crd-v1beta1-deprecation-remoteinstall/README.md
```diff
@@ -0,0 +1,150 @@
...
+- Develop a strategy that can be used in the future when further Kubernetes API deprecations inevitably happen
+
+### Non-Goals
+
+- Introduce high burden of backporting fixes 
+- Block introduction of new features/bugfixes into Crane 1.x
+
+## Proposal
+
+Common migration scenarios can be handled with this strategy:
```
shawn-hurley, line 66: I think we should add the restricted network migration here as well

enhancements/crane-1.x/crd-v1beta1-deprecation-remoteinstall/README.md
```diff
@@ -0,0 +1,150 @@
...
+  - No longer possible
+
+
+### Implementation Details/Notes/Constraints
+
+#### Migration of Config that previously lived in MigrationController CR to the MigCluster CR
+
+Config info previously in the MigrationController CR spec on the remote cluster can be moved to an equivalent ConfigMap on the control cluster that would be referenced from the MigCluster  CR.
+
+Alternatively we could include this config info directly on the MigCluster CR. This would require us to worry about schema changes, whereas the ConfigMap approach would be schema-less similar to how the MigrationConroller spec field is today.
```
shawn-hurley, line 89: Is there a potential versioning upgrade/downgrade problem here?

I think we need to add an upgrade downgrade section to talk about this change

enhancements/crane-1.x/crd-v1beta1-deprecation-remoteinstall/README.md
```diff
@@ -0,0 +1,150 @@
...
+
+Even on the control cluster, we should move all Velero, ConfigMap, and mig-log-reader setup tasks to the MigCluster controller so that we only have to maintain one set of templates, and we aren't doing half of our Velero installs in mig-operator, half in mig-controller.
+
+#### Achieving functionality equivalent to mig-operator for Velero dependency install tasks
+
+We may be able to make use of Jinja or Go templates to achieve remote cluster dependency installs.
+
+#### Requirement for highly-permissioned remote cluster SA token
+
+Currently, mig-operator sets up the `migration-controller` SA that we use to perform actions on the remote cluster. We would need the cluster admin of that cluster to create an SA token for us instead of having mig-operator do this. This should be a trivial difference.
```
shawn-hurley, line 101: Would we have a published role that they would need to use?

shawn-hurley (commented):
Thank you for having both proposals created! this is making it a really open discussion that I think is great!

enhancements/crane-1.x/crd-v1beta1-deprecation-remoteinstall/README.md
djwhatle: @shawn-hurley this role is used today, and could be handed to the user to create on their cluster: https://github.com/konveyor/mig-operator/blob/5a8ce7209b9a62b8afa66fafb877ad0a5d0fbb85/roles/migrationcontroller/templates/mig_rbac.yml.j2

enhancements/crane-1.x/crd-v1beta1-deprecation-remoteinstall/README.md
djwhatle: I mention some upgrade considerations in the `Risks and Mitigations` section, I'll create an upgrade/downgrade section and consolidate notes there.

enhancements/crane-1.x/crd-v1beta1-deprecation-remoteinstall/README.md
djwhatle: +1, will add.

djwhatle: @shawn-hurley section added.

enhancements/crane-1.x/crd-v1beta1-deprecation-remoteinstall/README.md
djwhatle: @shawn-hurley section added.

enhancements/crane-1.x/crd-v1beta1-deprecation-remoteinstall/README.md
```diff
@@ -0,0 +1,229 @@
...
+
+- Introduce high burden of backporting fixes 
+- Block introduction of new features/bugfixes into Crane 1.x
+
+## Proposal
+
+Discontinue Crane installation on OCP 3.x with mig-operator. Write remote install functionality to get Crane deps on all remote clusters.
+
+
+### Common migration scenarios affected by this enhancement:
```
djwhatle, line 69: @eriknelson @jwmatthews before proceeding with this plan for v1beta1 CRD deprecation, I'd like to make sure you guys are OK with these effects.

jwmatthews: We need to ensure we have a path forward for on-premise -> cloud.  I am assuming at moment that a worst case situation is the path forward is for the user to open a VPN/VPC connection between on-prem/cloud.  The burden of doing that is not desirable, but I think it would be a possibility.   I would like to discuss this with @tsanders-rh and @mberube99 next week to get their thoughts if this burden is acceptable or too great for our users.

Let's ensure we do continue to investigate if we can provide a temporary alternative with the stunnel approach or something else.   Let's not lose that as desired when we work this story.

enhancements/crane-1.x/crd-v1beta1-deprecation-remoteinstall/README.md
```diff
@@ -0,0 +1,229 @@
...
+
+- **OCP 3.9+ to 4.6+**
+  - OCP 4.6+: Install Crane 1.6+ 
+  - OCP 3.9+: Install Crane dependencies remotely
```
jwmatthews, line 73: @djwhatle 
Is this truly 3.9+ and not 3.7+?  If this is really 3.9, please help me learn what we are lacking in 3.7 to not support it.

enhancements/crane-1.x/crd-v1beta1-deprecation-remoteinstall/README.md
```diff
@@ -0,0 +1,229 @@
...
+
+- **OCP 4.6+ to 4.6+**
+  - OCP 4.6+: Install Crane 1.6+ 
+  - OCP 4.6+: Install Crane dependencies remotely
+
+- **OCP 4.5- to 4.6+**
+  - OCP 4.6+: Install Crane 1.6+ 
+  - OCP 4.5-: Install Crane dependencies remotely
+
+- **OCP 4.5- to 4.5-**
```
jwmatthews, line 83: I don't see this as a problem, we need to ensure we can migrate to the latest OCP releases, probably will need to support the latest 4.x EUS and latest ~2-3 releases in general.

enhancements/crane-1.x/crd-v1beta1-deprecation-remoteinstall/README.md
```diff
@@ -0,0 +1,229 @@
...
+  - No longer possible 
+
+- **OCP 3.9+ to 3.9+**
```
jwmatthews, line 86: I don't see this as a problem, sounds like a reasonable trade off to lose

enhancements/crane-1.x/crd-v1beta1-deprecation-remoteinstall/README.md
jwmatthews: @djwhatle I'm not sure if "restricted network migration" refers to the on-prem -> cloud case where we only have network access in one direction or if it refers to the "Disconnected" scenario where a customer is running both clusters in a restricted environment with limited outside connectivity.

Please consider if we need to worry about the "Disconnected" scenario, how would we install the images for the OCP3 cluster if they need to follow the mirroring procedures for getting images into the local network.  I think it'd be worth calling that out in this doc.

enhancements/crane-1.x/crd-v1beta1-deprecation-remoteinstall/README.md
alaypatel07: would be worth going through those. The particular example cited is a developer backdoor, not intended for users to muck with, in which cases when needed we should be able to jump on customer calls and edit the cluster role on remote cluster after creation.

enhancements/crane-1.x/crd-v1beta1-deprecation-remoteinstall/README.md
```diff
@@ -0,0 +1,229 @@
...
+- _Risk_: it's not clear what the upgrade path would look like for users jumping from Crane 1.5 to Crane 1.6, given that all of their previous Config data on the MigrationController CR would need to land in its new home in Crane 1.6. We would need to develop a strategy for migrating users Crane config to the new format
+  - _Mitigation_: if users upgrade their OCP 4.x clusters to the latest version, we could remotely read the config out of the MigrationController CR and copy it to the new location on the MigCluster spec (or wherever we decide to store config data)
+
+## Design Details
+
+### Upgrade / Downgrade Strategy
+
+We can provide automatic upgrades to this new method of managing remote cluster Cranes deps, however the downgrade process will be manual.
+
+#### Automatic Upgrade steps (Crane 1.5 to 1.6+):
```
alaypatel07, line 211: Deleting mig-operator.yaml deletes the entire namespace, i.e. including the service account that has gives the control cluster access to this remote cluster. If we attempt that before step 7 which needs access to the remote cluster from control cluster, I wonder if this will work. 

Give the above I might have some suggestions on this upgrade path:

1. copy all the config found in migrationcontroller.spec on the remove cluster to local configmap
2. scale down the mig-operator deployment to 0, that will stop mig-operator pod reconciling the migrationcontroller CR
3. start the mig cluster re-provisioning of velero, restic and mig-log-reader

I think the advantage of the above would be that we will not try to delete everything and restart, just update the things that are needed in place from the control cluster. We will stop the mig-operator pod that will stop all reconciling and the migraitoncontroller CR will essentially be dormant.

alaypatel07 (approved):
Added some nits. I also want to highlight one more thing, we don't have remote watches as of now. When reconciling from the control cluster do we plan on introducing them back or will this happen on a polling basis?

Apart from those details, the overall framework set up in the enhancements looks good to me.

enhancements/crane-1.x/crd-v1beta1-deprecation-remoteinstall/README.md
djwhatle: That's a typo, will fix.

enhancements/crane-1.x/crd-v1beta1-deprecation-remoteinstall/README.md
djwhatle: @pranavgaikwad and I will sync on Monday to discuss possibilities

enhancements/crane-1.x/crd-v1beta1-deprecation-remoteinstall/README.md
djwhatle: ACK, good call out

enhancements/crane-1.x/crd-v1beta1-deprecation-remoteinstall/README.md
djwhatle: @alaypatel07 is the idea to just leave the MigrationController CR on the source cluster in a dormant state permanently until the user intervenes?

I like the idea of a scale down instead of deletion.

djwhatle: @alaypatel07 does OLM not enforce 1 replica of mig-operator?

djwhatle: @alaypatel07 integrated your suggestion.

alaypatel07: AFAIK OLM does not enforce 1 replica, would be good to fact-check here though.

enhancements/crane-1.x/crd-v1beta1-deprecation-remoteinstall/README.md
djwhatle: @jwmatthews added section on disconnected, will sync with @jmontleon on Monday to get more details and fill in gaps.

enhancements/crane-1.x/crd-v1beta1-deprecation-remoteinstall/README.md
alaypatel07: @djwhatle also you will have to keep migrationcontroller CR to keep the connection to the cluster. The MigrationController CR has owner reference on the migration-controller SA, if you delete that, SA will be deleted and the token will be invalidated and the control cluster will lose access?

djwhatle:
> I also want to highlight one more thing, we don't have remote watches as of now. When reconciling from the control cluster do we plan on introducing them back or will this happen on a polling basis?

Given this limitation (we can't watch on a remote cluster):

- Option 1: assume user won't delete Deployments and ConfigMaps we put into `openshift-migration`. This seems like a reasonable thing to expect of users IMO.
   - If they do, require them to recreate their MigCluster to get their provisioned items back.

- **Option 2:  Run periodic checks to see if everything needed on remote end is still installed and in correct state.** If things don't look correct, enqueue a MigCluster reconcile. Putting this poll for correctness outside the main  reconcile  will avoid triggering reconcile of ~500 plans when not needed.
    - Example:  [RemoteClusterSource](https://github.com/konveyor/mig-controller/blob/master/pkg/controller/migcluster/migcluster_controller.go#L79-L87)

enhancements/crane-1.x/crd-v1beta1-deprecation-remoteinstall/README.md
djwhatle: > if you delete that, SA will be deleted and the token will be invalidated and the control cluster will lose access?

@alaypatel07 that's a good point, so we will need to keep that around then.

> AFAIK OLM does not enforce 1 replica, would be good to fact-check here though.

I just verified that OLM allows you to scale mig-operator to 0 replicas.

alaypatel07:
> This seems like a reasonable thing to expect of users IMO. There's no difference in this vs. expecting the user to not delete their MigrationController CR I don't think.

@djwhatle Fair point. In fact, I would argue this is safer because the user will have to log into the source cluster for anything. It helps to avoid unintentional mistakes.

enhancements/crane-1.x/crd-v1beta1-deprecation-remoteinstall/README.md
djwhatle: @jwmatthews are any specifics available on the exact network configurations we must support? That would help with determination of whether we can work out a communication method.

eriknelson (changes requested):

enhancements/crane-1.x/crd-v1beta1-deprecation-remoteinstall/README.md
```diff
@@ -0,0 +1,244 @@
...
+
+- __1__: We will need two different sets of velero images: Velero 1.6.z, Velero 1.7+. Each MTC release will need two Velero release branches (release-1.y.z
+and release-1.y.z-velero-1.6). We will need to test both branches.
+  - velero
+  - velero-restic-restore-helper
+  - velero-plugin-for-aws
+  - velero-plugin-for-gcp
+  - velero-plugin-for-microsoft-azure
+ 
+- __2__. Velero 1.6.z is planned to be be out-of-support upstream, which means that we may need to backport CVE and certain bugfixes (essentially blocker bugs that affect backup creation), since upstream CVE and bugfixes will only go to the
```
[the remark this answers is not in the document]
sseago: @eriknelson So in some cases, we get notified of CVEs affecting our velero builds before upstream fixes them -- that's what happened with the 1.6.2 CVEs. But there is a risk for CVEs that are reported directly to upstream that we don't know about -- for those, it's possible that velero 1.7+ for OCP4 clusters get fixed but our OCP3 1.6.3 velero release does not. It may be that "read through Velero release notes and investigate any CVEs mentioned" needs to be part of our 1.7+ rebase process.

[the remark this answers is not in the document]
sseago: It's probably worth noting that when the up-to-date velero updates to a new restic version, the same concerns apply there -- we need to investigate the restic changes and determine whether we want to either upgrade restic for our 1.6 velero stream or cherry-pick specific security fixes onto our restic fork of 0.12.0.

djwhatle:
Merging this as deferred, since we will plan to look at remote install work again in Crane 1.7.0 timeframe. Plan for now is to go with https://github.com/konveyor/enhancements/pull/35 for Crane 1.6.0.

PR #40, opened 2021-07-22
Adding enhancement for high level datamover design

shawn-hurley:
/cc @dymurray


enhancements/oadp/datamover-design/README.md
```diff
@@ -0,0 +1,110 @@
...
+1. Creating an data mover for production use
+2. Define interfaces/types and structures in this enhacement. These should be defined in follow on enhancement.
+
+## Proposal
+
+![DataMover Visual](./Datamover-Design.png)
+
+There are 3 new pieces here but first lets discuss the pieces that are being re-used. 
+
+We will create Velero plugin for [Volume Snapshotter and Backup Item Action](https://velero.io/docs/v1.6/custom-plugins/#plugin-kinds). This will allow us to determine when a Volume is being backed up and complete the necessary actions to move the data to the correct place. These are well known with lots of prior art. We should use the [CSI-plugin](https://github.com/vmware-tanzu/velero-plugin-for-csi) and the [restic](https://github.com/vmware-tanzu/velero/blob/5bd70fd8eef316d220317245e46dc6016c348dce/pkg/restic/repository_manager.go#L131) plugin as an example.
```
dymurray, line 53: I would mention cloud provider snapshots as well here

enhancements/oadp/datamover-design/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+### Goals
+
+List the specific goals of the proposal. How will we know that this has succeeded?
+
+1. An overall design, that will give a high level vision for how to allow for velero to interact with 
+
+### Non-Goals
+
+1. Creating an data mover for production use
```
dymurray, line 44: NIT: a*

enhancements/oadp/datamover-design/README.md
```diff
@@ -0,0 +1,110 @@
...
+## Proposal
+
+![DataMover Visual](./Datamover-Design.png)
+
+There are 3 new pieces here but first lets discuss the pieces that are being re-used. 
+
+We will create Velero plugin for [Volume Snapshotter and Backup Item Action](https://velero.io/docs/v1.6/custom-plugins/#plugin-kinds). This will allow us to determine when a Volume is being backed up and complete the necessary actions to move the data to the correct place. These are well known with lots of prior art. We should use the [CSI-plugin](https://github.com/vmware-tanzu/velero-plugin-for-csi) and the [restic](https://github.com/vmware-tanzu/velero/blob/5bd70fd8eef316d220317245e46dc6016c348dce/pkg/restic/repository_manager.go#L131) plugin as an example. 
+The 
+
+We will need to create a Data Mover Controller. This controlle will watch a DataMover CRD that will be where the status of moving data will be published. This Controller's only responsiblity to delegate via GRPC to the correct provided for Data Mover.
```
dymurray, line 56: NIT: `this controller`

NIT: "This controller will watch a DataMover CRD where the status of moving data will be published".

NIT: "delegate via GRPC to the correct provider for Data Mover"

enhancements/oadp/datamover-design/README.md
```diff
@@ -0,0 +1,110 @@
...
+![DataMover Visual](./Datamover-Design.png)
+
+There are 3 new pieces here but first lets discuss the pieces that are being re-used. 
+
+We will create Velero plugin for [Volume Snapshotter and Backup Item Action](https://velero.io/docs/v1.6/custom-plugins/#plugin-kinds). This will allow us to determine when a Volume is being backed up and complete the necessary actions to move the data to the correct place. These are well known with lots of prior art. We should use the [CSI-plugin](https://github.com/vmware-tanzu/velero-plugin-for-csi) and the [restic](https://github.com/vmware-tanzu/velero/blob/5bd70fd8eef316d220317245e46dc6016c348dce/pkg/restic/repository_manager.go#L131) plugin as an example. 
+The 
+
+We will need to create a Data Mover Controller. This controlle will watch a DataMover CRD that will be where the status of moving data will be published. This Controller's only responsiblity to delegate via GRPC to the correct provided for Data Mover. 
+
+A vendor will be able to supply a data mover, as long as it conforms to a certain interface, to provide moving the data for that particular set of data. For a default implementation we may want to consider how this will work.
```
dymurray, line 58: NIT: remove comma

enhancements/oadp/datamover-design/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+Another usecase that we should consider for the "DatamMver" CRD is to handle moving VolumeSnapshots to a different location.
```
dymurray, line 60: NIT: `DataMover`

enhancements/oadp/datamover-design/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+As a data storage provider, I would like to provide my own data mover to integrate with OADP as the standard backup and restore tool. 
+
+#### Story 2
+
+As a user, I would like to move my snapshot to a different location off cluster for backup.
+
+### Implementation Details/Notes/Constraints
+
+As there is no implementation details as these will be discussed in future enhacements.
```
dymurray, line 75: NIT: Remove first `As`

enhancements/oadp/datamover-design/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+### Risks and Mitigations
+
+There will be issues with security that we will need to consider. These should be focused on how we use RBAC. For Story 2 we will have to be sure tho verify that the user has access to the PVC/VolumeSnapshot that they are asking to move.
```
dymurray, line 79: NIT: `tho` -> `to`.

>For Story 2 we will have to be sure tho verify that the user has access to the PVC/VolumeSnapshot that they are asking to move. 

I'm curious how we achieve this... can we operate under the assumption that if a user creates a DataMover CR in a namespace that they have access to all PVCs in this namespace? We won't have any information about the originating creator of the DataMover CR right?

enhancements/oadp/datamover-design/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+### Implementation Details/Notes/Constraints
+
+As there is no implementation details as these will be discussed in future enhacements.
+
+### Risks and Mitigations
+
+There will be issues with security that we will need to consider. These should be focused on how we use RBAC. For Story 2 we will have to be sure tho verify that the user has access to the PVC/VolumeSnapshot that they are asking to move. 
+
+Another security concern is the permissions that would be required for a specific Data mover plugin to have access to the underlying data storage or location.
```
dymurray, line 81: +1

enhancements/oadp/datamover-design/README.md
```diff
@@ -0,0 +1,110 @@
...
+We have create a plugin for the openshift resources. 
+
+The CSI velero plugin already exists and can be used as a resource. 
+
+VolumeProvisioner's and VolumePopulators from the CSI addon are also good example of the concepts.
+
+
+## Drawbacks
+
+The major drawback with this approach is that it is another interface that the storage providers will need to conform to. To get buy in for this approach we will need to work closely with the vendors to make sure that the interface gives them what they need to complete the actions. This could take a lot of work to do well and could take a long time to complete.
```
dymurray, line 106: This is my concern as well. In fact you find that not a lot of people write third party Velero plugins that are extensively maintained outside of the core velero team.

enhancements/oadp/datamover-design/README.md
```diff
@@ -0,0 +1,110 @@
...
+1. Do we want to try and get our GRPC endpoints into CSI or Openstorage API or both?
+
+## Summary
+
+Creating a common interface for datamovers, with a shared controller for that interface and using GRPC for datamover plugins. We will also create a velero plugin that will manage this new CRD when PVC's or PV's are attempted to be backed up. 
+
+
+## Motivation
+
+As more and more vendors use OADP as a common interface for interacting with backup and restore on an openshift cluster, these vendors want a deeper integration into how to move data. Right now, the primary methods that OADP moves data is through restic backup of a volume to an object bucket or via CSI snapshots that are off cluster. Some vendors don't fit into either of these buckets and they would like to easily provide their own version of a data mover to fully support user backups.
```
dymurray, line 34: I would add another primary motivation for data mover (you touched on this) is that CSI snapshots don't provide real DR utility as they exist on the cluster we are taking a backup of. A generic data mover for CSI snapshots has been heavily requested

enhancements/oadp/datamover-design/README.md
shawn-hurley: We will use a validating webhook, to check that the user has access to the PVC that is being requested. 

I don't think that we should escalate privileges like that if we don't have to. I don't think we have to. 

For Backup requests, we will have to escalate, but if you just use the DataMover CRD then we can be better.

enhancements/oadp/datamover-design/README.md
shawn-hurley: Why would we need to move cloud CSI snapshots? These snapshots are already stored off the cluster? I am not understanding the use case for moving the data in this case.

enhancements/oadp/datamover-design/README.md
shawn-hurley: Considering that this is a feature that users have asked for from the vendor side, I am thinking that this might be achievable.

enhancements/oadp/datamover-design/README.md
shawn-hurley: I might be wrong here but here is my understanding.

I thought that currently, a cloud provider snapshot was stored off the cluster, in the cloud by that provider. While the Resources pointing to it are in fact on the cluster the data itself is off the cluster and can be used to restore from another cluster. 

A storage system that does store the data off the system is why we need to create a data mover.

jwmatthews: I think of the ocs4 case where a CSI snapshot has the data on the cluster with ocs, so for D/R we need to copy that data off cluster in-case cluster is lost.

enhancements/oadp/datamover-design/README.md
```diff
@@ -0,0 +1,109 @@
...
+VolumeProvisioner's and VolumePopulators from the CSI addon are also good example of the concepts.
+
+
+## Drawbacks
+
+The major drawback with this approach is that it is another interface that the storage providers will need to conform to. To get buy in for this approach we will need to work closely with the vendors to make sure that the interface gives them what they need to complete the actions. This could take a lot of work to do well and could take a long time to complete.
+
+## Alternatives
+
+No alternatives were explored. If there are ideas during the review, we should add them here and discuss why we may or may not choose them. If we choose one of them we should merge this enhancement as rejected and create an enhancement for that approach.
```
jwmatthews, line 109: Might be worth capturing last year we did an initial stab of a DataMover design, we had some conversations on improving the approach in this PR, 
https://github.com/konveyor/data-mover/pull/3, then the team changed course with other priorities

enhancements/oadp/datamover-design/README.md
jwmatthews: The vendors have a pattern of storing data in a 'Target', they may want to do this even if the data is off-cluster in the cloud vendor (EBS example).  Some will do this as another form of redundancy, perhaps they have a proprietary target, maybe it's tape backed up or something and they want to setup a Policy that will ensure backed up data eventually makes it way to the target.   

I think there will be several motivations for why someone is interested in a data mover:

- concerned about CSI snapshots of OCS4 being local to cluster and need it copied off for D/R 
- transfer between storage/cloud technologies
- desire to run some proprietary logic such as catalog/index of data and storing in a custom target (this was a popular theme)

enhancements/oadp/datamover-design/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Motivation
+
+As more and more vendors use OADP as a common interface for interacting with backup and restore on an openshift cluster, these vendors want a deeper integration into how to move data. Right now, the primary methods that OADP moves data is through restic backup of a volume to an object bucket or via CSI snapshots that are off cluster. Some vendors don't fit into either of these buckets and they would like to easily provide their own version of a data mover to fully support user backups.
+
+### Goals
+
+List the specific goals of the proposal. How will we know that this has succeeded?
+
+1. An overall design, that will give a high level vision for how to allow for velero to interact with
```
jwmatthews, line 40: In past we had considered a non-Velero usage of the Data Mover as well.

I'm unclear at moment if that is still needed.  I think the reasoning why we felt a non-velero usage for the DataMover was desired was for OCS4 and ensuring that a CSI snapshot (not from Velero) had a good pattern to be copied off cluster for disaster recovery needs.

enhancements/oadp/datamover-design/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Proposal
+
+![DataMover Visual](./Datamover-Design.png)
+
+There are 3 new pieces here but first lets discuss the pieces that are being re-used. 
+
+We will create Velero plugin for [Volume Snapshotter and Backup Item Action](https://velero.io/docs/v1.6/custom-plugins/#plugin-kinds). This will allow us to determine when a Volume is being backed up and complete the necessary actions to move the data to the correct place. These are well known with lots of prior art. We should use the [CSI-plugin](https://github.com/vmware-tanzu/velero-plugin-for-csi) and the [restic](https://github.com/vmware-tanzu/velero/blob/5bd70fd8eef316d220317245e46dc6016c348dce/pkg/restic/repository_manager.go#L131) plugin as an example. 
+
+We will need to create a Data Mover Controller. This controller will watch a DataMover CRD where the status of moving data will be published. This Controller's only responsiblity to delegate via GRPC to the correct Data Mover.
```
jwmatthews, line 55: Worth considering if there is anything attractive of an approach where the 'plugin' a vendor provides is an image to run and we execute it and mount the PV to pod.  Similar to a lightweight contract like we had with 'automation broker' bundles.


The thought is several of the Vendor plugins would be legacy code that is focused on processing a mounted directory, if there was an easy path of containerize their binary and adjust an entrypoint to invoke it with a few known parameters it may be appealing.

enhancements/oadp/datamover-design/README.md
shawn-hurley: I think that makes sense, by cloud provider snapshot I was thinking specifically about an aws-ebs snapshot for instance.

enhancements/oadp/datamover-design/README.md
shawn-hurley: I wonder if we could achieve the same, with another sub component to enable this workflow that would conform to the DataMover GRPC interface. 

I think that the above would be appealing, but we would end up having a class of problems around reading what the pod is doing, Do we feel that this is the primary use case or just a sub use case?

enhancements/oadp/datamover-design/README.md
shawn-hurley: I think I am seeing the use case here. I will ad this as a specific story thanks for bringing it up @dymurray @jwmatthews

enhancements/oadp/datamover-design/README.md
```diff
@@ -0,0 +1,109 @@
...
+- [ ] Test plan is defined
+- [ ] User-facing documentation is created
+
+## Open Questions [optional]
+
+1. Do we want to try and get our GRPC endpoints into CSI or Openstorage API or both?
+
+## Summary
+
+Creating a common interface for datamovers, with a shared controller for that interface and using GRPC for datamover plugins. We will also create a velero plugin that will manage this new CRD when PVC's or PV's are attempted to be backed up.
```
sseago, line 29: Should the velero backup be the triggering event, or the creation of a CSI VolumeSnapshot (which the velero backup will trigger via the CSI plugin). Decoupling it from velero would allow the data mover to act when CSI snapshots are created independently of velero backups.

enhancements/oadp/datamover-design/README.md
sseago: @jwmatthews I asked a similar question above. The question is whether the velero backup action is the main triggering event, or if it's the CSI VolumeSnapshot creation. For velero users, it may not matter which, since the velero backup will itself trigger creation of the VolumeSnapshot (which could potentially trigger the DataMover action). But if we're triggering off VolumeSnapshot creation rather than Velero backup actions, we'd also be able to handle cases where non-Velero actions triggered the creation of CSI snapshots.

enhancements/oadp/datamover-design/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+We will create Velero plugin for [Volume Snapshotter and Backup Item Action](https://velero.io/docs/v1.6/custom-plugins/#plugin-kinds). This will allow us to determine when a Volume is being backed up and complete the necessary actions to move the data to the correct place. These are well known with lots of prior art. We should use the [CSI-plugin](https://github.com/vmware-tanzu/velero-plugin-for-csi) and the [restic](https://github.com/vmware-tanzu/velero/blob/5bd70fd8eef316d220317245e46dc6016c348dce/pkg/restic/repository_manager.go#L131) plugin as an example. 
+
+We will need to create a Data Mover Controller. This controller will watch a DataMover CRD where the status of moving data will be published. This Controller's only responsiblity to delegate via GRPC to the correct Data Mover. 
+
+A vendor will be able to supply a data mover as long as it conforms to a certain interface to provide moving the data for that particular set of data. For a default implementation we may want to consider how this will work.
+
+Another usecase that we should consider for the "DatamMover" CRD is to handle moving VolumeSnapshots to a different location.
```
sseago, line 59: Typo: s/DatamMover/DataMover/

shawn-hurley:
Closing PR as other folks are moving this forward

PR #43, opened 2021-09-21
Proposal to use Ansible Execution Environments for hooks

pemcg:
This PR proposes using Ansible Execution Environments for Konveyor Forkflift/Crane hook images, and to use the ansible-builder tool to create the EE images.

### DIFF
--- a/enhancements/common/ansible-execution-environments.md
+++ b/enhancements/common/ansible-execution-environments.md
@@ -1,0 +1,114 @@
+---
+title: ansible-execution-environments
+authors:
+  - "@pemcg"
+reviewers:
+  - TBD
+approvers:
+  - TBD
+creation-date: yyyy-mm-dd
+last-updated: yyyy-mm-dd
+status: provisional
+see-also:
+  - 
+replaces:
+  - 
+superseded-by:
+  - 
+---
+
+# Use Ansible Execution Environments for Hooks
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] User-facing documentation is created
+
+## Summary
+
+Two of the Konveyor projects - Crane and Forklift - have a requirement to run pre and post migration _hooks_ (usually Ansible playbooks) for per-instance configuration either side of the container or VM migration.
+
+These playbooks need to run podified, from container images built to contain the playbooks and all dependencies such as Python, Ansible Runner, collections, and user-defined roles. They will be run using `ansible-runner`. This is conceptually very similar to an Ansible _Execution Environment_ (see [Execution Environments](https://ansible-runner.readthedocs.io/en/stable/execution_environments.html)).
+
+This enhancement proposes that Konveyor projects use the established Ansible Execution Environment (EE) technology for hooks, along with [Ansible Builder](https://ansible-builder.readthedocs.io/en/latest/) to build the hook images.
+
+## Motivation
+
+### Goals
+
+The main goal is to re-use as much of the existing Execution Environment and ansible-builder technology as possible. This will allow for:
+
+- Use existing technologies - no wheel re-invention.
+- Less custom development and maintenance effort for Konveyor.
+- Commonality of technologies between Ansible & Konveyor projects.
+- Greater end-user familiarity with the technology being used.
+
+### Non-Goals
+
+- Considerable development of the Execution Environment, `ansible-runner`, or `ansible-builder` technology.
+
+## Proposal
+
+### User Stories
+
+The following user stories can be considered for this proposal.
+
+#### Story 1
+
+As a migration user I would like to use an Ansible Execution Environment image that I have already created as a pre/post migration hook.
+
+#### Story 2
+
+As a migration user hook developer I would like to use the `ansible-builder` tool that I'm already familiar with to build my hook image.
+
+### Implementation Details/Notes/Constraints
+
+The following components will probably be required for each Konveyor project:
+
+- UI support
+- Hook definition CR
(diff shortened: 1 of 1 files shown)


jmontleon:
It sounds reasonable that we would support running images built using ansible-builder.

For the image that we provide and have building in quay, and particularly when it comes to productization and building downstream I don't believe we can use ansible-builder. If that's true, would we continue to do something like we are doing now, while ensuring the images are compatible?

enhancements/common/ansible-execution-environments.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Proposal
+
+### User Stories
+
+The following user stories can be considered for this proposal.
+
+#### Story 1
+
+As a migration user I would like to use an Ansible Execution Environment image that I have already created as a pre/post migration hook.
```
shawn-hurley, line 60: Why could I not use some other image that I built that works? does it have to be an Ansible Execution Environment?

enhancements/common/ansible-execution-environments.md
```diff
@@ -0,0 +1,114 @@
...
+
+#### Story 2
+
+As a migration user hook developer I would like to use the `ansible-builder` tool that I'm already familiar with to build my hook image.
```
shawn-hurley, line 64: do we still have the requirement to create this image on the fly inside of an openshift cluster? is ansible-builder able to be run inside a pod? where do we propose to save the images for the migration hook to use them?

shawn-hurley (changes requested):
I am unsure of how this works based on what I thought the ask was. 

Can we re-visit the goals sections and determine if:
1. It is still a goal to build these images in an openshift pod and save the resulting image in some registry to be used by the hook.
2. Does the hook runner ONLY work with ansible execution environments or does it just run a pod?
3. Is there some other technology that we may want to be run in a hook image (golang binary, bash script, etc.) if this is a non-goal then we should state that.

enhancements/common/ansible-execution-environments.md
pemcg: I was suggesting that we standardise on Ansible EE's as that's what Ansible Automation Platform 2.0 (née Tower) customers are being encouraged to move to.

shawn-hurley: What does standardization mean in this context is my question, are we saying that hooks will only run in an ansible EE environment, are we saying that for the "building and running based on playbooks/git repo" use-case that we will only use ansible EE. 

The second to me sounds good, the first feels restrictive. I am trying to understand how standardization for hooks will be beneficial to the end-user or us if it is just a pod that is run.

pemcg: I think we make the "building and running based on playbooks/git repo" use-case the path of least resistance for customers, mainly because they will hopefully be familiar with the process already through AAP 2.0 usage.

fdupont-redhat: I think we need to take a layered approach here.

The most generic option is to allow any image to run. The user is responsible for building it and it requires more expertise in image building. On the Forklift side, we don't provide any tooling to help with building images.

Then, we consider the Ansible playbook case. The user has a playbook as a single file (not that common). The playbook is stored in a ConfigMap that is mounted in the pod at runtime. The pod image is based on ansible-runner and ships the k8s modules for convenience. We expect customers to be more at ease with playbooks, so we provide a pre-built execution environment.

Then, we consider the users who are more advanced with Ansible and already have a library of collections, roles, playbooks. They maintain them in Git repositories and want to leverage them as hooks. Building the hook image may be painful overtime and we want to provide a good experience through Ansible EE. It provides tools to build container images with all the requirements preloaded. A first level would be to document the usage of Ansible EE in the migration hook context. Then, we can hide the Ansible EE with a controller and a UI, in order to provide a seamless experience.

enhancements/common/ansible-execution-environments.md
fdupont-redhat: Storing the images in the internal registry is the most straight forward mechanism, as the service account will have the privilege to write images to the forklift namespace. This is a similar approach to S2I. There's still a risk of image sprawl and high demand in storage space over time.

As a secondary option, we could also allow users to store the image in any registry, given that they provide the URL and credentials. In that case, we need only scratch space, and probably a throttling mechanism.

enhancements/common/ansible-execution-environments.md
shawn-hurley: @fdupont-redhat 

> Then, we consider the users who are more advanced with Ansible and already have a library of collections, roles, playbooks. They maintain them in Git repositories and want to leverage them as hooks. Building the hook image may be painful overtime and we want to provide a good experience through Ansible EE. It provides tools to build container images with all the requirements preloaded. A first level would be to document the usage of Ansible EE in the migration hook context. Then, we can hide the Ansible EE with a controller and a UI, in order to provide a seamless experience.

I don't disagree with the overall layered approach outlined, but my understanding is that this proposal is really about the approach mentioned in the quoted text. If that is the case I think it would be really good to understand a couple of things.

1. How and where do these Images get stored once created.
2. If we are building images, do we have any permission issues that we should look into? 
3. When we are building these images, what do we need to make ansible EE do this (think about all the networking issues that customers have and pulling in collections etc.)? if it is hidden from the user(with a controller, UI, and API) what is the benefit to using ansible EE to us vs something more common to this environment like s2i or shipwright?

jwmatthews:
I am unsure if this approach is still being considered to be worked in Forklift or Crane.

I spoke with @shawn-hurley and we decided it'd be best to merge this PR in now so there is a record of the approach committed to the repo to make it easier for future developers to be aware of this and perhaps pick it up or extend.

PR #44, opened 2021-09-27
Update PVC references automatically after PV migration

pranavgaikwad:


enhancements/crane-1.x/inplace-storage-conversion/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Goals
+
+> Allow storage class conversion within the same namespace when source and target PVC names are different 
+
+## Proposal
+
+### Approach 1: Update the user workloads to point to migrated PVCs
+
+During a state migration, the _MigPlan_ controller will determine whether any of the PVCs are being migrated within the same cluster and in the same namespace. If that's the case, it will raise a _Warning_ notifying the user about the problem. The _MigMigration_ controller will use this information to enable the automatic PVC update procedure. The default behavior would be to update the PVCs if such case arises. The users can disable the automatic updates either through a _Spec_ field, an annotation or a _MigrationController_ flag (to be decided).
```
shawn-hurley, line 53: If I am understanding correctly this is creating a new PVC with a new name correct and then updating? How are we propose to choose this new PVC name?

enhancements/crane-1.x/inplace-storage-conversion/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+Stages _[a - c]_ update the migrated PVC objects. Any failure happened in any one of those stages will abort the rest of the stages and the reconciler will move on to the next PVC.
+
+A failure in _[d]_ will still result in aborting the rest of stages. However, a failure here indicates that the original PVC object still exists and the workloads will continue to run without issues. 
+
+Stages _[e - g]_ happen after the original PVC definitions are deleted. If any of these stages fail, the controller will attepmt to restore the original PVC definition. An added retry mechanism can be added to ensure that the original PVC is restored correctly.
+
+### Pros and Cons
+
+> Approach 1 will only work on "known" workload resources and will not work on workloads managed through a Custom Resource. On the other hand, Approach 2 doesn't make any changes to the original workloads.
```
shawn-hurley, line 93: Can we use something like duck typing to looks for pod template spec to more "generically" handle this?

I think you still have a potential problem for operator-backed PVCs and applications but if that is the case then we may just have to document it?

enhancements/crane-1.x/inplace-storage-conversion/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+> Approach 1 will need to update user workloads automatically. This is different from MTC's previous expectations that a migration results in a clone of source resource definitions without any change in them. Approach 2 on the other hand, will not change the original workloads at all.
```
shawn-hurley, line 95: We must change certain things on source files to work on new clusters correct? (resource version for example). I wonder if this is acceptable?

enhancements/crane-1.x/inplace-storage-conversion/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+> Approach 2 is complex to implement and requires deleting the original PVC object potentially leaving the environment in a bad state. Approach 1 is easier to implement and doesn't touch original PVC object. In case of failures, the users can simply revert back the PVC references to the original ones.
```
shawn-hurley, line 97: +1 

I would almost suggest if we went this approach that we would need to write a PV/PVC reclaim and re-update controller. That feels like a huge expansion of work and if anything goes wrong could be bad.

enhancements/crane-1.x/inplace-storage-conversion/README.md
```diff
@@ -0,0 +1,133 @@
...
+	
+	AutoUpdatePVCRefs bool `json:"autoupdatepvcrefs,omitempty"`
+}
+```
+
+The default behavior is to turn the functionality on whenever PVCs are migrated within the same namespace using _State Transfer_. 
+
+When upgrading from 1.7- versions, to 1.7 versions this field will be absent on all _MigMigration_ resources. However, _MigMigration_ resources are supposed to go to completion. The users ideally shouldn't have _Running_ migrations during an upgrade. But there could be blocked migrations waiting on perhaps a plan to become Ready. All such blocked migrations will automatically enable this functionality upon upgrade without the user noticing. Also, the functionality is only 
+
+When downgrading from 1.7+ versions to an older version, this feature will simply be absent. The extra _spec_ field present in _MigMigration_ CRs wont make any difference on the behavior of the controller.
```
shawn-hurley, line 133: These will be ignored by ETCD as well, such that after upgrading we need to make sure the default value for this field, preserves the behavior for that resource. 

For this reason, I wonder if we need to re-consider what the default is for this field.

enhancements/crane-1.x/inplace-storage-conversion/README.md
pranavgaikwad: @shawn-hurley the new name is supposed to be provided by the user. It can be an existing PVC they may have created which uses the new storage class. If it's not an existing PVC, a new PVC by that name will be created.

enhancements/crane-1.x/inplace-storage-conversion/README.md
pranavgaikwad: @shawn-hurley good idea. As long as, CRs follow the generic Pod Template spec, we can definitely handle those cases. Given our experience with MTC so far, most user workloads have been standard Kubernetes workloads. so documentation is possible.

alaypatel07 (commented):
Thanks for the detailed steps. I think we should Approach 1 feels the most safer and workable, even with its limitation of just working with the known workload APIs

enhancements/crane-1.x/inplace-storage-conversion/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+### Goals
+
+> Allow storage class conversion within the same namespace when source and target PVC names are different 
+
+## Proposal
+During a state migration, the _MigPlan_ controller will determine whether any of the PVCs are being migrated within the same cluster and in the same namespace. If that's the case, it will raise a _Warning_ notifying the user about the problem. The _MigMigration_ controller will use this information to enable the automatic PVC update procedure. The default behavior would be to update the PVCs if such case arises. The users can disable the automatic updates either through a _Spec_ field, an annotation or a _MigrationController_ flag (to be decided).
```
eriknelson, line 50: Thinking about backwards compat -- if the key is not present, what behavior do we want (should match the default I assume).

If it's not present and we want that to indicate to the controller "automatically update the references", then the polarity of the field needs to be something like "do_not_automatically_update", so that `do_not_automatically_update_ref = false` means **do** perform the auto ref update. That gets confusing quickly.

I'm not really advocating for anything here, but I think we want to be explicit about backwards compatibility concerns in any enhancement *any* time we suggest a change to the CRD API.

enhancements/crane-1.x/inplace-storage-conversion/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+1. Quiesce workloads
+
+2. Migrate data using DVM
+
+3. Find source PVC references in the workloads and update them to new migrated PVCs instead
+
+4. Un-quiesce the workloads
+
+### Approach 2: Use Indirect Volume Migration
```
eriknelson, line 65: We've been flirting with the idea of moving away from restic entirely for a lot of reasons (it being a daemonset and the consequences that come with that comes to mind). I'm not sure we want to couple ourselves _more_ to restic if that's really the direction we want to be headed.

enhancements/crane-1.x/inplace-storage-conversion/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+1. Quiesce the workloads
+
+2. Backup source PVCs to replication repository and backup data using Restic
+
+3. Delete source PVCs
+
+4. Restore PVCs in the same namespace and restore data using Restic
+
+### Approach 3: Update PVs inplace without changing the user workloads
```
eriknelson, line 75: This approach scares me without a bullet proof fallback should a failure interupt this transaction. The entire swap has to be atomic, else we need to be able to easily rollback.

enhancements/crane-1.x/inplace-storage-conversion/README.md
```diff
@@ -0,0 +1,167 @@
...
+### Goals
+
+> Allow storage class conversion within the same namespace when source and target PVC names are different 
+
+## Proposal
+During a state migration, the _MigPlan_ controller will determine whether any of the PVCs are being migrated within the same cluster and in the same namespace. If that's the case, it will raise a _Warning_ notifying the user about the problem. The _MigMigration_ controller will use this information to enable the automatic PVC update procedure. The default behavior would be to update the PVCs if such case arises. The users can disable the automatic updates either through a _Spec_ field, an annotation or a _MigrationController_ flag (to be decided).
+
+The _MigMigration_ controller will choose any one of the below approaches to restore PVs:
+
+### Approach 1: Update the user workloads to point to migrated PVCs
```
eriknelson, line 54: I think I actually like this approach the most for a few reasons:
* It feels the safest in that we'd be able to delegate to the user whether or not they want us to clean up the original pvc afer we're done (i.e. delete), or if they want them to be preserved in case they need to rollback.
* This makes me think we might want to support rollback for a conversion that restores the original pvc references. This approach would make it easier to accomplish that.

enhancements/crane-1.x/inplace-storage-conversion/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+```golang
+type MigMigrationSpec struct {
+	[...]
+	
+	AutoUpdatePVCRefs bool `json:"autoupdatepvcrefs,omitempty"`
+}
+```
+
+The default behavior is to turn the functionality on whenever PVCs are migrated within the same namespace using _State Transfer_.
```
eriknelson, line 151: Do we have any concern with altering default behavior with minor version increments? Meaning in 1.6 we did not do this by default, and in 1.7, it's possible that by default, we'd have an entirely different behavior that's potentially undesirable. If a user is already familiar with the behavior as of 1.6, the default behavior would come as a surprise with 1.7?

@shawn-hurley I'm interested to hear your thoughts on this question.

enhancements/crane-1.x/inplace-storage-conversion/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+```golang
+apiVersion: migration.openshift.io/v1alpha1
+kind: MigMigration
+metadata:
+  annotations:
+    migration.openshift.io/auto-update-pvc-refs: "true"
+```
+
+Any existing resources from previous versions would not have the above annotation set. Thus, the functionality will be turned off for _MigMigration_ resources created in older versions of MTC.
```
eriknelson, line 167: From a semantics perspective, it feels more correct to have conditional behavior be a `spec` field, as in: "I'm a user and I'm expressing my intent for this migration to behave in this manner".

enhancements/crane-1.x/inplace-storage-conversion/README.md
eriknelson: Upon further reading, the issue of this approach only working for "known" workloads feels like it makes this not really an option. Can we elaborate on what is meant by "known" type? If it can't work generically, I'm inclined to take it off the table.

eriknelson (changes requested):

enhancements/crane-1.x/inplace-storage-conversion/README.md
alaypatel07: I think by "known" workloads we mean all the popular workloads which have PodTemplate embedded inside. mig-controller already has the logic to read those pod templates and create a stage pod for it. Using the same workflow but editing the pod template to use a migrated PVC is what this approach suggests. Since this already exists, we are not creating the situation any worse by implementing this approach. The unknown workloads would be pods using PVC that are managed by a specific operataor, this approach will break those, but so will be the stage pod logic and quiesce logic.

eriknelson: > The unknown workloads would be pods using PVC that are managed by a specific operataor, this approach will break those, but so will be the stage pod logic and quiesce logic.

+1, thank you that makes sense, can get this definition / detail added to the document to specify exactly when this breaks down?

I wonder if it would be acceptable to say: "Hey this automatic swap will succeed for traditional workloads that look like $DETAILS, but under this circumstance it is impossible to automatically update the references and you should disable the feature and handle it yourself"

enhancements/crane-1.x/inplace-storage-conversion/README.md
shawn-hurley: I agree this needs to follow two main things.

1. The default value must preserve the old behavior. Say an old resource exists, it should continue to work as defined in that CR does not get new behavior (changing deployments). 
2. That the new field must be optional and the default of the optional needs to conform to 1 above.

enhancements/crane-1.x/inplace-storage-conversion/README.md
shawn-hurley: I wonder if we even want to try and support something that is owned by operator and not a "pod templated controller". 

I can imagine that even the PVC name and storage class could be owned by some other controller in the operator world. I wonder if a "consult your operator if deployment and PVC are owned by the operator" may be the safest guidance?

enhancements/crane-1.x/inplace-storage-conversion/README.md
shawn-hurley: I think that this is not something that we can do IMO. a resource that works one way on 1.6 should not have radically different behavior on 1.7 IMO.

I think that we *COULD* in some release notes over a couple of versions change the default but I would question if that is really necessary here. 

I think that the UI could set this field by default in 1.7 and that makes sense to me but from an API perspective, I would not want to change this behavior.

enhancements/crane-1.x/inplace-storage-conversion/README.md
shawn-hurley: +1 the annotation just becomes an un-versioned spec value that we have to handle IMO.

pranavgaikwad:
@shawn-hurley @sseago @alaypatel07 I moved the alternative approaches we discarded to a different section and addressed concerns around default value of the field. PTAL!

enhancements/crane-1.x/inplace-storage-conversion/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+The only API change needed is the switch that disables the automatic update of PVC references. Currently, I have not put much thought on where to keep it. It could be a _Spec_ field on _MigMigration_ CR or an annotation or a _MigrationController_ variable passed as environment variable through the configmap. 
+
+If it were to be a _Spec_ field on _MigMigration_, it will be an additive _optional_ field added to _MigMigration_ API: 
+
+```golang
+type MigMigrationSpec struct {
+	[...]
+	
+	AutoUpdatePVCRefs bool `json:"autoupdatepvcrefs,omitempty"`
```
shawn-hurley, line 90: I think I would prefer this to be in a typed field as well

PR #47, opened 2021-09-28
Adding enhancement for plugin-management

JaydipGabani:


enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,131 @@
...
+
+The initial solution will ship with listing, installing, and uninstalling plugins from the managed repository only and will not support adding custom repository and list plugins from there. The document might include some details on how custom repository might be added, but that is a high level approach and is subjected to change.
+
+## Proposal
+
+There are two parts of the solutions, the first part includes solving how to list plugins from the repository. The second part includes installing/uninstalling the plugin inside `plugin/managed-plugins`.
+
+### First Part
+
+One of the proposed solution is to have a `plugins` directory at the root of such repositories and respective `<plugin-name.yml>` inside the said directory from where we are listing plugins. The format of the yml file should be something like this
```
shawn-hurley, line 53: I wonder if this should be versioned?

enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,131 @@
...
+
+```
+version: <plugin-version>
+shortDescription: <short description>
+description: <description>
+platforms:
+- selector:
```
shawn-hurley, line 60: Wondering why os and arch are not just high-level values here? do we ever need the ability to install multiple os/arch's on a single machine? Is there a scenario where a binary is able to run on two different os/arch combos?

enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,131 @@
...
+
+    Possible solutions - 
+
+    1. Must have unique plugin name across all plugin repositories (on a big scale this is more difficult to manage). OR
+    2. Provide a flag in `add-plugin` sub command to input a source from where user wants to install the plugin. For example `crane plugin-manager add <plugin-name> --source=<source-url>` or `crane plugin-manager add <plugin-name> --source=<name>` depending on which solution of the two is selected for implementation.
+
+
+## Result
+
+- `crane plugin-manager list` will list all the available plugins
```
shawn-hurley, line 125: does this list plugins that are available just for the current arch?

enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,131 @@
...
+- `crane plugin-manager add-plugin foo` will place the plugin `foo` inside `managed` directory inside the default plugin directory - i.e. `/plugins/managed/foo`
```
shawn-hurley, line 126: What options might we want to add here. I am thinking bare minimum we would want to add `-p --plugind-dir` for specifying the plugin. What about a particular version?

enhancements/crane-2.0/plugin-management/README.md
JaydipGabani: Yes! updating the description

enhancements/crane-2.0/plugin-management/README.md
JaydipGabani: There will be one manifest file for each plugin, it makes sense to keep these variables at a lower level to make sure that we only download binaries of a specific plugin for the current os-arch. In the case where we can use the same binary for different os-arch, we can have multiple entries with different os-arch and same binary uri. 

> do we ever need the ability to install multiple os/arch's on a single machine?

I don't think we need it, let me know if I am wrong.

JaydipGabani: To make it less confusing, this yml file will reside in GitHub repo and not on the local machine. The cli will traverse through the plugin file and filter out the the plugins for current local os/arch.

JaydipGabani: Does this make sense?

enhancements/crane-2.0/plugin-management/README.md
JaydipGabani: do you mean naming the files like `plugin-name-version.yml` ?

enhancements/crane-2.0/plugin-management/README.md
JaydipGabani: As a part of possible solution I discussed having a flag [here](https://github.com/konveyor/enhancements/pull/47/files#diff-5c8731efc9f26ab661c598e3b3eb22a0699cfb89d0e45ad158a18d0daa1c4d12R153), to also add the functionality for a specific version installation, I think adding one more flag makes sense.

enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,164 @@
...
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+
+## Summary 
+
+At a high level plugin management functionality will be
+
+- Able to list the available plugins
+- Able to install/uninstall plugins
+- Able to add/remove custom repositories to list the plugins from (long term goal)
```
eriknelson, line 37: Do updates need to be considered? @shawn-hurley I think you had some opinions here.

It's out of scope for an initial release but I want to be sure we don't make any decisions here that might cause problems with plugin updates down the line.

enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,164 @@
...
+
+## Motivation
+
+Apart from the offerings of crane-lib, we would like to add a way to allow users to extend the functionality of the tool by installing the managed or custom plugins.
```
eriknelson, line 41: I think it's worth calling out here that we believe a lot of the value of crane for users is going to be **easy access to plugins that have already been written that solve their problems**. I would expect more users to _consume_ plugins than _write_ them, and I think @jwmatthews has expressed that if users **must** write plugins to do useful things, then the tool has failed.

This enhancement seeks to define that ease of access to the extension ecosystem.

enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,164 @@
...
+
+### Goals
+
+1. Add a new command `plugin-manager` that is able to handle plugin management functionalities.
```
eriknelson, line 45: Could we just call the action `plugin`? I.E. `crane plugin install` `crane plugin uninstall`. I think I've seen this pattern with other tooling like vagrant.

enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,164 @@
...
+2. This command must provide sub-commands to fullfil the requirement such as listing, installing, and uninstalling plugins
```
eriknelson, line 46: I would also want the ability to add and remove remote repos, as well as listing specific repos, and seeing which plugin belongs to which repo when listing all my plugins.

Ex:

`crane plugin repo {add,remove} eriks-repo <URL> # Similar to git`

```
$ crane plugin list --repo=eriks-repo
Name                  Repo                   Version       Description
=============================================
erik-plugin          eriks-repo            v1.2           Does really cool stuff
```

enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,164 @@
...
+
+### Non-goals
+
+The initial solution will ship with listing, installing, and uninstalling plugins from the managed repository only and will not support adding custom repository and list plugins from there. The document might include some details on how custom repository might be added, but that is a high level approach and is subjected to change.
```
eriknelson, line 50: I have a concern here: I believe that we'll need to define exactly what it means to be a crane plugin repository as part of this enhancement, because our "official" repository needs to be compliant with that definition exactly like any other plugin repo.

enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,164 @@
...
+
+#### Story 3
+
+As a user, I would like to uninstall any plugin that I might want to remove, `crane plugin-manager remove-plugin foo`
+
+Should remove the plugin from appropriate directory to not have this plugin execute during `crane transform`
+
+## Design details
+
+There are two parts of the solutions, the first part includes solving how to list plugins from the repository. The second part includes installing/uninstalling the plugin inside `plugin/managed-plugins`.
```
eriknelson, line 82: I've seen "managed-plugins" used several times in the doc. Can we define what is meant by "managed"?

enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,164 @@
...
+
+```
+crane plugin-manager --help
+
+Usage:
+  crane plugin-manager [command]
+
+Available Commands:
+  list              lists all the available plugins
+  add-plugin        installs the desired plugin
```
eriknelson, line 115: I think `plugin` in these commands is redundant.

enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,164 @@
...
+
+### Second Part
+
+An additional command `plugin-manager` to manage all the plugins.
+
+```
+crane plugin-manager --help
+
+Usage:
+  crane plugin-manager [command]
```
eriknelson, line 111: I would advocate for just `crane plugin` as the action.

enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,164 @@
...
+
+Global Flags:
+  --debug   Debug the command by printing more information
+
+```
+
+
+#### To support additional repository to install plugins
+
+<b>Solution 1</b>:
```
eriknelson, line 129: I hadn't considered an env var approach, but I actually like as an option to potentially override the stateful file on the system.

Do we want to do both? If so, should they merge with one another, or should the env var *override* the statefully installed list in the file?

enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,164 @@
...
+
+Other than default repository (managed by our team), also read from env variable `CUSTOM_PLUGINS` to read from custom plugin github repository and list plugins.
+
+`CUSTOM_PLUGINS` variable holds the links to the github repositories separated by `;`. For example `CUSTOM_PLUGINS=github.com/konveyor/plugins;github.com/jgabani/plugins`. List command will iterate through all the custom repositories url and list all the plugins.
```
eriknelson, line 133: These values suggest that all plugins will be git repos, so effectively all that crane needs is a URL (coordinate) to a repo that contains a git repo.

These values also suggest that we expect them to be go repos that can be built with `go install/get`. What if I have a non-go repo? We've said that the interface to be a crane plugin is just stdout/in, so that means it literally could be anything, including a C binary that requires compilation and the C toolchain.

That implies to me that the plugin manager needs to expect executables, because we can't be expected to know how to compile anything under the sun.

enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,164 @@
...
+
+List command will iterate through all the custom repositories url from `source.yml` file and list all the plugins.
+
+##### Possible complications
+
+1. Conflicting plugin name within different plugin repositories that are independent. 
+
+    Possible solutions - 
+
+    1. Must have unique plugin name across all plugin repositories (on a big scale this is more difficult to manage). OR
```
eriknelson, line 152: >      1. Must have unique plugin name across all plugin repositories (on a big scale this is more difficult to manage)

I don't think this is the right path. The fully qualified name of a plugin should be something that incorperates both the plugin name and repo, as indicated by option 2 here. The tooling can accept a short name if it's unambiguous what it refers to (in the case that there is only one plugin with the provided name across all repos), or it can error out if a short name is used that matches on more than one plugin, and require the plugin to be fully qualified including the repo.

enhancements/crane-2.0/plugin-management/README.md
eriknelson: I think we need to have some notion of what version of the plugin is installed. I asked this question previously, but I'm guessing we don't ever want to allow for more than one version of the same plugin to be installed, in which case it's somewhat implicit what's present.

eriknelson (changes requested):

enhancements/crane-2.0/plugin-management/README.md
JaydipGabani: That's a part of a long-term goal I have discussed [here](https://github.com/konveyor/enhancements/pull/47/files#diff-5c8731efc9f26ab661c598e3b3eb22a0699cfb89d0e45ad158a18d0daa1c4d12R221) in the updates after our discussion.

enhancements/crane-2.0/plugin-management/README.md
JaydipGabani: The `<plugin-name.yml>` file should have the location of a pre-compiled binary, so we don't have to compile anything.

enhancements/crane-2.0/plugin-management/README.md
JaydipGabani: +1

enhancements/crane-2.0/plugin-management/README.md
sseago: I think every "plugin-manager" subcommand needs the plugin-dir arg. At the very least, add-plugin and remove-plugin need it, since that determines where the plugin is added to or removed from. "list" may not need it *if* list only looks at the repos and ignores what's currently installed, but if list needs to know what's currently installed when formatting its output, then it needs the plugin-dir as well -- this should be registered in exactly the same way as it's registered for the transform command.

enhancements/crane-2.0/plugin-management/README.md
sseago: So the notion of *upgrading* is out of scope. I think *updating* -- i.e. simply replacing a currently-installed plugin (in the plugin dir) with whatever version we''re pulling from the repo should be supported. At the very least, updating a plugin could be equivalent to remove followed by add. I'm not sure if anything else is needed at this point. I.e. "add" complains if it already exists, "update" removes then adds .

enhancements/crane-2.0/plugin-management/README.md
sseago: @eriknelson I think that's exactly what the above is describing. "extending the functionality of the tool by installing..." The edge case is users writing their own plugins, but in that case (at least for now) they're probably not using the plugin manager. They're just dropping their custom plugins into the plugin dir.

enhancements/crane-2.0/plugin-management/README.md
sseago: @eriknelson I think plugin-manager is slightly clearer, since its scope is managing the list of installed/configured plugins. A command like "plugin" might imply that it is also used to deal with running plugins, which is explicitly out-of-scope here. Running plugins is done by the "transform" command.

enhancements/crane-2.0/plugin-management/README.md
sseago: Yes -- adding/removing repos are the other commands that plugin-manager will eventually handle. In our last meeting we explicitly declared those operations out of scope for this first iteration, though.

enhancements/crane-2.0/plugin-management/README.md
sseago: @eriknelson the challenge is that it's not necessarily clear that our official repository will be compliant for the first release where we don't support adding/removing repos -- it will just be there. We can't make it compilant until we define what it is, but that's part of the next phase (along with add/remove repo).

enhancements/crane-2.0/plugin-management/README.md
sseago: @eriknelson I agree. "plugin-manager add" and "plugin-manager remove" are clearer. We *might* want to add a list command for listing installed plugins as well -- I know this already exists under transform, which is fine, but since we're interacting with plugin-manager here, we have add/remove actions. We need "list available" and "list installed" actions. I'm not sure whether these are both subcommands under list (and without invoking a subcommand, list just does one or the other), or whether we should be explicit: "add", "remove", "list-available", and "list-installed" -- I don't have a strong opinion on which of these we want here.

enhancements/crane-2.0/plugin-management/README.md
sseago: @eriknelson maybe we can have a discussion about this tomorrow. For reasons mentioned above,  Ithink "plugin-manager" is less ambiguous. But perhaps there's something else I'm missing.

enhancements/crane-2.0/plugin-management/README.md
sseago: Right. a plugin *is* an executable. And while it's possible to write a plugin as a shell script or a C program, we certainly don't want to encourage anyone writing plugins in anything other than the golang crane-lib binary-plugin framework. But yes, the file referenced by the manifest for the architecture being run on must be an executable file that will run on the current architecture.

enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,300 @@
...
+
+```
+crane plugin-manager list --help
+
+Usage:
+  crane plugin-manager list [flags]
+
+Flags:
+  -h, --help    help for plugin-manager list
+  --repo        List plugins from specific repo (optional)
```
sseago, line 174: If --repo is not specified here, this will list plugins from across all repos, is that correct?

enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,300 @@
...
+
+```
+crane plugin-manager add --help
+
+Usage:
+  crane plugin-manager add <name> [flags]
+
+Flags:
+  -h, --help    help for plugin-manager add
+  --repo        Install plugin from specific repo (optional), if not passed iterate through all the repo and install the desire plugin. In case of conflicting name ask user to specify the repo from which to install the plugin
```
sseago, line 189: So if more than one plugin with the name exists and repo is not specified we error out with a message listing the matching repos, asking them to specify one?

enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,300 @@
...
+  --version     Install specific plugin version (if not passed, installs latest plugin version)
```
sseago, line 190: Also needs a `--plugin-dir` (`-p`) param here, which specifies where the plugin will be installed. The default value will be "plugins" as with the `transform` command.

enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,300 @@
...
+Global Flags:
+  --debug   Debug the command by printing more information
+
+```
+
+```
+crane plugin-manager remove --help
+
+Usage:
+  crane plugin-manager remove <name>
```
sseago, line 201: Also needs a `--plugin-dir` (`-p`) param here, which specifies where the plugin will be removed from. The default value will be "plugins" as with the `transform` command.

sseago: Are plugins namespaced by the repo? I.e. is it valid to have two plugins with the same name installed as long as they have a different repo key? I'd think this would be desirable, but we should work through the implications. If this will be allowed, then remove needs an optional `--repo` flag just like add does.

enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,300 @@
...
+
+
+#### To support additional repository to install plugins
+
+<b>Solution 1</b>:
+
+Other than default repository (managed by our team), also read from env variable `CUSTOM_PLUGINS` to read from custom plugin github repository and list plugins.
+
+`CUSTOM_PLUGINS` variable holds the links to the github repositories separated by `;`. For example `CUSTOM_PLUGINS=github.com/konveyor/plugins;github.com/jgabani/plugins`. List command will iterate through all the custom repositories url and list all the plugins.
+
```
sseago, line 220: Everything else is managed as an explicit flag. If we're going with this approach, we should probably use a flag instead of an environment variable for consistency. Note that unrelated to this, we also want to eventually implement the ability to pass in flag values via an input file, so if/when we do that, the contents of this could be included in that as well.

enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,300 @@
...
+<b>Solution 2</b>:
+
+Add subcommands within `plugin-manager` to support adding/removing any plugin repos. That could look something like this
+
```
sseago, line 224: I think Solution 2 is more consistent with the cli-oriented crane design than solution 1.

enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,300 @@
...
+```
+crane plugin-manager repo --help
+
+Usage:
+  crane plugin-manager repo [command]
+
+Available Commands:
+  add           Add the desired plugin repo to install the plugins from 
+  remove        Remove the added plugin repo
```
sseago, line 234: We need `list` as well.

enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,300 @@
...
+
+##### Possible complications
+
+1. Conflicting plugin name within different plugin repositories that are independent. 
+
+    Possible solutions - 
+
+    1. Must have unique plugin name across all plugin repositories (on a big scale this is more difficult to manage). OR
+    2. Provide a flag in `add` sub command to input a source from where user wants to install the plugin. For example `crane plugin-manager add <plugin-name> --source=<source-url>` or `crane plugin-manager add <plugin-name> --source=<name>` depending on which solution of the two is selected for implementation.
+
```
sseago, line 290: There's a third option here. Rather than a user-defined "namespace" for installed plugins, lets just namespace them by repo name. If you have one repo installed, you can have at most 1 managed plugin with the a given name. With 2 repos, you can have 2, etc. In addition to this, the set of unmanaged plugins will be in its own pseudo-repo namespace. I say pseudo-repo because the plugin manager will completely ignore it. Won't show up in repo list, or plugin list, etc. However, since the `transform` command will need to be made repo-aware ( to deal with the possibility of plugin "crane-lib/pod" as well as "foo-repo/pod"), the local unmanaged plugins (which simply live in the top-level "plugins" dir, or possibly in a hard-coded "local" directory -- we probably need to call that out somewhere in this design) could also have names clashing with the managed ones. But that's OK -- transform doesn't actually care whether names are unique, and in fact with locally-managed plugins there is no way of preventing two plugins in the plugins dir from reporting the same name.

enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,300 @@
...
+
+## Result
+
+- `crane plugin-manager list` will list all the available plugins for the current os arch.
+- `crane plugin-manager add foo` will place the plugin `foo` inside `managed` directory inside the default plugin directory - i.e. `/plugins/managed/foo`
```
sseago, line 295: As discussed above, between "managed" and "foo" would need to be a directory based on the repo key.

enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,300 @@
...
+- `crane plugin-manager remove foo` will remove the plugin `foo` from `plugins/managed/` directory
+
+## Changes in existing code
+
+Transform command need to look at `/plugins/` directory recursively to include all the plugins and handle possible conflicts and priorities.
```
sseago, line 300: I suspect the current conflict/priority code will be fine for the first release. Since plugin priority and plugin skip functionality is based exclusively on name, though, two plugins with the same name in different repos will have the same priority and will either both be skipped, or both not skipped (depending on name match). We may want to eventually modify the skip and priority code to optionally take a repo-namespaced plugin name.

enhancements/crane-2.0/plugin-management/README.md
eriknelson: > I think every "plugin-manager" subcommand needs the plugin-dir arg. At the very least, add-plugin and remove-plugin need it, since that determines where the plugin is added to or removed from.

+1, I agree we should have a cli argument for the plugin directory, but I'd like to see that treated as an override. In my discussion with Jaydip, we agreed there's probably a `$XDG_CONFIG_DIR/crane.yml` file that can serve as a config, and upon a first run of the tool, it probably makes sense to set a default location there so the user doesn't have to pass the argument on the cli with every run of a command that needs it?

@JaydipGabani this actually brings to mind a potential reason to use a `crane/` dir rather than only a `crane.yml` file (thinking maybe a `.config/crane/config.yml` file and maybe a plugin dir where the plugins are downloaded? Although I'm not sure a config directory is the right place for the plugins. There is also traditionally a `~/.cache` directory, but I'm not sure that's the right default location either. Interested in some other opinions here. Generally, where is a sensible default place to put plugins?

> "list" may not need it if list only looks at the repos and ignores what's currently installed

We probably want a list command that lists those that are installed, and a list remote command that lists those that are available via currently configured repos. Node/Ruby/Python virtualenv tools all provide some variant of this. For example, nvm has an `ls` command for locally installed versions that are available, and `ls-remote` that lists all those that are available for install. I've always preferred that to a cli arg like `ls` and `ls --remote` because it's plainly obvious from --help what they are, since they're top level commands.

enhancements/crane-2.0/plugin-management/README.md
eriknelson: > So the notion of upgrading is out of scope.

How do you define upgrade vs. update? 

Also wondering if we want the `crane` binary to have a command that updates itself to a newer version, but that's probably also out of scope.

> At the very least, updating a plugin could be equivalent to remove followed by add. I'm not sure if anything else is needed at this point. I.e. "add" complains if it already exists, "update" removes then adds .

+1

enhancements/crane-2.0/plugin-management/README.md
eriknelson: > They're just dropping their custom plugins into the plugin dir.

I think generally we should seek to build the tool such that manual management / installation of plugins is something that isn't commonly needed. But +1 I think your reading of the sentence is fair. I think it's worth calling out in the enhancement that it is an goal that most users shouldn't have to write their own plugins, but of course that's also where its power lies.

enhancements/crane-2.0/plugin-management/README.md
eriknelson: +1 Scott, I think I can get behind that. I don't have a strong opinion on this one.

enhancements/crane-2.0/plugin-management/README.md
eriknelson: Mentioned in another comment but I think you've convinced me otherwise.

enhancements/crane-2.0/plugin-management/README.md
sseago: @eriknelson Plugin directory already has a default, which is just "plugins". Changing this to look in crane.yml makes sense, but only if it's done for all commands (not just plugin-manager), so that might be a change that is orthogonal to this feature. Along those lines, I'd suggest the following: 
- Provide a `--plugin-dir` arg for this command exactly as is provided for `transform`
- If plugin-dir isn't supplied, use plugin-dir from crane.yml (Obviously, `transform` needs this new behavior as well.)
- If plugin-dir isn't supplied and crane.yml doesn't supply it (or the file doesn't exist), fall back to the current `plugins` default dir

enhancements/crane-2.0/plugin-management/README.md
sseago: Updating the crane binary to a new version seems to be completely unrelated to this. We're talking about "I have plugin $foo installed, but I want to install the new $foo plugin". Since plugins don't have state, this is simply swapping out one binary (and manifest?) with another. In that context, "update" is identical to "remove" followed by "add".

enhancements/crane-2.0/plugin-management/README.md
sseago: Manual management won't be needed if the only plugins used are published in plugin repos. And yes, most users shouldn't have to write their own plugins, but we also don't intend to prevent users from being able to do so. Even if only to assist plugin writers, the crane binary/cli must have the ability to invoke manually-built-and-installed plugins. It would essentially function as a "local" repo which the plugin-manager won't ever interact with, but the `transform` command will see is another list of plugins it invokes.

enhancements/crane-2.0/plugin-management/README.md
sseago: If we had `plugin install` and `plugin run` as supported subcommands, then I think `plugin` would be the right name. But the plugin manager's functionality is roughtly equivalent to rpm, where `m` stands for `manager` there.

enhancements/crane-2.0/plugin-management/README.md
JaydipGabani: Yes, that is correct.

enhancements/crane-2.0/plugin-management/README.md
JaydipGabani: That makes the most sense to me, open to suggestions!

sseago: The other possibility would be to install both, but I don't think we should do that by default. A user most likely has a particular plugin from a particular repo in mind and doesn't realize that the same name is used elsewhere. I think if 2 share a name and the repo isn't specified, we should fail and provide a message listing the repos which contain it. Remove without a repo specified should behave similarly (if there are 2 installed plugins in different repos with that name).

enhancements/crane-2.0/plugin-management/README.md
JaydipGabani: Agree

enhancements/crane-2.0/plugin-management/README.md
JaydipGabani: @sseago are you suggesting we should allow users to install multiple plugins with same name as long as they belong to different repo? If so, are you suggesting that we place plugin binaries in a similar manner to following:

plugin `foo` belonging to `bar` repo would be placed inside `/plugins/managed/bar` and plugin `foo` belonging to `foo` repo would be placed inside `/plugins/managed/foo` ?

enhancements/crane-2.0/plugin-management/README.md
JaydipGabani: @sseago wondering if instead of failing we just prompt the user for input for a specific repo from which the plugin is needed? and on top of that, we can also ask users to input `a, if one wants to install all the plugins with the same name across different repos. However, we might need to be careful with the implementation when the `crane plugin-manager add` command specifies a version, we will need to query the repos while respecting the asked version of the plugin.

sseago: The problem with "prompting the user" is that I imagine that some subset of our supported use cases will be scripted invocation. At least for the first release, it's probably better to error out with a detailed message. Maybe eventually we should support an "interactive" mode, but we can't assume interactive.

enhancements/crane-2.0/plugin-management/README.md
sseago: This is just repos added locally. There's no way of knowing what repos are available remotely, since that would imply some higher-level structure of "list of repos" which we aren't dealing with at all.

enhancements/crane-2.0/plugin-management/README.md
sseago: @JaydipGabani That's exactly what I was thinking. Whatever we decide to be the appropriate subdirectory corresponding to a repo would be the "fo" and "bar" above. I think @eriknelson had suggested in a different comment that names should be unique "within a repo" rather than "across all repos" -- essentially each repo's plugins are indepentent of another's.

enhancements/crane-2.0/plugin-management/README.md
JaydipGabani: Ohh, makes sense :+1: , I will change the desctiption

enhancements/crane-2.0/plugin-management/README.md
JaydipGabani: Okay :+1:

enhancements/crane-2.0/plugin-management/README.md
JaydipGabani: Thanks! This makes things clearer

enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,304 @@
...
+```
+crane plugin-manager remove --help
+
+Usage:
+  crane plugin-manager remove <name>
+
+Flags:
+  -h, --help                 Help for plugin-manager remove
+  -p, --plugin-dir string    The path where binary plugins are located (default "plugins")
+  --repo string              Install plugin from specific repo (optional), if not passed iterate through all the repo and install the desire plugin. In case of conflicting name ask user to specify the repo from which to install the plugin
```
sseago, line 207: "remove" instead of "install" here.

enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,304 @@
...
+```
+crane plugin-manager add --help
+
+Usage:
+  crane plugin-manager add <name> [flags]
+
+Flags:
+  -h, --help                 help for plugin-manager add
+  -p, --plugin-dir string    The path where binary plugins are located (default "plugins")
+  --repo string              Install plugin from specific repo (optional), if not passed iterate through all the repo and install the desire plugin. In case of conflicting name ask user to specify the repo from which to install the plugin
```
sseago, line 190: "desired plugin" rather than "desire plugin"

sseago (commented):
A couple of minor comments on the `--repo` documentation

enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,310 @@
...
+
+#### Story 4
+
+As a user, I would like to uninstall any plugin that I might want to remove, `crane plugin-manager remove foo`
+
+Should remove the plugin from appropriate directory to not have this plugin execute during `crane transform`
+
+## Design details
+
+There are two parts of the solutions, the first part includes solving how to list plugins from the repository. The second part includes installing/uninstalling the plugin inside `plugin/managed-plugins/<repo-name>`.
```
sseago, line 94: I'd say `<plugin-dir>/managed-plugins/<repo-name>` here since the default "plugins" can be overridden with a cli flag.

enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,310 @@
...
+
+The format of the yml file should be something like this
+
+
+```
+name: <plugin-name>
+shortDescription: <short description>
+description: <description>
+versions:
+    version: <plugin-version>
```
sseago, line 123: For correct yaml, this line (and equivalent ones below) need a dash to indicate the start of the neext list item:
```
  - version: <plugin-version>
```

enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,310 @@
...
+            os: <os name>
+            arch: <arch>
+      uri: <uri of binary>
+      sha: <sha of binary>
+    - selector:
+        matchLabels:
+            os: <os name>
+            arch: <arch>
+      uri: <uri of binary>
+      sha: <sha of binary>
```
sseago, line 136: Also, for each version we also need to list the optional fields data. This should include the same metadata as the `OptionalFields` slice returned by the plugin's metadata API call:
```
type OptionalFields struct {
        FlagName string `json:"flagName"`
        Help     string `json:"help"`
        Example  string `json:"example"`
}
```

enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,310 @@
...
+crane plugin-manager list --help
+
+Usage:
+  crane plugin-manager list [flags]
+
+Flags:
+  -h, --help                 Help for plugin-manager list
+  --repo string              List plugins from specific repo (optional)
+  --installed bool           If this flag is passed, list all the installed plugins
+  -p, --plugin-dir string    The path where binary plugins are located (default "plugins")
```
sseago, line 188: Now that we're adding optional fields to the manifest format we may want a flag to indicate whether to return field documentation, probably false by default (since it could be pretty verbose) but with an option to list it. Same with versions available.

enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,310 @@
...
+```
+crane plugin-manager remove --help
+
+Usage:
+  crane plugin-manager remove <name>
+
+Flags:
+  -h, --help                 Help for plugin-manager remove
+  -p, --plugin-dir string    The path where binary plugins are located (default "plugins")
+  --repo string              Install plugin from specific repo (optional), if not passed iterate through all the repo and install the desire plugin. In case of conflicting name the command fails and asks user to specify the repo from which to install the plugin
```
sseago, line 221: "install the desire plugin" -> "remove the desired plugin"

enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,310 @@
...
+```
+crane plugin-manager add --help
+
+Usage:
+  crane plugin-manager add <name> [flags]
+
+Flags:
+  -h, --help                 help for plugin-manager add
+  -p, --plugin-dir string    The path where binary plugins are located (default "plugins")
+  --repo string              Install plugin from specific repo (optional), if not passed iterate through all the repo and install the desire plugin. In case of conflicting name the command fails and asks user to specify the repo from which to install the plugin
```
sseago, line 204: "install the desire plugin" -> "install the desired plugin"

enhancements/crane-2.0/plugin-management/README.md
JaydipGabani: @sseago I feel this should not be needed for plugin-manifest, the purpose here is to have enough information to fetch appropriate binaries and not contain metadata about the plugin itself, I am of the opinion that we should ask for metadata information and not bombard the plugin-manifest file with information. Open to discussion on the benefits of adding this informatoin.

sseago: I'm not sure what you mean by "ask for metadata information" -- I don't think plugin-manager will be invoking plugins, since we can't invoke them until they're installed. for not-yet-installed plugins the only way to know what parameters a plugin takes would be for it to be in the manifest file. But yeah, we can discuss this next week if necessary to work out whether we want the plugin manager to be able to report on plugin parameters.

enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,322 @@
...
+The format of the yml file should be something like this
+
+
+```
+name: <plugin-name>
+shortDescription: <short description>
+description: <description>
+versions:
+	- version: <plugin-version>
+		flagName: <flag-name>
```
sseago, line 124: It looks like the other attributes for a version are indented more deeply than the first one. The indent was fine before, we just needed that extra dash.

sseago: Also, flagname/help/example are attributes of a single OptionalFields struct, but there can be more than one, so we should probably have an optionalFields attribute here, which points to a slice of OptionalFields objects.

enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,322 @@
...
+
+Usage:
+  crane plugin-manager list [flags]
+
+Flags:
+  -h, --help                 Help for plugin-manager list
+  --repo string              List plugins from specific repo (optional)
+  --installed bool           If this flag is passed, list all the installed plugins
+  -p, --plugin-dir string    The path where binary plugins are located (default "plugins")
+	--info bool								 If passed, returns with metadata information for all the version of specific plugin. This flag is to be used with "--name" flag
```
sseago, line 194: We should probably call this something like "optionals" or "parameters" or "optional-fields" rather than "info"

enhancements/crane-2.0/plugin-management/README.md
shawn-hurley: upgrading implies something has to happen in a sequence and some things need to be kept. In this case update, to me, means to move the binary forward. IDK if that distinction is correct? 

What do you think of this? @eriknelson @sseago @JaydipGabani

enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,322 @@
...
+At a high level plugin management functionality will be
+
+- Able to list the available plugins
+- Able to install/uninstall plugins
+- Able to upgrade any/all installed plugins (long term goal)
+- Able to add/remove custom repositories to list the plugins from (long term goal)
+
+## Motivation
+
+Apart from the offerings of crane-lib, we would like to add a way to allow users to extend the functionality of the tool by installing the managed or custom plugins. And we want users to use developed plugins as seamlessly as possible. We want them to consume out plugins and make it possible to easily do so.
```
shawn-hurley, line 42: nit: consume out plugins / consume out of binary plugins ???

enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,322 @@
...
+```
+
+The format of the yml file should be something like this
+
+
+```
+name: <plugin-name>
+shortDescription: <short description>
+description: <description>
+versions:
```
shawn-hurley, line 122: Can you help me understand the use case for multiple versions?

enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,322 @@
...
+
+```
+crane plugin-manager list --help
+
+Usage:
+  crane plugin-manager list [flags]
+
+Flags:
+  -h, --help                 Help for plugin-manager list
+  --repo string              List plugins from specific repo (optional)
```
shawn-hurley, line 191: While I agree we need this in the future, I don't want this to be in the help if it is not a goal of this particular enhancement

enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,322 @@
...
+  
+
+Global Flags:
+  --debug   Debug the command by printing more information
+
+```
+
+
+
+#### To support additional repository to install plugins
```
shawn-hurley, line 243: I think we should merge as is as this section is not in the goals of the enhancement. While I think it is good that we have some ideas on how to do this and they currently don't seem to block. Maybe we move this to a future considerations section?

shawn-hurley (commented):
I think we need to figure out two things then I am good with merging this:

1. How do we handle the default for plugin dir when installing, does this mean a change to the transform default as well?
2. the plugin manifest file can we discuss the need for multiple versions and for the binaries.

enhancements/crane-2.0/plugin-management/README.md
sseago: @shawn-hurley I think the inital slash there was a typo. I think the intent was "plugins" as the default -- i.e. relative to current directory. But yes, we could stop using a relative path as the default and instead put the default under $HOME -- although we'd want to do the same with export, transform, and output for consistency if we did that. Either way, it's overridable via `-p` or `--plugin-dir`, and either way, both plugin-manager and transform need to use the same values.

sseago:
@shawn-hurley Yes, for 1) transform and plugin-manager must use the same plugin-dir handling. If it's passed explicitly via -p, then use that, else use the default. The question is what's the default? Do we stick with `./plugins` as we're currently using for transform, or do we move to something in `~/.crane/plugins` If we do the latter, we need to also move the other defaults to `~/.crane/export`, etc.

enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,322 @@
...
+shortDescription: <short description>
+description: <description>
+versions:
+	- version: <plugin-version>
+		flagName: <flag-name>
+		help: <help-message>
+		example: <example>
+		platforms:
+		- selector:
+				matchLabels:
```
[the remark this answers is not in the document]
JaydipGabani: +1, this works

enhancements/crane-2.0/plugin-management/README.md
JaydipGabani: Ohh, makes sense. updating!

enhancements/crane-2.0/plugin-management/README.md
JaydipGabani: I feel in the future there could be a need where the specific version of  (could be the latest) binary of a plugin is not supported in a specific version of crane and the user might be in the need to install a particular version of a plugin. This solution can be handy at such times. It could happen with custom plugins or our plugins.

enhancements/crane-2.0/plugin-management/README.md
JaydipGabani: @shawn-hurley should I create a section saying "future scope" and mention that any functionality related to the custom plugin repository will not be available for initial release. However, develop the necessary skeleton in the code.

Or should I remove these descriptions altogether?

enhancements/crane-2.0/plugin-management/README.md
JaydipGabani: Done.

enhancements/crane-2.0/plugin-management/README.md
JaydipGabani: +1

enhancements/crane-2.0/plugin-management/README.md
```diff
@@ -0,0 +1,309 @@
...
+One of the proposed solution is to have a `plugins` github directory at the root of such repositories and respective `<plugin-name-version.yml>` inside the said directory from where we are listing plugins. 
+
+Example plugin repository layout:
+
+```
+.
+└── plugins/
+    ├── crane-x-1.0.yaml
+    ├── crane-y-2.4.yaml
+    └── crane-z-1.2.yaml
```
sseago, line 111: So to clarify -- in this case, the corresponding <plugin-name> in all 3 yaml files is "crane", and the versions are "x-1.0", "y-2.4", and "z-1.2"? I imagine the yaml itself should have a "version:" field with the corresponding information in it as well.

shawn-hurley (approved):
LGTM

eriknelson (approved):
Nice job on this @JaydipGabani. I'm excited to see this come together.

We're going to want to file a sequence of jiras to captures these stories, I'll catch up with you in slack about that.

@sseago @shawn-hurley if you guys are satisfied, let's merge this and get started!

sseago (approved):

PR #48, opened 2021-10-04
Propose changes to Migration Plan to improve storage conversion experience

pranavgaikwad:
This PR proposes new changes to Migration Plan API to improve UX of Storage conversion.

### DIFF
--- a/enhancements/crane-1.x/migplan-ux-improvements/README.md
+++ b/enhancements/crane-1.x/migplan-ux-improvements/README.md
@@ -1,0 +1,140 @@
+---
+title: migplan-ux-improvements
+authors:
+  - "@pranavgaikwad"
+reviewers:
+  - "@djwhatle"
+  - "@sseago"
+  - "@alaypatel07"
+  - "@shawn-hurley"
+approvers:
+  - "@djwhatle"
+  - "@sseago"
+  - "@alaypatel07"
+  - "@shawn-hurley"
+creation-date: 2021-10-04
+last-updated: 2021-10-04
+status: implementable
+see-also:
+  - "../state-only-migrations/README.md" 
+replaces:
+  - "N/A"
+superseded-by:
+  - "N/A"
+---
+
+# MigPlan UX Improvements
+
+This enhancement proposes API changes to build a full end-to-end UX around _Storage Conversion_. The document explains how proposed changes consolidate different types of migrations in one place and provide a better overall experience for all three types of migrations currently available in MTC.
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [x] Design details are appropriately documented from clear requirements
+- [x] Test plan is defined
+- [ ] User-facing documentation is created
+
+## Open Questions [optional]
+
+This is where to call out areas of the design that require closure before deciding
+to implement the design.  For instance, 
+ > 1. This requires exposing previously private resources which contain sensitive
+  information.  Can we do this? 
+
+## Summary
+
+In MTC 1.7, we envision to implement first class UX around _Storage Conversion_ where the users will use MTC to change the underlying storage of their PVCs. _Storage Conversion_ is already possible in MTC 1.6, but it doesn't have a specific guided experience. The users need to use _State Migration_ in a certain way to achieve storage conversion. Currently, the PV selection and the discovery part in State Migration takes place in the existing _MigPlan_ API. There are no changes made in the API itself for state migrations. Unlike usual migrations, both _state migration_ and the _storage conversion_ are not designed for migrating entire namespaces. The user experience of current _MigPlan_ is suitable for migrations of entire namespaces. There are problems when it comes to using the same API for _State Migrations_ or _Storage Conversion_. With minor changes to API, we can build first class experience around _Storage Conversion_. Similar changes will also improve UX of _State Migrations_. This enhancement mainly highlights user experience issues around _Storage Conversion_ and proposes changes in _MigPlan_ API to consolidate the experience of different types on migrations in one place.
+
+## Motivation
+
+_Storage Conversion_ will allow users to convert underlying storage of PVCs. The internal mechanism will be handled through existing _State Migration_. With MTC 1.6, storage conversion is already possible. However, there is no obvious way for users to know how to achieve it without supporting documentation. There are details that users may not necessarily know (such as selecting the same clusters, not mapping namespaces, mapping PVCs to different names, etc). We need to provide a full end-to-end experience for storage conversion. Also, there are two different options the users can convert storage of their PVCs using _State Migration_. We want to give them only one option. For that, it is necessary to make some changes in the _MigPlan_ API. It will facilitate providing better warning messaging and validations based on type of migration being performed. Similar changes will also improve UX of _State Migrations_. 
+
+### Goals
+
+> Propose solution to implement first class user experience around storage conversion
+
+> Propose approach to consolidate different types of migrations through Migration Plan
+
+> Explain how proposes changes will improve existing state migration experience
+
+### Non-Goals
+
+> Propose controller changes for storage conversion
+
+## Proposal
+
+* _Storage conversion_ is already possible in MTC 1.6 but it doesn't have a guided experience. The users need to set the right namespace and PV mappings in the Migration Plan, set the source & destination clusters to the same cluster, map storage classes and then perform a _State Migration_ to achieve storage conversion. For end users, this is hard to figure out without reading detailed supporting documentation. If we could mark a _MigPlan_ to be exclusively used for storage conversion, we can automatically choose the source, destination _MigCluster_ resources. We can also choose the namespace mappings correctly without the user having to provide it. Additionally, we can implement specific validations based on the type of the _MigPlan_. This wouldn't be possible if we don't know which type of migration the current plan will be used for.
+
+* If the users perform a _State Migration_ by selecting a subset of PVCs in the plan, those settings are persisted on the plan itself. However, that _MigPlan_ is still available for normal migrations. This presents with some edge cases that are hard to guard against. For instance, if a PVC is skipped from a state migration, it remains skipped during a final migration as well. But if a workload is using that PVC, it would fail to come up in the destination cluster. It would be hard for a user to reach the root cause of the issue if they do not know the details. On the other hand, if we could use the _MigPlan_ exclusively for _State Migration_ we can disable its use for normal migrations and vice versa. Based on the `IncludedResources` API field, we can generate helpful warnings on _MigPlan_ beforehand.
+
+* Based on the type of the migration, we can also enable/disable indirect migration and hooks options. Also, for _State Migration_ we can provide users with further options for configuring permission/SCC information which wouldn't be available for normal migrations. The details of permission handling will be covered in a different enhancement.
+
+_MigPlan_ API will mainly introduce two new fields:
+
+```yaml
+// MigPlanSpec defines the desired state of MigPlan
+type MigPlanSpec struct {
+	[...]
+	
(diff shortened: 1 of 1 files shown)


enhancements/crane-1.x/migplan-ux-improvements/README.md
```diff
@@ -0,0 +1,140 @@
...
+```yaml
+// MigPlanSpec defines the desired state of MigPlan
+type MigPlanSpec struct {
+	[...]
+	
+	// Specifies whether migration plan will be used for state migration
+	IsStateMigration  bool `json:"isStateMigration,omitempty"`
+	
+	// Specifies whether migration plan will be used for storage conversion
+	IsStorageConversion  bool `json:"isStorageConversion,omitempty"`
```
sseago, line 83: Is it valid to select both state migration and storage conversion? i.e. are there 3 possible values here ("normal, "state migration", "storage conversion") or four (two independent boolean values)? If there are only 3 values, would it be better to have a "type" field with "normal", "state", and "storage conversion" as valid values?

sseago (commented):
In addition to the earlier question, are these two options going to be exposed via the migplan wizard, or does that UI create "normal" migrations only?

enhancements/crane-1.x/migplan-ux-improvements/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+* If the users perform a _State Migration_ by selecting a subset of PVCs in the plan, those settings are persisted on the plan itself. However, that _MigPlan_ is still available for normal migrations. This presents with some edge cases that are hard to guard against. For instance, if a PVC is skipped from a state migration, it remains skipped during a final migration as well. But if a workload is using that PVC, it would fail to come up in the destination cluster. It would be hard for a user to reach the root cause of the issue if they do not know the details. On the other hand, if we could use the _MigPlan_ exclusively for _State Migration_ we can disable its use for normal migrations and vice versa. Based on the `IncludedResources` API field, we can generate helpful warnings on _MigPlan_ beforehand.
+
+* Based on the type of the migration, we can also enable/disable indirect migration and hooks options. Also, for _State Migration_ we can provide users with further options for configuring permission/SCC information which wouldn't be available for normal migrations. The details of permission handling will be covered in a different enhancement.
```
shawn-hurley, line 70: Is there a way to infer this from other fields? Is this specifically for the UI to know what to show or are there going to be new validations on the mig migration? I am having trouble understand from an API perspective how these fields help me understand how to do X or Y.

shawn-hurley (commented):
I am wondering if there is something else that can be done here than adding this field to the API. If I am an API user, I don't think that these fields help me understand what to do and I think that it may cause more confusion, because if I understand correctly if I choose to do some things now my plan will be invalid?

enhancements/crane-1.x/migplan-ux-improvements/README.md
djwhatle: +1 to that Scott. I think an early example of API short-sightedness was introduction of the `isStage` field, which made the API a bit awkward when we later introduced other types of migrations (e.g. Rollback).

Having a type field like you suggest makes sense.

enhancements/crane-1.x/migplan-ux-improvements/README.md
```diff
@@ -0,0 +1,140 @@
...
+	
+	// Specifies whether migration plan will be used for state migration
+	IsStateMigration  bool `json:"isStateMigration,omitempty"`
+	
+	// Specifies whether migration plan will be used for storage conversion
+	IsStorageConversion  bool `json:"isStorageConversion,omitempty"`
+}
+```
+
+These fields will tie every plan to a specific type of migration. The plan cannot be used for a different type of migration once that field is set. The field can only be changed if there are no migrations assigned with it.
```
djwhatle, line 87: How will we enforce that a plan cannot be used for different types of migrations later?

enhancements/crane-1.x/migplan-ux-improvements/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+> Explain how proposes changes will improve existing state migration experience
+
+### Non-Goals
+
+> Propose controller changes for storage conversion
+
+## Proposal
+
+* _Storage conversion_ is already possible in MTC 1.6 but it doesn't have a guided experience. The users need to set the right namespace and PV mappings in the Migration Plan, set the source & destination clusters to the same cluster, map storage classes and then perform a _State Migration_ to achieve storage conversion. For end users, this is hard to figure out without reading detailed supporting documentation. If we could mark a _MigPlan_ to be exclusively used for storage conversion, we can automatically choose the source, destination _MigCluster_ resources. We can also choose the namespace mappings correctly without the user having to provide it. Additionally, we can implement specific validations based on the type of the _MigPlan_. This wouldn't be possible if we don't know which type of migration the current plan will be used for.
```
djwhatle, line 66: At a high level, what in the migplan logic is influenced by changing a migplan to these other types?

enhancements/crane-1.x/migplan-ux-improvements/README.md
pranavgaikwad: @djwhatle @sseago I agree. I thought about having a Typed field, but reconciling existing MigPlans from older versions would be difficult if we introduce a typed field instead of a boolean. So absence of the boolean != absence of the typed field. If we can find a better solution to the problem (perhaps, set default to normal migrations) we should definitely make it a Typed field.

enhancements/crane-1.x/migplan-ux-improvements/README.md
pranavgaikwad: @shawn-hurley Its for UI and backend both. Imagine, in the API, there are various ways the user can configure a MigPlan for storage conversion. Two main examples are - 1. Storage convert by changing the namespace 2. Storage convert by preserving the namespace but changing the PV names. We want to standardize and support only one workflow (we decided to go with no.2). So as soon as the user indicates a "type" of migration the plan will be used for, we will pre-fill rest of the values (setting source and destination cluster to the same, omission of indirect options), the controller will trigger specific validations for Storage Conversion. In future, we will also leverage the intra-cluster State Migration for storage conversion that you worked on (which is based on Service instead of Route)

shawn-hurley: > we will pre-fill rest of the values (setting source and destination cluster to the same, omission of indirect options), the controller will trigger specific validations for Storage Conversion. In future, we will also leverage the intra-cluster State Migration for storage conversion that you worked on (which is based on Service instead of Route)

How does this happen for an API user?

enhancements/crane-1.x/migplan-ux-improvements/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+* If the users perform a _State Migration_ by selecting a subset of PVCs in the plan, those settings are persisted on the plan itself. However, that _MigPlan_ is still available for normal migrations. This presents with some edge cases that are hard to guard against. For instance, if a PVC is skipped from a state migration, it remains skipped during a final migration as well. But if a workload is using that PVC, it would fail to come up in the destination cluster. It would be hard for a user to reach the root cause of the issue if they do not know the details. On the other hand, if we could use the _MigPlan_ exclusively for _State Migration_ we can disable its use for normal migrations and vice versa. Based on the `IncludedResources` API field, we can generate helpful warnings on _MigPlan_ beforehand.
```
alaypatel07, line 68:
```
For instance, if a PVC is skipped from a state migration, it remains skipped during a final migration as well. But if a workload is using that PVC, it would fail to come up in the destination cluster.
```

Isnt this true for normal migplans as well? if a user selects 1/2 of the pvc for a plan to "stage" it and then runs final migration, they'll have workloads not coming up? so this is an existing problem not just for state migration?

enhancements/crane-1.x/migplan-ux-improvements/README.md
eriknelson: +1 I came to ask the same question

eriknelson (changes requested):

enhancements/crane-1.x/migplan-ux-improvements/README.md
sseago: @alaypatel07 The difference is that at least at one point the regular UI only allowed skip for PVCs that didn't belong to applications -- i.e. not attached to either a running pod or a quiesced Deployment/DC/etc. However, to support state migration I think we opened that up to allow skipping of all PVCs, so yes, if that is the case, this is an issue for regular migration as well.

pranavgaikwad: @alaypatel07 no its not. For normal migplans, we cannot select PVs.

enhancements/crane-1.x/migplan-ux-improvements/README.md
```diff
@@ -0,0 +1,140 @@
...
+- [ ] Enhancement is `implementable`
+- [x] Design details are appropriately documented from clear requirements
+- [x] Test plan is defined
+- [ ] User-facing documentation is created
+
+## Open Questions [optional]
+
+This is where to call out areas of the design that require closure before deciding
+to implement the design.  For instance, 
+ > 1. This requires exposing previously private resources which contain sensitive
```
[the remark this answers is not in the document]
pranavgaikwad: Just template text

enhancements/crane-1.x/migplan-ux-improvements/README.md
sseago: @pranavgaikwad at least for PVs that are not bound to pods/deployments, they *can* be selected to skip in the wizard. Also, I thought there was a PR along the way that updated the available actions to include "Skip" for everything, although maybe that didn't get merged?

enhancements/crane-1.x/migplan-ux-improvements/README.md
```diff
@@ -0,0 +1,181 @@
...
+type MigMigrationSpec struct {
+	Type *MigrationType `json:"storageConversion,omitempty"`
+}
+
+type MigrationType string
+```
+
+MigrationType has four possible values: `Final`, `Stage`, `StateTransfer`, `StorageConversion`, `RollBack`.
+
+This will present some problems with existing _MigMigrations_. Since migrations are supposed to reach to completion, these problems are relatively easier to solve.
```
shawn-hurley, line 127: what happens if I have a Migration YAML that I keep using? if this field is set to `nil` what is the behavior?

enhancements/crane-1.x/migplan-ux-improvements/README.md
```diff
@@ -0,0 +1,181 @@
...
+}
+
+type MigrationType string
+```
+
+MigrationType has four possible values: `Final`, `Stage`, `StateTransfer`, `StorageConversion`, `RollBack`.
+
+This will present some problems with existing _MigMigrations_. Since migrations are supposed to reach to completion, these problems are relatively easier to solve. 
+
+Existing CRs will not have the type field set, instead they will have the old booleans. In 1.7, we can deprecate the booleans and add controller logic to correctly set the Type field based on the old booleans. Then in later versions, we can remove the booleans altogether. Downgrades will be only possible from 1.7 to lower versions.
```
shawn-hurley, line 129: +1 this seems like the correct approach, have we considered a validating or mutating admission webhook to do this?

enhancements/crane-1.x/migplan-ux-improvements/README.md
alaypatel07: yes, I think that is a valid missing scenario, seems like the best way to handle that would be a webhook. The question is how hard is it to build and maintain webhook? do we want to take that pain given the strategic direction we are heading to? Unfortunately, I do not know the answer to those questions.

PR #54, opened 2021-11-23
adding enhancement for management of custom plugin repositories

JaydipGabani:
This enhancement proposes solutions for:
- Adding/Removing custom crane-plugin repository
- Installing/Uninstalling/Listing plugins from custom crane-plugin repository

### DIFF
--- a/enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
+++ b/enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
@@ -1,0 +1,163 @@
+---
+title: Manage custom plugin repository
+authors:
+  - "@jgabani"
+reviewers:
+  - "@sseago"
+approvers:
+  - "@sseago"
+creation-date: 2021-11-23
+last-updated: 2021-11-23
+status: provisional|implementable|implemented|deferred|rejected|withdrawn|replaced
+---
+
+# Manage custom plugin repository
+
+This document proposes a solution to manage custom plugin repositories. The document describes the needed changes in current implementation, and all the sub-commands that are needed for this feature.
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] User-facing documentation is created
+
+## Open Questions
+
+- Do we want the user to have the control to override/remove default plugin source?
+
+## Summary
+
+Along with the our own default plugin-repository, we would like to give the users an option to add their own plugin-repositories from which they can discover, install, and uninstall any custom plugins. If the custom repository is defined within the constrained mentioned [here](https://github.com/konveyor/crane-plugins/blob/main/README.md) and looks like the default [crane-plugins](https://raw.githubusercontent.com/konveyor/crane-plugins/main/index.yml) repository, then the goal is that a user should be able add such repository and crane should be able to look at this added repository same as it looks at default crane-plugins.
+
+## Motivation
+
+The ability to add/remove custom crane-plugin repositories is needed in a restrictive environment where the access to the repository from outside the network is revoked, as well as for environments where custom plugins not provided in the crane-plugins repo are needed. In such cases, if a user can add their own custom-plugin repository then it would be beneficial for them to leverage `crane plugin-manager` command to manage plugins.
+
+### Goals
+
+- User should be able to add any plugin repository
+- User should be able to remove any plugin repository
+- User should be able to discover plugins from plugin repository
+- User should be able to install plugins from plugin repository
+- User should be able to uninstall plugins installed from added plugin repository
+- User should be able to list all the plugin repositories available
+
+### Non-Goals
+
+Not applicable
+
+## Proposal
+
+We provide new subcommands, `add-repo|remove-repo|list-repos`, to `crane plugin-manager` for managing custom plugin repositories.
+
+### User Stories
+
+#### Story 1
+
+As a user, I would like to manage plugin repositories (add|list|remove)
+
+#### Story 2
+
+As a user, I would like to manage my plugins using configured plugin repositories (install|uninstall|list)
+
+### Risks and Mitigations
+
+#### Risks
+- If the user doesn't setup the custom plugin repository in the desired manner, the user might face some issues working with `crane plugin-manager`
+- If the user overrides the setting for default plugin repo and replace it with nil, `crane plugin-manager` might become useless
+
+#### Mitigations
+- Testing the feature thoroughly to make sure that we don't break `crane plugin-manager list/add/remove` 
+- Provide user with sufficient documentation on how to setup custom plugin repositories, how to develop custom plugins and how to properly add/remove custom crane-plugin repository
+- A validation against a plugin repo that ensures a plugin-repository (that `crane` refers to) is valid. Meaning that added custom plugin repository have valid plugin manifests.
+
+
+## Design Details
+
+### Prerequisites
+
+- User have their own plugins or modified version of default provided plugins developed in a similar manner as [this](https://github.com/konveyor/crane-plugin-openshift) plugin.
+- User have a custom plugin repository setup in similar manner as [crane-plugin](https://raw.githubusercontent.com/konveyor/crane-plugins/main/index.yml)
+
+#### Solution
+
+##### Part 1
+
+A config file `repos.yaml`, placed in `--plugin-dir`. This file contains plugin repository name as key and respective URL as value. Absence of this file in a plugin directory would mean crane would create a new repos.yaml in the plugin directory linking only to the crane-plugins repo. However, if an empty `repos.yaml` file is present means user need to add at least one plugin repository such as [crane-plugins](https://github.com/konveyor/crane-plugins). It makes sense to maintain `repos.yaml` for different `--plugin-dir` since, different `--plugin-dir` would have different set of plugins potentially installed from different sources.
+
+- The file `repos.yaml` would look like following:
+  
+```
+foo: <accessible location of an index.yaml file>
+bar: <accessible location of an index.yaml file>
+```
+
+##### Part 2
+
+We are proposing three new sub commands under `crane plugin-manager` to manage custom repositories (i.e `repos.yaml` file).
+
+- To add custom crane-plugin repository
+
+```
+crane plugin-manager add-repo -h
+to add custom crane-plugin repository
+
+Usage:
+  crane plugin-manager add-repo <name> <URL> [flags]
+
+Flags:
+  -h, --help   help for add-repo
+
+Global Flags:
+  -p, --plugin-dir string    The path where binary plugins are located (default "plugins")
+```
+
+On high level this command will look at `repos.yaml` located in local plugin directory, and add an entry with respective name and URL passed in the command. If passed name or URL is already added on the file then notifies user of the same.
+
+- To remove custom crane-plugin repository
+```
+crane plugin-manager remove-repo -h
+to remove custom crane-plugin repository
+
+Usage:
+  crane plugin-manager remove-repo <name> [flags]
+
+Flags:
+  -h, --help   help for remove-repo
+
+Global Flags:
+  -p, --plugin-dir string    The path where binary plugins are located (default "plugins")
+```
+The command will remove the entry from `repos.yaml` located in local plugin directory, if the respective name is not present then notifies the user of the same. Removing a plugin repository from `repos.yaml` also results in removing all the plugins installed from respective plugin repository, 
+
+- To list all crane-plugin repositories
+
+```
+crane plugin-manager list-repos -h
+to list all crane-plugin repository
+
+Usage:
+  crane plugin-manager list-repos [flags]
+
+Flags:
+  -h, --help   help for list-repos
+
+Global Flags:
+  -p, --plugin-dir string    The path where binary plugins are located (default "plugins")
+```
+
+- Similarly `crane plugin-manager list, add` will also look at the same config file to discover plugins from, to install plugins from, when used with `--repo` flag and otherwise when there is no conflicts of plugin name. 
+  
+**Note:** 
+- All above command will create `repos.yaml` if it does not exists with a single entry that is - `crane-plugins: https://raw.githubusercontent.com/konveyor/crane-plugins/crane-cli/index.yml` a path to our crane-plugins.
+- To update an entry in `repos.yaml`, first the repo needs to be removed(i.e `crane plugin-manager remove-repo foo -p bar`), and then needs to be added with updated path(i.e `crane plugin-manager add-repo foo <path> -p bar`)
+
+
+## Drawbacks
+
+By giving the flexibility of having a custom plugin repository, user have much more power to crane and have much higher chances of breaking `crane plugin-manager` because of misconfigurations on custom plugin repositories.
+
+## Alternatives
+
+Instead of maintaining a file in path `--plugin-dir` we ask users to create a file with the name `repos.yaml` and place it in `--plugin-dir` with appropriate entries, and we list, add and remove plugins by reading all the repos mentioned in the file `repos.yaml`. In this approach there is no need to add any commands, however the responsibility to create an appropriate config file for `crane` to consume lies on the user.


enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
```diff
@@ -0,0 +1,162 @@
...
+
+Usage:
+  crane plugin-manager add-repo <name> <URL> [flags]
+
+Flags:
+  -h, --help   help for add-repo
+
+```
+
+On high level this command will look at a file placed at `home/user/.conf/crane.yaml`, and add an entry with respective name and URL passed in the command. If passed name or URL is already added on the file then notifies user of the same.
```
djzager, line 116: `/home/user/.conf/crane.yaml` seems like an odd location, maybe you want to be using `$XDG_CONFIG_HOME` (ie. https://wiki.archlinux.org/title/XDG_Base_Directory ). I could see `$XDG_CONFIG_HOME/crane/config.yaml`.

Has this file been mentioned previously? What relation does this have to:

```
Global Flags:
  -c, --config-file string   Path to config file which contains a yaml representation of cli flags. Explicit flags take precedence over config file values.
```

Can the location of this file be overridden with some analogue to the `KUBECONFIG` (ie `CRANECONFIG`)?

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
```diff
@@ -0,0 +1,162 @@
...
+bar: <github raw url for an index.yaml file>
+```
+
+## Drawbacks
+
+By giving the flexibility of having a custom plugin repository, user have much more power to crane and have much higher chances of breaking `crane plugin-manager` because of misconfigurations on custom plugin repositories.
+
+## Alternatives
+
+Instead of maintaining a file we ask users to create such file with the name `crane.yaml` and place it under a specific path with appropriate entries, and we list, add and remove plugins by reading all the repos mentioned in the file `crane.yaml`. In this approach there is no need to add any commands, however the responsibility to create an appropriate config file for `crane` to consume lies on the user.
```
djzager, line 162: This review is the first time I've heard of this `crane.yaml`...has it been implemented already or would it be a part of this effort? Could you provide more detail on this `crane.yaml`...it's relation to the config-file that can be used for individual commands...and what this file will look like?

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
```diff
@@ -0,0 +1,162 @@
...
+The command will remove the entry from `home/<user>/.conf/crane.yaml`, if the respective name is not present then notifies the user of the same.
+
+- To list all crane-plugin repositories
+
+```
+crane plugin-manager list-repo -h
+to list all crane-plugin repository
+
+Usage:
+  crane plugin-manager list-repo [flags]
```
djzager, line 138: Maybe `list-repos`?

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
```diff
@@ -0,0 +1,162 @@
+---
+title: neat-enhancement-idea
```
djzager, line 2: Should give it an actual title :sunglasses:

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
djzager: Yeah. The way this reads it sounds like something that has to be included with the install of the binary, specifically this part:
> On high level this command will look at a file placed at `home/user/.conf/crane.yaml`

What happens if that file isn't there? Do we create it?

djzager (commented):
Thanks for this effort. I think being able to have custom plugin repositories is a great idea but I do wonder if we can support this without adding a bunch of new subcommands we have to support...at leeast we could consider waiting until we know what is painful about managing it manually.

Some general feedback included, I would like the idea of `crane.yaml` to be given more detail before going in for another round of feedback.

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
```diff
@@ -0,0 +1,162 @@
...
+
+### Non-Goals
+
+Not applicable
+
+## Proposal
+
+We will provide cli commands within `crane plugin-manager` to add and remove any custom plugin repository.
+
+### User Stories [optional]
```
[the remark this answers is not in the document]
JaydipGabani: :+1:

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
```diff
@@ -0,0 +1,162 @@
...
+- User should be able to uninstall plugins installed from custom plugin repository
+- User should be able to list all the plugin repositories available
+
+### Non-Goals
+
+Not applicable
+
+## Proposal
+
+We will provide cli commands within `crane plugin-manager` to add and remove any custom plugin repository.
```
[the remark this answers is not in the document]
JaydipGabani: :+1:

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
```diff
@@ -0,0 +1,162 @@
...
+## Design Details
+
+### Prerequisites
+
+- User have their own plugins or modified version of default provided plugins developed in a similar manner as [this](https://github.com/konveyor/crane-plugin-openshift) plugin.
+- User have a custom plugin repository setup in similar manner as [crane-plugin](https://github.com/konveyor/crane-plugins)
+
+#### Solution
+
+We are proposing two new sub commands under `crane plugin-manager` to manage custom repositories.
```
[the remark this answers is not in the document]
JaydipGabani: :+1:

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
JaydipGabani: :+1:

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
JaydipGabani: :+1:

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
JaydipGabani: `crane.yaml` is a config file that contains a map with the key being the name of a plugin-repository (I.e `default`, `my-crane-plugins` etc.) and its value being URL/path of the repository. It is the same file we are maintaining through proposed commands in this enhancement (i.e add-repo. remove-repo). This is a common config file for `crane` rather than a config file containing flag inputs to be consumed by `crane` commands.

JaydipGabani:
> Thanks for this effort. I think being able to have custom plugin repositories is a great idea but I do wonder if we can support this without adding a bunch of new subcommands we have to support...at leeast we could consider waiting until we know what is painful about managing it manually.
> 
> Some general feedback included, I would like the idea of `crane.yaml` to be given more detail before going in for another round of feedback.

Yeah, I discussed an alternative to adding these subcommands in this enhancement. In this case, one solution that I could think of is that we ask users to create a file with the name `crane.yaml` and place it at `$XDG_CONFIG_HOME/crane/` with appropriate entries, and we list, add and remove plugins by reading all the repos mentioned in the file `$XDG_CONFIG_HOME/crane/crane.yaml`. In this approach there is no need to add any commands, however the responsibility to create an appropriate config file for `crane` to consume lies on the user.

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
JaydipGabani: 1. No we have not thought of `$XDG_CONFIG_HOME/crane/crane.yaml` before. Its config file for `crane` not for individual commands (i.e with flag `-c, --config-file`). Right now the location of the config file is predetermined which is as you suggested `$XDG_CONFIG_HOME/crane/crane.yaml`, however we can have an option to override with `CRANECONFIG`, if desired.

2. Yeah, if the file is not already present then we create the file

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
```diff
@@ -0,0 +1,162 @@
...
+
+### Risks and Mitigations
+
+#### Risks
+- If the user doesn't setup the custom plugin repository in the desired manner, the user might face some issues working with `crane plugin-manager`
+- If the user overrides the setting for default plugin repo and replace it with nil, `crane plugin-manager` might become useless
+
+#### Mitigations
+- Testing the feature thoroughly to make sure that we don't break `crane plugin-manager list/add/remove` 
+- Provide user with sufficient documentation on how to setup custom plugin repositories, how to develop custom plugins and how to properly add/remove custom crane-plugin repository
```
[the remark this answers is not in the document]
djzager: An example of a mitigation to an existing risk would be validating the user's input before accepting it as a plugin repo.

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
```diff
@@ -0,0 +1,148 @@
...
+
+Usage:
+  crane plugin-manager add-repo <name> <URL> [flags]
+
+Flags:
+  -h, --help   help for add-repo
+
+```
+
+On high level this command will look at a file placed at `$XDG_CONFIG_HOME/crane/crane.yaml`, and add an entry with respective name and URL passed in the command. If passed name or URL is already added on the file then notifies user of the same.
```
djzager, line 101: My understanding is that this config file (that would probably make more sense as `$XDG_CONFIG_HOME/crane/config.yaml` ...total nitpick) is totally separate from the `--config-file` you would use at the command line. 

Am I understanding this clearly? I'm really concerned that having multiple configuration files that do different things will be incredibly confusing to all parties involved.

If this configuration file is itself a net new idea, I believe it deserves it's own enhancement that we can discuss in depth.

djzager (changes requested):
Assuming I understand the config file mentioned correctly, I do believe that serious consideration should be given to whether it deserves an enhancement entirely to itself.

I have reservations about the idea that we have a configuration file for the plugin-manager and a separate (optional) configuration file that controls `cranes` behavior, that I think would be best hashed out in it's own PR.

Let me know what you think @JaydipGabani @sseago ... maybe we can chat about this in scrum?

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
```diff
@@ -0,0 +1,162 @@
...
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] User-facing documentation is created
+
+## Open Questions [optional]
+
+- Do we want the user to have the control to override/remove default plugin source?
```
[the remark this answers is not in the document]
sseago: I think the current env var override for the default source is a workaround for the fact that add/remove plugin repos doesn't exist yet. I think once we have the ability to add custom repos, the way to override the default location would be simply to add the desired one -- possibly also removing the default one, if we allow that). In the crane-runner use case, this would mean adding the *local* plugin repo and removing/disabling the default github-based one.

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
sseago: So the existing "config file" is really more of a "user input file" -- it (partially or completely) replaces the full set of flags passed into a particular crane invocation. This file is actually storing state *between* crane invocations -- i.e. a repo added in one invocation is available in the next. Maybe instead of config.yaml it should be repos.yaml? Also, since adding/removing plugins is handled by manipulating subdiretories of <plugin-dir>, I think this repositories.yaml file should probably live within the <plugin-dir> itself. <plugin-dir>/repositories.yaml would list the repos, <plugin-dir>/managed/<repo> would contain installed plugins for a given repo, etc.

sseago: Since it's not a general-purpose "configuration file" -- it's really just a list of installed repos in a particular plugin dir, I don't really think it warrants its own enhancement, assuming that its scope is limited to what I proposed above and clearly defined in this enhancement.

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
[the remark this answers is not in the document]
sseago: So we can support removing the default repo by adding a bool value to the repos.yaml file in <plugin-dir> (see below for context on this) -- something like `defaultRepoDisabled`, where calling `remove-repo` with the default repo specified would set this bool.

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
djzager: > So the existing "config file" is really more of a "user input file" -- it (partially or completely) replaces the full set of flags passed into a particular crane invocation. 

So long as we appreciate the potential confusion that is `--config-file ${NOT_REALLY_A_CONFIG_FILE}` :grin: , I'm good.

> This file is actually storing state between crane invocations -- i.e. a repo added in one invocation is available in the next. Maybe instead of config.yaml it should be repos.yaml?

I think calling it `repos.yaml` is a great idea.

> Also, since adding/removing plugins is handled by manipulating subdiretories of , I think this repositories.yaml file should probably live within the itself. /repositories.yaml would list the repos, /managed/ would contain installed plugins for a given repo, etc.

I'm not fully aware of the precedence for this, but having it drop a `repos.yaml` (I like it and I'm sticking with it) in the current working directory eschews convention...right?

Maybe we could use this specifically https://pkg.go.dev/os#UserConfigDir to determine where to place this crane repos configuration file:

`craneRepos := os.UserConfigDir() + "/crane/repos.yaml"`

> I don't really think it warrants its own enhancement

I think that you are right, if this config file is scaled back to only dealing with the configured repositories then it doesn't need it's own. The only caveat I would add is that it does deserve it's own subsection in this enhancement (before digging into the subcommands) to fully describe what it is + where it lives + what it does.

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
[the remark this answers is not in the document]
JaydipGabani: [Here](https://github.com/konveyor/crane/blob/main/internal/plugin/plugin_manager_helper.go#L142) is the setting to override default source of the plugin repository. Right now there is one source always present. If the env variable is not set then we point to default source that is `konveyor/crane-plugins`.

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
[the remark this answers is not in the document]
JaydipGabani: In this case, I feel we would need to define what is an acceptable validation of custom plugin repo looks like, 

Do we validate if the path provided is a valid one for repo?

Do we check if the structure of the custom plugin repository is defined as required?

Do we check the content of `index.yml` file and `manifests` and validate the proper format and entries exist there?

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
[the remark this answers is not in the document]
sseago: @JaydipGabani I'm thinking that once we have the ability to add repos we can remove this env var. If a user (or crane-runner, as it will need this) needs to override, the right thing to do will be to remove the default and add the custom one.

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
sseago: @djzager s/current working directory/<plugin-dir>/

sseago:
```
Maybe we could use this specifically https://pkg.go.dev/os#UserConfigDir to determine where to place this crane repos configuration file:

```
I don't think this is correct. I think we want this to be the plugin dir that we're explicitly dealing with.

sseago: As for "**NOT_REALLY_A_CONFIG_FILE**" -- do we want to call this flag `--input-file` instead?

sseago: The point is that the list of installed plugins is different based on a different passed-in `--plugin-dir` -- I think the list of installed repos should be similarly modified by that param.

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
```diff
@@ -0,0 +1,163 @@
...
+
+As a user, I would like to manage plugin repositories (add|list|remove)
+
+#### Story 2
+
+As a user, I would like to manage my plugins using configured plugin repositories (install|uninstall|list)
+
+### Risks and Mitigations
+
+#### Risks
```
shawn-hurley, line 66: What other risks are there? do we have upgrade risks how do we handle when the crane CLI is updated and the plugin repo no longer supports the crane CLI version (request/response)?

What about dealing with plugins having the same name? How do we handle naming conflicts? can repos have the same name?

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
```diff
@@ -0,0 +1,163 @@
...
+- If the user doesn't setup the custom plugin repository in the desired manner, the user might face some issues working with `crane plugin-manager`
+- If the user overrides the setting for default plugin repo and replace it with nil, `crane plugin-manager` might become useless
+
+#### Mitigations
+- Testing the feature thoroughly to make sure that we don't break `crane plugin-manager list/add/remove` 
+- Provide user with sufficient documentation on how to setup custom plugin repositories, how to develop custom plugins and how to properly add/remove custom crane-plugin repository
+- A validation against a plugin repo that ensures a plugin-repository (that `crane` refers to) is valid.
```
shawn-hurley, line 73: How are we exposing errors when they occur with a plugin manager and a repo?

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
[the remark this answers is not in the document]
djzager: I'm not sure how far we go here. It's more that I do believe the mitigations we list should address the risks. So, if we say:
> If the user doesn't setup the custom plugin repository in the desired manner, the user might face some issues working with `crane plugin-manager`

How are we planning to mitigate this specific risk? We could validate., but how much can we validate? This is the place for you to go into these details about all the validations we could do and the necessary cost/benefit analysis. Once you have that you can say, in the enhancement, what level of validation you think is appropriate based on cost/benefit analysis and reviewers can share their thoughts.

In the end, we have a historical record of this conversation of all the risks/mitigations and relevent cost/benefit analysis to show why we ended with the implementation we decided.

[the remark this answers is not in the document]
sseago: "how much can we validate" -- so currently we parse the index file for the list action. One possibility is that we could use the same parsing code to validate the passed-in URL on `add-repo` -- if it returns a list of plugins (internally, not for display to the user), then we declare it good. If the action fails, then we error out as "not a valid plugin repo".

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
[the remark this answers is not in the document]
sseago: So based on prior discussion, the "default" plugin source is merely a "crane-plugins" plugin source that we automatically add to the template repos.yaml on creation, which means we can remove, re-add it just like any other plugin repo once created.

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
```diff
@@ -0,0 +1,163 @@
...
+- [ ] Test plan is defined
+- [ ] User-facing documentation is created
+
+## Open Questions
+
+- Do we want the user to have the control to override/remove default plugin source?
+
+## Summary
+
+Along with the our own default plugin-repository, we would like to give the users an option to add their own plugin-repositories from which they can discover, install, and uninstall any custom plugins. If the custom repository is defined within the constrained mentioned [here](https://github.com/konveyor/crane-plugins/blob/main/README.md) and looks like the default [crane-plugins](https://github.com/konveyor/crane-plugins) repository, then the goal is that a user should be able add such repository and crane should be able to look at this added repository same as it looks at default crane-plugins.
```
sseago, line 31: Rather than reference the github repo path, reference the actual path that contains the index file: https://raw.githubusercontent.com/konveyor/crane-plugins/crane-cli/index.yml

sseago: It's the file content that matters here, not the fact that it's a github repo.

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Motivation
+
+The ability to add/remove custom crane-plugin repositories is needed in a restrictive environment where the access to the repository from outside the network is revoked. In such cases, if a user can add their own custom-plugin repository then it would be beneficial for them to leverage `crane plugin-manager` command to manage plugins.
```
sseago, line 35: "...in a restrictive environment where the access to the repository from outside the network is revoked, as well as for environments where custom plugins not provided in the crane-plugins repo are needed."

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
```diff
@@ -0,0 +1,163 @@
...
+
+### Goals
+
+- User should be able to add any custom plugin repository
```
sseago, line 39: I'm thinking we remove "custom" here (and below), since they can add/remove *any* plugin repository, custom or provided-by-konveyor.

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
```diff
@@ -0,0 +1,163 @@
...
+- User should be able to remove any custom plugin repository
+- User should be able to discover plugins from custom plugin repository
```
sseago, line 41: s/custom/added/

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
```diff
@@ -0,0 +1,163 @@
...
+- Provide user with sufficient documentation on how to setup custom plugin repositories, how to develop custom plugins and how to properly add/remove custom crane-plugin repository
+- A validation against a plugin repo that ensures a plugin-repository (that `crane` refers to) is valid.
+
+
+## Design Details
+
+### Prerequisites
+
+- User have their own plugins or modified version of default provided plugins developed in a similar manner as [this](https://github.com/konveyor/crane-plugin-openshift) plugin.
+- User have a custom plugin repository setup in similar manner as [crane-plugin](https://github.com/konveyor/crane-plugins)
```
sseago, line 81: Link to the raw index file as above.

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
```diff
@@ -0,0 +1,163 @@
...
+
+#### Solution
+
+##### Part 1
+
+A config file `repos.yaml`, placed in `--plugin-dir`. This file contains plugin repository name as key and respective URL as value. Absence of this file in a plugin directory would mean crane will only refer to default plugin repository. However, if an empty `repos.yaml` file is present means user need to add at least one plugin repository such as [crane-plugins](https://github.com/konveyor/crane-plugins). It makes sense to maintain `repos.yaml` for different `--plugin-dir` since, different `--plugin-dir` would have different set of plugins potentially installed from different sources.
```
sseago, line 87: "Absence of this file in a plugin directory would mean crane would create a new repos.yaml in the plugin directory linking only to the crane-plugins repo."

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
```diff
@@ -0,0 +1,163 @@
...
+
+- The file `repos.yaml` would look like following:
+  
+```
+foo: <github raw url for an index.yaml file>
```
sseago, line 92: "github raw or other url/local path to an index.yaml file" -- it doesn't need to be on github, and it won't be on github for local repos.

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
```diff
@@ -0,0 +1,163 @@
...
+
+Global Flags:
+  -p, --plugin-dir string    The path where binary plugins are located (default "plugins")
+```
+
+- Similarly `crane plugin-manager list, add` will also look at the same config file to discover plugins from, to install plugins from, when used with `--repo` flag and otherwise when there is no conflicts of plugin name. 
+  
+**Note:** 
+- All above command will create `repos.yaml` if it does not exists with a single entry that is - `crane-plugins: https://raw.githubusercontent.com/konveyor/crane-plugins/crane-cli/index.yml` a path to our crane-plugins.
+- To update an entry in `repos.yaml`, first the repo needs to be removed(i.e `crane plugin-manager remove-repo foo -p bar`), and then needs to be added with updated path(i.e `crane plugin-manager add-repo foo <path> -p bar`)
```
sseago, line 154: Question: does removing a repo *also* remove the plugins currently installed from that repo?

shawn-hurley (changes requested):
I think we need to consider some scenarios.

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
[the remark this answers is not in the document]
sseago: Yeah, I had a similar comment. I think the motivation is probably equally "new/custom repos" and "'new ways of accessing the default one" -- although in terms of content of the enhancement, I think this addition doesn't really change much.

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
sseago: I think plugin names need to be unique *within a repo* but not across repos. @JaydipGabani is that right?
Repo names must be unique, but they're user-supplied rather than pulled out of the repo url itself.

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
JaydipGabani: That is achievable if desired, but do we want this behavior?

sseago: My concern is if we add a repo, then install a plugin, then remove the repo, then add a new one with the same name, there may be some weirdness if there are plugins from completely different repo sources listed as if they are from the same repo.

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
JaydipGabani: +1

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
JaydipGabani: :+1:

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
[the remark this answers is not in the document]
JaydipGabani: > A validation against a plugin repo that ensures a plugin-repository (that `crane` refers to) is valid. Meaning that added custom plugin repository have valid plugin manifests.

I think this validation should do for now :thinking:

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
JaydipGabani: :+1:

enhancements/crane-2.0/plugin-management/custom-plugin-repo.md
JaydipGabani: Yeah, removing all the plugins alongside the repo makes sense

shawn-hurley (approved):
/lgtm sorry for the delay

jwmatthews:

PR #56, opened 2021-12-20
Add Crane 2.0 Proxy Enhancement

jmontleon:


enhancements/crane-2.0/proxy/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+- [x] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+
+## Summary 
+
+The Crane 2.0 Proxy effort is focused on creating a proxy to run in clusters 
+alongside other Crane 2.0 in cluster features. The proxy will enable access to
+source clusters.
```
djzager, line 32: Could you provide some additional context. For example:

> The proxy will enable access to source clusters.

Like for a subsequent call to `crane export`?

enhancements/crane-2.0/proxy/README.md
```diff
@@ -0,0 +1,65 @@
...
+## Non-goals
+- Though storing user credentials plays a part in this Enhancement we do not
+currently see a requirement to impersonate users.
+- Mutating data: At this time we do not see a requirement to modify or mutate
+data returned by clusters.
+
+## Proposal
+
+- Store a list of proxies in a CR. This can be a shared CR for other components
+to assist looking up credentials.
```
djzager, line 48: CR == Custom Resource? Who then owns this CR (https://github.com/konveyor/mtk-operator or some other controller)?

djzager (commented):
A little out of the loop on what the end-goal is here. Originally thought this was for the CLI, but mentions of CR makes me think we are talking about a new controller. I would really benefit from more context on the problem we are trying to solve.

enhancements/crane-2.0/proxy/README.md
jmontleon: Yes, mtk-operator. I think Shawn can weigh in on whether something exists that we can put this information into, if we should create something shared (which I think was the intent) or standalone.

enhancements/crane-2.0/proxy/README.md
jmontleon: This proxy will be for use when crane is run in cluster, for the UI to access as well as crane eport, etc.

I think one of the big goals is to avert CORS issues early on,

djzager: Is this where `crane` running in a pod in the "destination" cluster would be able to access the "source" cluster without a proxy...but the UI will need to talk to this proxy to get information from the "source" cluster?

jmontleon: CORS is mostly an issue for the Web UI. Crane itself could probably use or not use the proxy. Not sure it provides any benefit. Another case where @shawn-hurley could weigh in.

jmontleon: It might be beneficial for simplifying troubleshooting all traffic to take the same path, so that we don't have a case where the UI is able to set up a connection, query for resources, etcl, all looks good and then crane fails to export because _something_ like a firewall rule doesn't take the different sources into account, but if it's running in the same pod I'm not sure how valid that concern is. Mostly just thinking out loud to see if we can collectively come up with a reason why crane should or shouldn't use the proxy.

enhancements/crane-2.0/proxy/README.md
djzager: Should this enhancement include details about the CRD, the controller watching it, etc?

enhancements/crane-2.0/proxy/README.md
shawn-hurley: My initial gut feeling is that crane export in the pipeline should not use the proxy, I don't think that we want the proxy to be necessary for the tasks and pipelines that we create/install to work. 

With this said an option to use the proxy would be interesting, this would allow us to debug the concern Jason brought up easily. 

This entire thing is because CORS usually creates a degraded experience and we have seen a need for querying certain information from remote clusters to expose to the user to make decisions.

enhancements/crane-2.0/proxy/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Summary 
+
+The Crane 2.0 Proxy effort is focused on creating a proxy to run in clusters 
+alongside other Crane 2.0 in cluster features. The proxy will enable access to
+source clusters.
+
+## Motivation
+
+The primary motivation is to prevent CORS issues. We must also store user
```
shawn-hurley, line 36: I would probably add to the motivation that it is to allow us to show information for users of the UI. We don't expect this to be used with any other piece of crane 2.0 WDYT?

enhancements/crane-2.0/proxy/README.md
shawn-hurley: *  Within the CR we will store a cluster/namespace/secret combination

this is for on startup/crash/restore we can re-find the list of clusters that have been defined by the users to us?

I wonder if a config map with a defined schema might be just as useful and require less stuff. 
We are considering that validation of this combination is done by the UI with calls through the reverse proxy. If we are not managing this, I would think just a list of information might be better suited for something like a config map. As long as we are not expecting this list to grow above 1MB.

shawn-hurley (commented):
Can we potentially add a section on, 

where do we expect this code to be installed or run out of with (I believe it may as well be inside the console plugin I think)?

enhancements/crane-2.0/proxy/README.md
jmontleon: Added

enhancements/crane-2.0/proxy/README.md
jmontleon: What CR or ConfigMap we read this from is the big question for me right now.

I've been working on a [PoC](https://github.com/jmontleon/reverse-proxy-poc/blob/main/main.go). There it's reading from the MigrationController CR Spec as a placeholder, then pulling the URL out of a secret using the namespace/name combo and setting up the connections, proving that proxying to multiple clusters is feasible, and so far not terribly difficult.

jmontleon:
@shawn-hurley I feel like we should also add our intent around SSL. My thoughts are that this proxy either runs in the same container as the UI and we therefore don't need SSL or we rely on a route rather than having to deal with cert nonsense, but open to opinions on why either or both of those are bad ideas.

enhancements/crane-2.0/proxy/README.md
jmontleon: I added clarification that as of now our intent is that this is solely for the UI. If we decide there is a reason to use it for other communication, we can revisit that.

shawn-hurley:
+1 on route to handle SSL for our connections to our service.

I could see a situation where we may need to provide per call certs for certain clusters (maybe internal managed certs or something like a proxy?). do we think we need to add this here?

enhancements/crane-2.0/proxy/README.md
shawn-hurley: If we don't want to manage a CR and have to have statuses and eventing and a controller, I would suggest we look into using a well-defined config map as this entire proxy/plugin thing will have to be a cluster singleton.

jmontleon: I altered the PoC code and enhancement to use a configmap instead, because it makes things easier and we can't see a need for a configmap.

jmontleon:
Added a note regarding intention to use a route and edge termination to provide SSL. I'm not sure if we'd need per call certs, etc. I'm honestly not sure how we could even set that up without creating multiple proxies on different ports, unless you have ideas.

shawn-hurley:
I don't think it would be per call, but could be per cluster right?

jmontleon:
Wouldn't the connection be made before the server even decides which cluster it's connecting to? How would it know which certificate to present at the start of communication, unless you listened on a separate port per cluster or something to that effect?

PR #57, opened 2021-12-23
Crane kustomization support enhancement

djzager:


enhancements/crane-2.0/kustomize-support/README.md
```diff
@@ -0,0 +1,128 @@
...
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] User-facing documentation is created
+
+## Open Questions
+
+- Should the `crane export`ed manifests be the base **OR** should we construct
+    a completely stripped base with the original and transform+apply manifests
+    defined as overlays **OR** (easiest) allow the transform+apply manifests to
+    be the base?
```
djzager, line 39: Originally, I thought that this was a question that needed to be answered as part of this enhancement. After spending some time thinking about it I no longer believe that to be the case.

We absolutely could provide this minimal level of Kustomize support, especially with GA considerations, and later expand that support in future releases.

enhancements/crane-2.0/kustomize-support/README.md
```diff
@@ -0,0 +1,128 @@
...
+configuration management solution embedded in `kubectl` (and `oc`) as
+`apply --kustomization` that allows user to define
+["variants"](https://kubectl.docs.kubernetes.io/references/kustomize/glossary/#variant)
+as
+["overlays"](https://kubectl.docs.kubernetes.io/references/kustomize/glossary/#overlay)
+against a
+["base"](https://kubectl.docs.kubernetes.io/references/kustomize/glossary/#base)
+kustomization without modifying the original (aka "base") manifests.
+It's important that Crane help users improve their situation when migrating
+their applications. This enhancement proposes a new feature to the `apply`
```
shawn-hurley, line 53: Have we considered a new subcommand rather than an option? 

I am wondering if this might be more ergonamic?

enhancements/crane-2.0/kustomize-support/README.md
```diff
@@ -0,0 +1,128 @@
...
+Kustomize as much as possible, a couple examples include:
+
+- [Namespace overrides](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/namespace/)
+- [Patches](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/patches/)
+- [Replacements](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/replacements/)
+- [Change replicas](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/replicas/)
+
+### Goals
+
+- Allow user to further modify the results of `crane apply` with `kustomize`
```
shawn-hurley, line 74: I think we may want to reword the goal, or maybe I need to be convinced on this point. 

My thought was that this was a feature for "Allow users to easily modernize they're exported and transformed manifests to a repeatable deployment tool".  

WDYT? maybe this muddies the water more, but while you can modify results in both places this breaks the problem into two subsets. Changes that need to be made on export and changes that need to be made when repeatedly deploying into clusters?

enhancements/crane-2.0/kustomize-support/README.md
```diff
@@ -0,0 +1,128 @@
...
+    ${CRANE_APPLY_OUTPUT_DIR}`
+
+### Non-Goals
+
+- Support CI/CD onboarding from crane directly. The purpose of providing
+    Kustomize support is to assist the user whichever route they choose to take.
+
+## Proposal
+
+As the resources from `crane export` are iterated over:
```
shawn-hurley, line 86: could this be done on an output of an apply?

shawn-hurley (changes requested):
I think that getting the use case written along with the goal might be helpful to determine what the goal should be or vice versa.

enhancements/crane-2.0/kustomize-support/README.md
sseago: @shawn-hurley I would think so. transform+apply produces a new yaml tree which is in the same format (with same filenames) as the export dir but with changes applied. Really, anything that takes an export dir as an input *could* be working on the results of an apply if the prior apply output dir is given as the export-dir param -- of course this includes transform+apply as well. You can absolutely transform and apply a set of resources that have previously been transformed and applied. Actually, I'd think for the kustomize use case the norm would be working on the output of apply rather than export, if I'm understanding the expected use cases correctly. Maybe we should use something like "input-dir" to make it clear that we're not necessarily talking about raw export results here.

enhancements/crane-2.0/kustomize-support/README.md
djzager: Yeah. Need to spend some more time refining the goal.

shawn-hurley (approved):
/lgtm

PR #59, opened 2022-01-28
docs: tekton and crane ui integration

djzager:
The Tekton + Crane UI Integration enhancement.

### DIFF
--- a/enhancements/crane-2.0/tekton-ui-integration/README.md
+++ b/enhancements/crane-2.0/tekton-ui-integration/README.md
@@ -33,21 +33,23 @@ superseded-by:
 
 1. How will the Crane Workflows, specifically the TaskRuns responsible for
    applying the users manifests onto the destination cluster, impersonate the
-   user making the request? A non-exhaustive list of options:
-   * We could rely on Tekton's default behavior that is to bind Pods
-       (synonymous with TaskRuns) to the `default` serviceAccount.
-   * The UI could bind the appropriate ClusterTasks in the generated Pipeline
-       to the `admin` serviceAccount in the destination namespace. **This
-       approach is assumed in this enhancement**
-   * The UI could generate a Kubeconfig impersonating the current user but
-       it's not clear how the UI would have access to the cluster's
-       coordinates.
+   user making the request? 
+   * **ANSWER** This will be covered in a follow-up enhancement. See [MIG-1076](https://issues.redhat.com/browse/MIG-1076).
+       Until superceded, the UI will create a Secret for the credentials to
+       access the destination cluster and these will be consumed the same way
+       source cluster credentials are consumed.
 1. Getting the operator to override the Crane Runner container image specified
    in ClusterTask Steps. Right now ClusterTasks all reference a
    `quay.io/konveyor/crane-runner:latest` image.
+   * **ANSWER** In the same way operators use environment variables defined on
+       operator deployments in ClusterServiceVersions (or overridden via
+       Subscription Configs), the operator will override the image specified in
+       the deployed ClusterTasks.
 1. Do we want to to execute the `crane-transfer-pvc` ClusterTask for each PVC to
    be migrated or should `crane transfer-pvc` support handling all the PVCs in a
    namespace?
+   * **ANSWER** We are going to run the `crane-transfer-pvc` for each PVC to be
+       migrated.
 1. Is it possible that OpenShift's Pipelines UI will be updated to handle moving
    a PipelineRun out of "PipelineRunPending"?
 
@@ -351,38 +353,59 @@ spec:
         You can get this information in your current environment using
         `kubectl config get-contexts` to describe your one or many
         contexts.
-    - name: src-namespace
+    - name: source-namespace
       type: string
       description: |
         The source cluster namespace in which pvc is synced.
-    - name: pvc-name
+    - name: source-pvc-name
       type: string
       description: |
-        The name of the pvc to be synced.
+        The name of the pvc to be synced from source cluster.
+    - name: destination-context
+      type: string
+      description: |
+        The name of the context from kubeconfig representing the destination
+        cluster.
+
+        You can get this information in your current environment using
+        `kubectl config get-contexts` to describe your one or many
+        contexts.
+    - name: destination-pvc-name
+      type: string
+      description: |
+        The name to give pvc in destination cluster.
+      default: ""
+    - name: destination-namespace
+      type: string
+      description: |
+        The source cluster namespace in which pvc is synced.
+      default: ""
+    - name: pvc-storage-class-name
+      type: string
+      description: |
+        The name of the storage class to use in the destination cluster.
+      default: ""
+    - name: pvc-requests-storage
+      type: string
+      description: |
+        The source cluster namespace in which pvc is synced.
+      default: ""
     - name: endpoint-type
       type: string
       description: |
         The name of the networking endpoint to be used for ingress traffic in the destination cluster
+      default: ""
   steps:
     - name: crane-transfer-pvc
       image: quay.io/konveyor/crane-runner:latest
       script: |
-        # Configure internal cluster
-        kubectl config set-cluster internal \
-          --server=https://kubernetes.default.svc:443 \
-          --certificate-authority=/var/run/secrets/kubernetes.io/serviceaccount/ca.crt
-        kubectl config set-credentials internal \
-          --token="$(</var/run/secrets/kubernetes.io/serviceaccount/token)"
-        kubectl config set-context internal \
-          --cluster=internal \
-          --user=internal \
-          --namespace="$(</var/run/secrets/kubernetes.io/serviceaccount/namespace)"
-
         /crane transfer-pvc \
           --source-context=$(params.source-context) \
-          --destination-context=internal \
-          --pvc-name $(params.pvc-name) \
-          --pvc-namespace $(params.src-namespace) \
+          --destination-context=$(params.destination-context) \
+          --pvc-name $(params.source-pvc-name):$(params.destination-pvc-name) \
+          --pvc-namespace $(params.source-namespace):$(params.destination-namespace) \
+          --pvc-storage-class-name $(params.pvc-storage-class-name) \
+          --pvc-requests-storage $(params.pvc-requests-storage) \
           --endpoint $(params.endpoint-type)
       env:
         - name: KUBECONFIG
@@ -415,15 +438,15 @@ spec:
       description: |
         Namespace to use when scaling down resources
       default: ""
-    - name: type-name-resource
+    - name: resource-type
       type: string
       description: |
-        The resource to be scaled down in type/name format (ie. deployment/mysql or rc/foo)
+        The resource type to be scaled down.
   steps:
     - name: kubectl-scale-down
       image: quay.io/konveyor/crane-runner:latest
       script: |
-        kubectl scale --context "$(params.context)" --namespace "$(params.namespace)" --replicas=0 "$(params.type-name-resource)"
+        kubectl scale --context "$(params.context)" --namespace "$(params.namespace)" --replicas=0 "$(params.resource-type)" --all
       env:
         - name: KUBECONFIG
           value: $(workspaces.kubeconfig.path)/kubeconfig
@@ -446,14 +469,14 @@ metadata:
   name: crane-kustomize-init
 spec:
   params:
-    - name: sourceNamespace
+    - name: source-namespace
       type: string
       description: Source namespace from export.
     - name: labels
       type: string
       description: Add one or more labels
       default: ""
-    - name: namePrefix
+    - name: name-prefix
       type: string
       description: Set the namePrefix field in the kustomization file.
       default: ""
@@ -461,7 +484,7 @@ spec:
       type: string
       description: Sets the value of the namespace field in the kustomization file.
       default: ""
-    - name: nameSuffix
+    - name: name-suffix
       type: string
       description: Set the nameSuffix field in the kustomization file.
       default: ""
@@ -470,10 +493,10 @@ spec:
       image: quay.io/konveyor/crane-runner:latest
       script: |
         # Copy apply resources into kustomize workspace
-        cp -r "$(workspaces.apply.path)/resources/$(params.sourceNamespace)/." "$(workspaces.kustomize.path)"
+        cp -r "$(workspaces.apply.path)/resources/$(params.source-namespace)/." "$(workspaces.kustomize.path)"
 
         pushd "$(workspaces.kustomize.path)"
-        kustomize init --autodetect --labels "$(params.labels)" --nameprefix "$(params.namePrefix)" --namespace "$(params.namespace)" --nameSuffix "$(params.nameSuffix)"
+        kustomize init --autodetect --labels "$(params.labels)" --nameprefix "$(params.name-prefix)" --namespace "$(params.namespace)" --nameSuffix "$(params.name-suffix)"
         kustomize build
         popd
         tree "$(workspaces.kustomize.path)"
@@ -527,15 +550,14 @@ spec:
 ### UI Integration with ClusterTasks
 
 With knowledge of the supported ClusterTasks, the UI will instantiate Pipelines and
-PipelineRuns based on a user's wizard responses. The source cluster's
+PipelineRuns based on a user's wizard responses. The source and destination cluster's
 credentials will be saved as a secret, for example:
 
 ```yaml
 apiVersion: v1
 kind: Secret
 metadata:
-  name: example-cluster-credentials
-  namespace: placeholder
+  generateName: source-cluster-
 data:
   url: ... (base64-encoded)
   token: ... (base64-encoded)
@@ -557,15 +579,16 @@ OpenShift.
 
 ### User Stories
 
-#### Stateless Application Migration
+#### Application Migration
 
 As an application owner, I want to migrate my stateless application from a
-source to destination cluster. 
+source to destination cluster.
 
-For simplicity, we will say the source cluster's API Server URL and Token have
-been stored in a secret named `source-cluster-creds` and the user wishes to
-migrate everything from namespace `guestbook` where no PVCs exist (or where PVCs
-exist but none were selected).
+Through completion of the wizard, the UI will store source and destination
+cluster's API Server URL and Token in secrets named `source-cluster-1234` and
+`destination-cluster-1234` respectively and the `guestbook` namespace was
+selected for migration where no PVCs exist (or where PVCs exist but none were
+chosen).
 
 The generated Pipeline:
 
@@ -576,9 +599,11 @@ metadata:
   name: guestbook-migration
 spec:
   params:
-    - name: cluster-secret
+    - name: source-cluster-secret
       type: string
     - name: source-namespace
+      type: string
+    - name: destination-cluster-secret
       type: string
     - name: optional-flags
       type: string
@@ -586,12 +611,24 @@ spec:
     - name: shared-data
     - name: kubeconfig
   tasks:
-    - name: generate-kubeconfig
+    - name: generate-source-kubeconfig
       params:
         - name: cluster-secret
-          value: "$(params.cluster-secret)"
+          value: "$(params.source-cluster-secret)"
         - name: context-name
-          value: internal
+          value: source
+      taskRef:
+        name: crane-kubeconfig-generator
+        kind: ClusterTask
+      workspaces:
+        - name: kubeconfig
+          workspace: kubeconfig
+    - name: generate-destination-kubeconfig
+      params:
+        - name: cluster-secret
+          value: "$(params.source-cluster-secret)"
+        - name: context-name
+          value: destination
       taskRef:
         name: crane-kubeconfig-generator
         kind: ClusterTask
@@ -600,8 +637,11 @@ spec:
           workspace: kubeconfig
     - name: export
       runAfter:
-        - generate-kubeconfig
+        - generate-source-kubeconfig
+        - generate-destination-kubeconfig
       params:
+        - name: context
+          value: source
         - name: namespace
           value: "$(params.source-namespace)"
       taskRef:
@@ -649,7 +689,7 @@ spec:
       runAfter:
         - apply
       params:
-        - name: sourceNamespace
+        - name: source-namespace
           value: "$(params.source-namespace)"
         - name: namespace
           value: "$(context.taskRun.namespace)"
@@ -665,6 +705,9 @@ spec:
     - name: kubectl-apply-kustomize
       runAfter:
         - kustomize-init
+      params:
+        - name: context
+          value: destination
       taskRef:
         name: kubectl-apply-kustomize
         kind: ClusterTask
@@ -679,14 +722,13 @@ The generated PipelineRun:
 apiVersion: tekton.dev/v1beta1
 kind: PipelineRun
 metadata:
-  generateName: guestbook-migration
+  generateName: guestbook-migration-
 spec:
-  serviceAccountNames:
-    - taskName: kubectl-apply-kustomize
-      serviceAccountName: admin
   params:
-    - name: cluster-secret
-      value: source-cluster-creds
+    - name: source-cluster-secret
+      value: source-cluster-1234
+    - name: destination-cluster-secret
+      value: destination-cluster-1234
     - name: source-namespace
       value: guestbook
     - name: optional-flags
@@ -706,15 +748,29 @@ spec:
     name: guestbook-migration
 ```
 
-#### State Migration
+#### Application and State Migration
 
-As an application owner, I want to migrate the state of my application from a
+As an application owner, I want to migrate my application with state from a
 source to destination cluster.
 
-For this example, we'll assume that the user's cluster API Server URL and Token
-have been stored in a secret named `source-cluster-creds` and the user wishes to
-migrate the state from namespace `guestbook` where two PVCs exist and are
-selected for migration: `redis-data01` and `redis-data02`.
+Through completion of the wizard, the UI will store source and destination
+cluster's API Server URL and Token in secrets named `source-cluster-1234` and
+`destination-cluster-1234` respectively and the `guestbook` namespace where
+two PVCs exist and are selected for migration named `redis-data01` and
+`redis-data02`.
+
+For this use case, the UI will generate two Pipelines and PipelineRuns:
+
+* State Transfer
+* Application and Migration
+
+##### State Transfer Pipeline and PipelineRun
+
+This Pipeline and PipelineRun is responsible for migrating the state of the
+application from the source to destination cluster. It's important to note that
+these don't quiesce the application before transferring state which makes it
+suitable for running to limit how long a State and Application Migration will
+take.
 
 The generated Pipeline:
 
@@ -725,20 +781,33 @@ metadata:
   name: guestbook-state-transfer
 spec:
   params:
-    - name: cluster-secret
+    - name: source-cluster-secret
       type: string
     - name: source-namespace
       type: string
+    - name: destination-cluster-secret
+      type: string
   workspaces:
-    - name: shared-data
     - name: kubeconfig
   tasks:
-    - name: generate-kubeconfig
+    - name: generate-source-kubeconfig
       params:
         - name: cluster-secret
-          value: "$(params.cluster-secret)"
+          value: "$(params.source-cluster-secret)"
         - name: context-name
-          value: internal
+          value: source
+      taskRef:
+        name: crane-kubeconfig-generator
+        kind: ClusterTask
+      workspaces:
+        - name: kubeconfig
+          workspace: kubeconfig
+    - name: generate-destination-kubeconfig
+      params:
+        - name: cluster-secret
+          value: "$(params.source-cluster-secret)"
+        - name: context-name
+          value: destination
       taskRef:
         name: crane-kubeconfig-generator
         kind: ClusterTask
@@ -747,14 +816,17 @@ spec:
           workspace: kubeconfig
     - name: redis-data01
       runAfter:
-        - generate-kubeconfig
+        - generate-source-kubeconfig
+        - generate-destination-kubeconfig
       params:
         - name: source-context
-          value: internal
+          value: source
         - name: source-namespace
           value: "$(params.source-namespace)"
-        - name: pvc-name
+        - name: source-pvc-name
           value: redis-data01
+        - name: destination-context
+          value: destination
         - name: endpoint-type
           value: route
       taskRef:
@@ -765,14 +837,17 @@ spec:
           workspace: kubeconfig
     - name: redis-data02
       runAfter:
-        - generate-kubeconfig
+        - generate-source-kubeconfig
+        - generate-destination-kubeconfig
       params:
         - name: source-context
-          value: internal
+          value: source
         - name: source-namespace
           value: "$(params.source-namespace)"
-        - name: pvc-name
+        - name: source-pvc-name
           value: redis-data02
+        - name: source-context
+          value: destination
         - name: endpoint-type
           value: route
       taskRef:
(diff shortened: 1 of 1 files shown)


djzager:
cc @jmontleon @JaydipGabani

enhancements/crane-2.0/tekton-ui-integration/README.md
```diff
@@ -0,0 +1,1080 @@
...
+- [ ] User-facing documentation is created
+
+## Open Questions
+
+1. How will the Crane Workflows, specifically the TaskRuns responsible for
+   applying the users manifests onto the destination cluster, impersonate the
+   user making the request? A non-exhaustive list of options:
+   * We could rely on Tekton's default behavior that is to bind Pods
+       (synonymous with TaskRuns) to the `default` serviceAccount.
+   * The UI could bind the appropriate ClusterTasks in the generated Pipeline
```
shawn-hurley, line 39: would the UI do this even if the user does not have these permissions?

enhancements/crane-2.0/tekton-ui-integration/README.md
```diff
@@ -0,0 +1,1080 @@
...
+       to the `admin` serviceAccount in the destination namespace. **This
+       approach is assumed in this enhancement**
+   * The UI could generate a Kubeconfig impersonating the current user but
```
shawn-hurley, line 42: Would the UI have access to the user info, and be able to ask the user to do an OAuth flow and get the user's OAuth token? wouldn't we presumably have this?

enhancements/crane-2.0/tekton-ui-integration/README.md
```diff
@@ -0,0 +1,1080 @@
...
+       it's not clear how the UI would have access to the cluster's
+       coordinates.
+1. Getting the operator to override the Crane Runner container image specified
+   in ClusterTask Steps. Right now ClusterTasks all reference a
+   `quay.io/konveyor/crane-runner:latest` image.
+1. Do we want to to execute the `crane-transfer-pvc` ClusterTask for each PVC to
```
shawn-hurley, line 48: each PVC is my vote, and make it async.

enhancements/crane-2.0/tekton-ui-integration/README.md
```diff
@@ -0,0 +1,1080 @@
...
+
+## Proposal
+
+### Crane Runner Container Image
+
+Crane Runner will publish and support a container image including the following
+binaries:
+
+* crane - for executing crane commands
+* oc & kubectl - for interacting with cluster resources and/or managing cluster
```
shawn-hurley, line 130: Doesn't oc and kubectl come with kustomize?

enhancements/crane-2.0/tekton-ui-integration/README.md
```diff
@@ -0,0 +1,1080 @@
...
+```
+
+[Interacting with Pipelines using the developer perspective in OpenShift](https://docs.openshift.com/container-platform/4.9/cicd/pipelines/working-with-pipelines-using-the-developer-perspective.html#op-interacting-with-pipelines-using-the-developer-perspective_working-with-pipelines-using-the-developer-perspective)
+centers around the Pipelines, for this reason the UI will **not** be
+instantiating PipelineRuns with embeded Pipeline specifications. After the user
+completes the wizard, the responses will be used to instantiate Pipeline(s) and
+PipelineRun(s). Examples of the different Pipelines and PipelineRuns are
+presented in the [User Stories](###user-stories) section.
+
+When the user decides to defer execution of the generated
```
shawn-hurley, line 552: there is no UI component to do this, and it is on the status of the pipeline not the spec?

enhancements/crane-2.0/tekton-ui-integration/README.md
djzager: What permissions must a user have to access the Crane UI for a namespace at all?

enhancements/crane-2.0/tekton-ui-integration/README.md
```diff
@@ -0,0 +1,1080 @@
...
+
+For ClusterTasks that require access to the destination cluster, we will use the
+`admin` serviceAccount in the namespace where the PipelineRun is instantiated.
+This should sufficiently limit the scope of the TaskRuns to one namespace.
+
+## Design Details
+
+### Upgrade / Downgrade Strategy
+
+It is important that we treat the ClusterTasks supported by Crane Runner **as
```
shawn-hurley, line 1034: is there some way to version these? a label or annotation?

shawn-hurley (changes requested):
I think that we should consider some approach for versioning on the cluster tasks.

I also believe that we should use the user's oath in some way to have access to the destination cluster rather an trying to rely on an admin user always being present.

enhancements/crane-2.0/tekton-ui-integration/README.md
djzager: being able to run `kustomize init` requires the kustomize binary.

enhancements/crane-2.0/tekton-ui-integration/README.md
djzager: Annotate them with the container image (crane runner image) that held them? This problem might be related to how do we get the operator to override the `image:` field in the ClusterTasks.

enhancements/crane-2.0/tekton-ui-integration/README.md
djzager: https://tekton.dev/docs/pipelines/pipelineruns/#pending-pipelineruns It's definitely on the `.spec` of the PipelineRun.

> there is no UI component to do this, 

@mturley and I were talking about how to better support it in the Pipelines UI. I'll update this to include the 3 methods currently supported to move a pending pipelinerun.

enhancements/crane-2.0/tekton-ui-integration/README.md
mturley: I was asking @spadgett about this, unfortunately the plugin SDK does not let me read the current user's OAuth token, it only gives me functions to call to use it under the hood for k8s resource CRUD. I think we'd have to add a form field to the wizard UI to collect it.

See discussion on Slack: https://coreos.slack.com/archives/C011BL0FEKZ/p1643320736160600

mturley:
Just to copy here: there are a few properties the UI collects (as designed currently) that I don't see how to map to params on the ClusterTasks here for each PVC here: migration type (from prior discussion it sounds like this may not be relevant) target storage class name, capacity, and a boolean for whether to verify the copy.

Slack discussion: https://coreos.slack.com/archives/C02ELCA262K/p1643662825288189

enhancements/crane-2.0/tekton-ui-integration/README.md
```diff
@@ -0,0 +1,1181 @@
...
+        cluster.
+
+        You can get this information in your current environment using
+        `kubectl config get-contexts` to describe your one or many
+        contexts.
+      default: ""
+    - name: namespace
+      type: string
+      description: |
+        The namespace from which to export resources.
```
eriknelson, line 218: RC: "The namespace in the source cluster from which"

enhancements/crane-2.0/tekton-ui-integration/README.md
```diff
@@ -0,0 +1,1181 @@
...
+        valueFrom:
+          secretKeyRef:
+            name: $(params.cluster-secret)
+            key: url
+      - name: CLUSTER_TOKEN
+        valueFrom:
+          secretKeyRef:
+            name: $(params.cluster-secret)
+            key: token 
+  workspaces:
```
eriknelson, line 186: The workspaces on a ClusterTask are really references that point to definitions that are defined in the pipeline, correct?

enhancements/crane-2.0/tekton-ui-integration/README.md
```diff
@@ -0,0 +1,1181 @@
...
+      type: string
+      description: |
+        The name to give pvc in destination cluster.
+      default: ""
+    - name: destination-namespace
+      type: string
+      description: |
+        The source cluster namespace in which pvc is synced.
+      default: ""
+    - name: pvc-storage-class-name
```
eriknelson, line 383: Do we want to name this something like `dest-pvc-storage-class-name` to make the parameter self evident that this is your desired target storageclass?

enhancements/crane-2.0/tekton-ui-integration/README.md
```diff
@@ -0,0 +1,1181 @@
...
+    - name: pvc-requests-storage
+      type: string
+      description: |
+        The source cluster namespace in which pvc is synced.
+      default: ""
+    - name: endpoint-type
+      type: string
+      description: |
+        The name of the networking endpoint to be used for ingress traffic in the destination cluster
+      default: ""
```
eriknelson, line 397: If no endpoint is specified, what is `transfer-pvc`'s default behavior?

enhancements/crane-2.0/tekton-ui-integration/README.md
```diff
@@ -0,0 +1,1181 @@
...
+   in ClusterTask Steps. Right now ClusterTasks all reference a
+   `quay.io/konveyor/crane-runner:latest` image.
+   * **ANSWER** In the same way operators use environment variables defined on
+       operator deployments in ClusterServiceVersions (or overridden via
+       Subscription Configs), the operator will override the image specified in
+       the deployed ClusterTasks.
+1. Do we want to to execute the `crane-transfer-pvc` ClusterTask for each PVC to
+   be migrated or should `crane transfer-pvc` support handling all the PVCs in a
+   namespace?
+   * **ANSWER** We are going to run the `crane-transfer-pvc` for each PVC to be
```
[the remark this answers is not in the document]
eriknelson: This may be answered elsewhere in the document but as I'm reading it occurs to me that another question that has to be answered here is: if we run one clustertask per pvc transfer...is the the UI that is responsible for adding N clustertasks, in serial, to the pipeline for each pvc that has been selected?

enhancements/crane-2.0/tekton-ui-integration/README.md
```diff
@@ -0,0 +1,1181 @@
...
+          value: $(workspaces.kubeconfig.path)/kubeconfig
+  workspaces:
+    - name: kubeconfig
+      description: |
+        The kubeconfig for accessing the source cluster.
+```
+
+#### crane-kubectl-scale-down
+
+The `kubectl-scale-down` ClusterTask is responsible for scaling down the
```
eriknelson, line 421: @pranavgaikwad is this task sufficient to accomplish the abstract idea of "Quiesce"? What does MTC do if they are one-off pods with volumes mounted?

enhancements/crane-2.0/tekton-ui-integration/README.md
```diff
@@ -0,0 +1,1181 @@
...
+* oc & kubectl - for interacting with cluster resources and/or managing cluster
+    configuration.
+* kustomize - for customizing the manifests after `crane apply`.
+
+### Crane Runner ClusterTasks
+
+Crane Runner will support, at a minimum, the ClusterTasks described in this
+section. These are the primary integration between Crane's UI component and Tekton.
+The UI must know about each of these ClusterTasks, the parameters and workspaces
+they require, and how they are to be ordered in Pipelines.
```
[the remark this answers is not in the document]
eriknelson: I've gotten far enough to find this answer:

> centers around the Pipelines, for this reason the UI will **not** be
instantiating PipelineRuns with embeded Pipeline specifications.

I wonder if we briefly tease this and link to later in the doc?

enhancements/crane-2.0/tekton-ui-integration/README.md
```diff
@@ -0,0 +1,1181 @@
...
+
+When the user decides to defer execution of the generated
+PipelineRun it will be given a `.spec.status` of "PipelineRunPending". Later,
+when the user wants to execute the PipelineRun, all the user must do is remove
+the `.spec.status` from the PipelineRun or managed via the Pipelines UI in
+OpenShift. 
+
+### User Stories
+
+#### Application Migration
```
eriknelson, line 582: RC: We should title this "Stateless Application Migration" to call out the stateless use-case

enhancements/crane-2.0/tekton-ui-integration/README.md
```diff
@@ -0,0 +1,1181 @@
...
+          resources:
+            requests:
+              storage: 10Mi
+    - name: kubeconfig
+      emptyDir: {}
+  pipelineRef:
+    name: guestbook-migration
+```
+
+#### Application and State Migration
```
eriknelson, line 751: RC:  I think we want to reframe this for UX purposes as "Stage" and "Cutover" pipelines. It's a pattern that our users are already going to be familiar with that folks are going to be looking for. I think we'll want a direct answer for that.

That probably means just renaming some things here, where the state migration is "Stage" and the "Cutover" migration is the "Final" one.

@pranavgaikwad can you review this and see if there is anything that Stage and Cutover does in MTC that we're missing here that you consider to be critical?

enhancements/crane-2.0/tekton-ui-integration/README.md
```diff
@@ -0,0 +1,1181 @@
...
+generated Kubeconfig (ie. `crane-export` and `crane-transfer-pvc`), and
+others that interact with the
+[mounted secrets to access Kubernetes API from inside the Pod](https://kubernetes.io/docs/tasks/run-application/access-api-from-pod/#directly-accessing-the-rest-api)
+(ie. `crane-transfer-pvc` and `kubectl-apply-kustomize`).
+
+We must be careful not to expose credentials when performing the `oc login`
+using the API server URL and token. The plan is to mitigate this using `set +x`
+to prevent the login line from being included in Pod logs.
+
+For ClusterTasks that require access to the destination cluster, we will use the
```
eriknelson, line 1115: RC: I think this is the problem we need to blackbox and have that discussion about @shawn-hurley @djzager.

I think this is a non-goal of this enhancement, but we should call out that this is a difficult problem that needs its own enhancement with the details around how this is going to work in a safe manner for a multi-tenant environment.

enhancements/crane-2.0/tekton-ui-integration/README.md
```diff
@@ -0,0 +1,1181 @@
...
+
+### Upgrade / Downgrade Strategy
+
+It is important that we treat the ClusterTasks supported by Crane Runner **as
+our API contract**, we are not able to version our ClusterTasks via OCI layering
+and will need to be careful to keep the UI in sync with the ClusterTasks. The
+greatest probably of API incompatibilities are from adding/removing parameters
+or workspaces from ClusterTasks.
+
+To help mitigate these risks, whatever container image specified for Crane
```
eriknelson, line 1129: I think an assertion by the pipeline that these are all fall within acceptable and expected versions and it fails fast and loudly with a specific error would be a good thing to add.

eriknelson (changes requested):
Overall I think this is a really thorough enhancement and it satisfies our need to get our design thoughts captured so we can begin the implementation. There are several open questions we should document and in some cases defer to follow-on enhancements. Let's just make sure those are all called out, and I will make sure that we have spikes on the roadmap to get those questions answered. Thank you for this @djzager, I learned a lot with the read.

enhancements/crane-2.0/tekton-ui-integration/README.md
djzager: Correct. In a Task, workspaces define where the data is located on disk for the individual steps...Tekton's doc say it better than I:

> Tasks specify where a Workspace resides on disk for its Steps. At runtime, a TaskRun provides the specific details of the Volume that is mounted into that Workspace.

https://github.com/tektoncd/pipeline/blob/main/docs/workspaces.md#workspaces-in-tasks-and-taskruns

I think I tried to answer this in the Glossary under Workspaces. Let me know if I need to give it another go.

enhancements/crane-2.0/tekton-ui-integration/README.md
djzager: it'll fallback to the `crane` binaries default `nginx-ingress`

enhancements/crane-2.0/tekton-ui-integration/README.md
```diff
@@ -0,0 +1,1181 @@
...
+
+### Crane Runner Container Image
+
+Crane Runner will publish and support a container image including the following
+binaries:
+
+* crane - for executing crane commands
+* oc & kubectl - for interacting with cluster resources and/or managing cluster
+    configuration.
+* kustomize - for customizing the manifests after `crane apply`.
```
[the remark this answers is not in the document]
djzager: Yeah. We need it for changing the namespace to wherever we are being run. It has some benefits later but we only **need** it for `kustomize init`.

Updating this for more detail.

enhancements/crane-2.0/tekton-ui-integration/README.md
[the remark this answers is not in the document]
djzager: I'd like some input from @alaypatel07 here but I have a few thoughts:

1. The `crane` binary itself should be sufficient for throttling parallel migrations. If you were to run a script with `crane transfer-pvc` for 100 PVCs to be migrated, how should `crane` handle that? Maybe this requires it's own RFE? Once we know how `crane` handles it, we can modify the associated ClusterTask to work inside those confines.
2. `i.e., if they got 4 of the 5 pvcs, they shouldn't have to do everything over again.` it may be worth researching [tekton's retries](https://tekton.dev/vault/pipelines-v0.14.3/pipelines/#using-the-retries-parameter) to see if that's something we should consider. In the scenario that 1 PVC transfer fails, and you re-run the `crane transfer-pvc` wouldn't the amount of work to be done with the 4 previously successful transfers be minimal? Or is the concern here failures outside of `rsync` (ie. routes, tunnels, etc)?
3. The UI is responsible for adding a step to the pipeline **for each** pvc to be transferred.

enhancements/crane-2.0/tekton-ui-integration/README.md
djzager: could you expand on this?

enhancements/crane-2.0/tekton-ui-integration/README.md
mturley: This is a good question. Currently I have it set up so anyone who can see the developer perspective of the console (everyone?) can access it for any namespace they can see. There is a mechanism to limit this in the plugin SDK that I haven't explored yet.

enhancements/crane-2.0/tekton-ui-integration/README.md
```diff
@@ -0,0 +1,1200 @@
...
+we want to hand off the workflow to Tekton and the Pipelines UI to allow the
+user to run the workflow at their discretion. This work will allow a seemless
+experience for Crane UI and command-line users alike.
+
+### Goals
+
+* The UI component outputs a Tekton Pipeline that can be executed at the users'
+    discretion.
+* The Tekton Pipeline can be modified and re-used by the end user.
+* The ClusterTasks provided by Crane Runner should still be consumable via
```
alaypatel07, line 120: sorry a bit rusty on getting back bear with me a little...

Is this for allowing the users to mix and match the ClusterTasks? for example a user wants to have their own quiesce mechanism for their app?

Do we expect the user to provide a custom crane-runner image with their tasks or do we expect the user to create those tasks in the destination kube server themselves? 

I wonder if a little more explanation on this goal would be good for the record books

enhancements/crane-2.0/tekton-ui-integration/README.md
```diff
@@ -0,0 +1,1200 @@
...
+
+* User impersonation. It is desirable for the ClusterTasks that interact with
+    the destination cluster to do so as the user who submitted the request. This
+    is a difficult problem that requires it's own enhancement. See [Open Questions](##open-questions).
+
+## Proposal
+
+### Crane Runner Container Image
+
+Crane Runner will publish and support a container image including the following
```
alaypatel07, line 138: question: when supporting a container image with these binaries, have we thought about how we are going to get the version/commit info back when debugging? I think it was an afterthought with MTC we might want to start thinking about it from the ground up maybe file an RFE if there isn't one?

enhancements/crane-2.0/tekton-ui-integration/README.md
```diff
@@ -0,0 +1,1200 @@
...
+section. These are the primary integration between Crane's UI component and Tekton.
+The UI must know about each of these ClusterTasks, the parameters and workspaces
+they require, and how they are to be ordered in Pipelines. This is to facilitate
+the UI generating Pipelines and PipelineRuns (more details provided in
+[UI Integration with ClusterTasks](###ui-integration-with-clustertasks))
+allowing a seamless transition from Crane's UI to the Pipeline's UI.
+
+#### crane-kubeconfig-generator
+
+The `crane-kubeconfig-generator` ClusterTask is responsible for taking a secret
```
alaypatel07, line 162: Is there a way we can store the token in something like a vault instance?

My hope is to use something tighter to store the tokens than kube secrets. If this is possible I think it will kill two birds, i.e. increase security and solve the problems of passing the destination token to cluster tasks

enhancements/crane-2.0/tekton-ui-integration/README.md
djzager: This is so users can consume our ClusterTasks in their own pipelines. The UI won't be doing anything to support custom ClusterTasks (at least that's not what's being proposed here).

enhancements/crane-2.0/tekton-ui-integration/README.md
djzager: I think we should open a Jira to tackle that. Being able to id the container image that provided all the things (ClusterTasks, `crane` and other binaries) is as far as I want to go in this enhancement. Including build info in the crane-runner image is outside the scope of this enhancement imo.

enhancements/crane-2.0/tekton-ui-integration/README.md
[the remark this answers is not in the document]
alaypatel07: Had a conversation with @djzager on this, we think the biggest question to be answered, that also considers @eriknelson's points is "how granular do we want to get with the control of parallelization with state-transfer?"

If we want to be in charge of all the decisions, I think it would make sense to put this in transfer-pvc sub-command, this will be more work, like it would need the typed config feature to pass the data to transfer-pvc

If we are okay to loose some control we can let tekton run things in parallel

enhancements/crane-2.0/tekton-ui-integration/README.md
```diff
@@ -0,0 +1,1200 @@
...
+      type: string
+      description: |
+        The name to give pvc in destination cluster.
+      default: ""
+    - name: destination-namespace
+      type: string
+      description: |
+        The source cluster namespace in which pvc is synced.
+      default: ""
+    - name: dest-pvc-storage-class-name
```
mturley, line 401: What storage class is used if `dest-pvc-storage-class-name` is left blank? In the UI we make the user choose a destination storage class for each PVC, and I was going to look into how we should determine the default selection there. Maybe there should be some way in the UI for the user to indicate "this is blank, use whatever Crane wants to use".

Also, I imagine this `crane-transfer-pvc` ClusterTask is where we'd need to add params for the capacity of the new PVC and the flag for whether to verify the copy, assuming we still want to collect those in the UI.

enhancements/crane-2.0/tekton-ui-integration/README.md
eriknelson: @alaypatel07 one of the benefits of using Tekton is the ability to easily customize pipelines with your own tasks. The MTRHO UI experience will take an opinionated approach by building the basic pipeline that helps you run crane tasks in a controlled manner. I.E. a "Stage" and "Cutover" pipeline, that are really an abstraction over the sequence of crane tasks necessary to achieve that type of migration. Naturally for the same reason you might want a hook in MTC, you could hook in another task into the generated pipeline anywhere you'd like. Similarly, you could swap out our quiesce task that scales down for something more sophisticated like pushing a db in to read only mode to retain uptime.

We had these same conversations with MTC, and in practice nobody ever got more sophisticated in this manner with the downstream produce. I would be surprised if that happened here, but at the same time, we also had more sophisticated customers that were using MTC as a building block to accomplish complex orchestrations, so it's not out of the question.

If nothing else, these types of flexibility are tremendously valuable for *us* when we're on the hook supporting migrations. We need the escape hatches that allow us to do the things we need to do in a way that early monolithic MTC (CAM) didn't allow for.

enhancements/crane-2.0/tekton-ui-integration/README.md
eriknelson: This is critical for jumpstarting debugging. We have to be able to tie an image in a customer environment back to the upstream commit that has those assets. In the past we just solved this with an image label that has the upstream commit SHA that it was built from. I will file a Spike for this, although I'm not sure it's a problem that's broad enough to warrant its own enhancement. Probably a decision documented on the Jira would do fine, since there is precedent for this. I would just ask the question of whether or not the commit SHA of crane-runner is enough for us, or if we need to include anything else.

enhancements/crane-2.0/tekton-ui-integration/README.md
eriknelson: This is good! Was mostly for my own edification.

enhancements/crane-2.0/tekton-ui-integration/README.md
djzager: Yeah. I was under the impression that downstream pipelines already placed this type of build info as labels on the published container image. My guess for the goal of the Jira would be to support something like:

```
docker pull ${CRANE_RUNNER_IMAGE}
docker inspect --format='{{ index .Config.Labels "vcs-ref" }}' ${CRANE_RUNNER_IMAGE}
```

enhancements/crane-2.0/tekton-ui-integration/README.md
```diff
@@ -0,0 +1,1200 @@
...
+  generateName: source-cluster-
+data:
+  url: ... (base64-encoded)
+  token: ... (base64-encoded)
+type: Opaque
+```
+
+[Interacting with Pipelines using the developer perspective in OpenShift](https://docs.openshift.com/container-platform/4.9/cicd/pipelines/working-with-pipelines-using-the-developer-perspective.html#op-interacting-with-pipelines-using-the-developer-perspective_working-with-pipelines-using-the-developer-perspective)
+centers around the Pipelines -- PipelineRuns that aren't associated with a
+PipelineRun are very difficult to discover or manage -- for this reason the
```
alaypatel07, line 587: s/PipelineRun/Pipeline?

enhancements/crane-2.0/tekton-ui-integration/README.md
alaypatel07: yeah, this one is a little weirdly named, maybe it should have been the "spec.pause" field IMO to stay consistent with other upstream APIs

shawn-hurley (approved):
looks good

enhancements/crane-2.0/tekton-ui-integration/README.md
```diff
@@ -0,0 +1,1181 @@
...
+       operator deployments in ClusterServiceVersions (or overridden via
+       Subscription Configs), the operator will override the image specified in
+       the deployed ClusterTasks.
+1. Do we want to to execute the `crane-transfer-pvc` ClusterTask for each PVC to
+   be migrated or should `crane transfer-pvc` support handling all the PVCs in a
+   namespace?
+   * **ANSWER** We are going to run the `crane-transfer-pvc` for each PVC to be
+       migrated.
+1. Is it possible that OpenShift's Pipelines UI will be updated to handle moving
+   a PipelineRun out of "PipelineRunPending"?
```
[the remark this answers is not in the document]
djzager: https://github.com/openshift/console/issues/11060

enhancements/crane-2.0/tekton-ui-integration/README.md
djzager: Adding this to the open questions with link to relevant Jira once it exists :sunglasses:

alaypatel07 (commented):
Thanks for the thorough write up, overall it makes a lot of sense.

/lgtm

enhancements/crane-2.0/tekton-ui-integration/README.md
djzager: Storage class name is optional so it shouldn't be required by our ClusterTask or `crane`. I'm uncertain if the UI should require this -- I am not fully aware of the behavior when left blank -- and would defer to @shawn-hurley or @eriknelson. Maybe leaving it as `""` should be acceptable?

After talking with @alaypatel07 I'm not certain what priority we should give to "verify the copy". I feel like we at least need namespace, size, storageClassName and I'm iffy on accessModes, changing the PVC name, and this verify copy.

enhancements/crane-2.0/tekton-ui-integration/README.md
djzager: https://github.com/konveyor/crane-runner/issues/33 since I couldn't figure out how to word a Jira.

enhancements/crane-2.0/tekton-ui-integration/README.md
```diff
@@ -0,0 +1,1200 @@
...
+      default: ""
+    - name: dest-pvc-storage-class-name
+      type: string
+      description: |
+        The name of the storage class to use in the destination cluster.
+      default: ""
+    - name: pvc-requests-storage
+      type: string
+      description: |
+        The source cluster namespace in which pvc is synced.
```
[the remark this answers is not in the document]
djzager: Unintentional.

It was a mistake to not include these parameters in the pipeline/pipelineRun somewhere. Need to address this piece before this can be called done.

enhancements/crane-2.0/tekton-ui-integration/README.md
djzager: updated https://issues.redhat.com/browse/MIG-1074 as this will really need to be handled there (with updates to transfer-pvc and crane-runner's ClusterTasks).

enhancements/crane-2.0/tekton-ui-integration/README.md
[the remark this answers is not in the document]
djzager: We have a whole spike dedicated to researching this question https://issues.redhat.com/browse/MIG-1082 , we also have the bug in crane https://github.com/konveyor/crane/issues/66 .

All of this is documented in the enhancement. I consider this resolved.

enhancements/crane-2.0/tekton-ui-integration/README.md
eriknelson: > I am not fully aware of the behavior when left blank -- and would defer to @shawn-hurley or @eriknelson. Maybe leaving it as "" should be acceptable?

I need to brush off the recent conversation we had @mturley in MTC about its behavior, I think we want to replicate that here.

IMO, if the key is present but explicitly set to "", that feels like a deliberate setting of "None", i.e. no storageclass. I'm trying to recall if that is even something that's meaningful, or what the use-case is for that. If the key is missing, then that feels like it should implicitly fall back to the default storageclass. @djzager I'm unsure if the distinction is even possible with the Tekton params API, or if it will all appear to be "" when read.

"verify copy" is very important. I'm not sure it's MVP, but the reason is that in a regulated industry, this is sometimes an audit requirement. Meaning if we *dont* have this, it will exclude entire classes of users from this tool even being an option to them for usage, and we do have several customers who need this.

It should be relatively simple to pass through to the transfer-pvc command as a switch; we know crane-lib supports it because MTC is using it. I think we should strive to have this for GA.

In early versions of MTC we did not have PVC name mapping and it was heavily requested. I'm struggling to come up with the use-case for needing to have the ability to adjust the accessMode. @pranavgaikwad @sseago do you see either of these being a high priority so that they should be on the critical path for MTRHO GA?

shawn-hurley: Adjusting the access mode in this context, I think, is something that we should be worried about doing. 

There is currently no way in RBAC to deny this, *but* it is considered a security risk attempting to be fixed.

Even if there is a use case, I will vote that in a namespace admin copy, we don't allow this.

djzager: > IMO, if the key is present but explicitly set to "", that feels like a deliberate setting of "None", i.e. no storageclass. I'm trying to recall if that is even something that's meaningful, or what the use-case is for that. If the key is missing, then that feels like it should implicitly fall back to the default storageclass. @djzager I'm unsure if the distinction is even possible with the Tekton params API, or if it will all appear to be "" when read.

@eriknelson from a Tekton perspective, it is possible to require it be specified (ie. no default) and that would allow for "" to be explicit.

> Adjusting the access mode in this context, I think, is something that we should be worried about doing.

eek. yeah. After chatting with @alaypatel07 and getting more info, I think https://issues.redhat.com/browse/MIG-1074 is the best place to make final decisions on which flags we are going to add for MVP/GA and address it's impact on crane-runner/crane UI. I linked to this thread from the issue to provide appropriate context to whoever goes to implement the new flags. Let me know if we need any more info.

eriknelson (approved):

PR #60, opened 2022-03-11
Custom targets and sources in advanced options

rromannissen:
Enhanced advanced options screen to accommodate sources and targets from custom rules. Updated some mockups.


gildub (approved):
LGTM

PR #61, opened 2022-03-14
docs: introduce crane ui + tekton auth enhancement

djzager:


enhancements/crane-2.0/tekton-ui-integration-auth/README.md
```diff
@@ -0,0 +1,228 @@
...
+1. Ask the user for their token and store it in a secret.
+1. Send the user to the token exchange to retrieve a token and store that in a
+   secret.
+1. Introduce new CRD surface + controller + and WebHook.
+1. Force the Pipelines to run as "admin" by binding a Service Account to the
+   "admin" RoleBinding.
+1. Map the user's permissions (via a `SelfSubjectRulesReview`) onto a Role and
+   bind that to a generated Service Account.
+1. Allow the user to specify a Service Account to use in the migration.
+
```
alaypatel07, line 136: Been discussing with @djzager about allowing migration for a separate namespace, we rejected that because of the below mentioned reason

```suggestion
1. Create a separate namespace where the user can run the migration, call it
   migration namespace for clarity. The idea is to create resources and secrets 
   in the migration namespace so it is secure from other users in the user namespace. 
   However, this does not solve the original problem because the template for
   creating the namespace is specified by the cluster-admin at the cluster level,
   more details [here](https://docs.openshift.com/container-platform/4.10/applications/projects/configuring-project-creation.html#modifying-template-for-new-projects_configuring-project-creation). 
   Because it is a cluster-wide configuration, the same user groups that have read
   access in the namespace being migrated will continue to have the read access in
   the migration namespace. Hence this is not any better than migrating in the existing 
   namespace.
```

jmontleon:
https://github.com/konveyor/enhancements/pull/61/files#diff-cd9526599c6ee566a2efc8ad0754db69269852e84ea5616576761e00bcd1be66R193

mturley:
@jmontleon that's a concern I had a while back as well, thanks for bringing that back up. That's another use case that would be well served by us having our own place in the UI to present actions for these pipelines/runs. so we could have some "refresh credentials" button somehow. @vconzola

djzager:
@mturley @jmontleon the secret-service does support updating the token (via PUT/PATCH). I'm not certain how that would be surfaced in the UI.

jwmatthews:

PR #62, opened 2022-03-15
Application Inventory / Pathfinder / Container Advisor integration

rofrano:
This proposal is to integrate Tackle Container Advisor (TCA) with both the Tackle Application Inventory for the purpose of automatically and consistently tagging applications with their technologies, and with Tackle Pathfinder for including containerization advice in the Pathfinder report for the purpose of aiding in determining the 6R disposition of an application.


rofrano:
@jortel This is the initial draft of the Application Inventory / Pathfinder / TCA integration. I've updated it based on a conversation with @rromannissen but we need your input on the implementation details. Thanks.

PR #63, opened 2022-03-28
Crane state transfer: file ownership change

pranavgaikwad:


enhancements/crane-2.0/permission-change/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+- Propose a ClusterTask to change file ownership after state transfer
+
+### Non-Goals
+
+- Change file ownership during state transfer
+
+## Proposal
+
+A ClusterTask will run an Ansible Playbook to change ownership of files. The job of the ClusterTask is to simply run the playbook and pass variables to it. The actual ownership change will be handled by the Ansible Playbook.
```
shawn-hurley, line 55: Why an ansible playbook and not like a POSIX script?

pranavgaikwad: @shawn-hurley Ansible k8s makes it much easier to discover PVCs, read annotations, template the Pod definition (Pod definition will be dynamic based on PVCs in the namespace)

pranavgaikwad: @shawn-hurley It could be. Fixing permissions will work at a namespace level (it would be inefficient to launch pod per PVC). On the other hand, transfer-pvc works at PVC level. I felt there was a disconnect there. Apart from that, I think it definitely can be a subcommand.

shawn-hurley: I think I am asking why we are bringing in a whole new technology that someone would have to have to run the full workflow from a CLI.

pranavgaikwad:
> Have we explored the possibility of running rsyncd with a different uid/gid?
> 
> The idea is to not run rsyncd as root user but normal user in the namespace and assume rsyncd read and write access to PVC.
> 
> This would get rid of the original problem of workloads breaking when uid/gid has changed.
> 
> Attaching reference where I was reading about uid/gid in rsyncd conf file for quick access
> 
> ```
> uid
> This parameter specifies the user name or user ID that file transfers to and from that module should take place as when the daemon was run as root. In combination with the lqgidrq parameter this determines what file permissions are available. The default is uid -2, which is normally the user lqnobodyrq.
> gid
> This parameter specifies the group name or group ID that file transfers to and from that module should take place as when the daemon was run as root. This complements the lquidrq parameter. The default is gid -2, which is normally the group lqnobodyrq.
> ```
> 
> https://linux.die.net/man/5/rsyncd.conf

@alaypatel07 Using the `uid` `gid` options in Rsyncd has an issue. The owner/group of the root directory of the volume mount are derived from the UID/GID accessible to the Container process. The containers UID/GID depend on SecurityContext of the Pod and/or SCC. Therefore, in order for Rsync daemon uid/gid thing to work, the volume mount must be accessible to that uid/gid. Which implies that the Container process should start with that UID and the UID should belong to that GID. This is only possible when we create a custom SCC that allows that. Without the custom SCC, it won't work in OpenShift. For Kubernetes, it may be possible simply by updating the Container definition. I will do some research on that.

enhancements/crane-2.0/permission-change/README.md
```diff
@@ -0,0 +1,202 @@
...
+Whenever these options are set, the `--chown` option will be added to `rsync` command in the client Pod:
+
+```sh
+rsync --chown=10000680000:10000680000 <...>
+```
+
+Rsync chown option is similar to running `chown` after the transfer. No matter what the owners and groups are on the source side, files on the destination will be owned by owner and group specified in the `--chown=<uid>:<gid>` option.
+
+The CLI users are expected to know the UID & GID values beforehand. The UI can automatically discover the right values based on the annotations in the destination namespace and pass them in the ClusterTask.
+
```
djzager, line 101: I like that we have a way to handle it in the UI seamlessly.

enhancements/crane-2.0/permission-change/README.md
```diff
@@ -0,0 +1,202 @@
...
+- Discuss pros and cons of proposed solutions
+### Non-Goals
+
+- Change permissions of files 
+
+## Open Questions
+
+- What if we implement both the solutions discussed in this doc?
+## Proposal
+### Solution 1: Change ownership of files during state transfer
```
djzager, line 73: My preference is for this solution personally. Via the command-line or clustertasks the user should be able to skip this part. In the UI case, I think we should be opinionated about this and just do it for them.

djzager (commented):
I prefer solution 1 knowing that we _could_ support both. My only concern is with non-admin state migrations, if it's possible, what architecture changes may need to be made to support it, and how those changes might affect this work.

With respect to what is in this PR, my vote is for solution 1 and we can pursue solution 2 if there is a need/requirement/want.

enhancements/crane-2.0/permission-change/README.md
jmontleon: I like Solution 1 as well. I think this should be strongly preferred.

The only reason I would advocate for maintaining Solution 2 is the case where someone performs a migration and only then realizes they  need to fix permissions. Being able to do so after the fact would be very helpful.

pranavgaikwad:
> I prefer solution 1 knowing that we _could_ support both. My only concern is with non-admin state migrations, if it's possible, what architecture changes may need to be made to support it, and how those changes might affect this work.
> 
> With respect to what is in this PR, my vote is for solution 1 and we can pursue solution 2 if there is a need/requirement/want.

For non-admin, on OpenShift, it means being able to run Rsync daemon with `restricted` UID. @alaypatel07 and I discussed this and I think there is a way to achieve that.

shawn-hurley:
+1 on solution 1 as long as it does not require any extra SCC

enhancements/crane-2.0/ownership-change/README.md
```diff
@@ -0,0 +1,223 @@
...
+### Non-Goals
+
+- Change permissions of files
+- Propose a solution for admin users only
+## Proposal
+
+Currently, the Rsync Daemon & Rsync Client both run as `root` user. It is achieved by adding a custom SCC `rsync-anyuid` which allows running the Rsync Pod as any UID. In the proposed solution, the Rsync Pods will run as less privileged user by default. This ensures that non-admin users can run Rsync Pods without issues. Optionally, admin users can choose to run Rsync as `root` user. 
+### Default Operation
+
+By default, the Rsync Pods will run with minimum capabilties. They will drop `SETGID`, `SETUID`, `KILL` and `MKNOD`:
```
djzager, line 75: For me, it would really help me if this were more explicit. Are we creating a custom SCC, if we are, who is creating it and when? **OR** maybe this is handled in the Pod definition itself. It's somewhat ambiguous now how all these pieces come together to work for non-admin users.

enhancements/crane-2.0/ownership-change/README.md
```diff
@@ -0,0 +1,223 @@
...
+  drop:
+    - KILL
+    - MKNOD
+    - SETGID
+    - SETUID
+```
+
+On OpenShift clusters, `restricted` SCC allows Pods that drop above mentioned capabilities. Therefore, Rsync Pods will use `restricted` SCC on OpenShift clusters. On Kubernetes clusters, an equivalent of `restricted` PodSecurityPolicy or PodSecurityAdmission will determine the admission of the Rsync Pods. It is hard to tell whether all Kubernetes clusters will have an equivalent of `restricted` PodSecurityPolicy or PodSecurityAdmission. However, we will ensure that the capabilities of Rsync Pods are kept to the minimum such that they do not require a privileged user.
+
+Rsync Daemon Pod will not run as `root` user by default. Therefore, `uid`, `gid` fields from Rsync Daemon Pod's configuration will be removed:
```
djzager, line 88: These paragraphs make me think that it all has to do with the Pod definition and making it "admissible" on OpenShift/Kubernetes clusters, is that right?

Explaining what we will do more directly would help me better understand.

djzager: The portion about Kubernetes, as I understand it the less we ask for in the Pod, the more likely it is to be accepted in clusters with restricted PodSecurityPolicy/PodSecurityAdmission...is that right? That sounds reasonable to me. Is this something we would be able to detect in `crane` and error out appropriately?

djzager (changes requested):
Looking great. Some clarity on our approach and I think it's ready to go.

enhancements/crane-2.0/ownership-change/README.md
pranavgaikwad: @djzager We are not creating a custom SCC. We are simply updating the SecurityContext of Rsync containers and dropping capabilities that require privileges. On OpenShift we expect that the Pods will be admitted via `restricted` SCC which exists by default. On Kubernetes, we expect an equivalent of `restricted` SCC as a PodSecurityPolicy which will admit the Rsync Pods.

enhancements/crane-2.0/ownership-change/README.md
pranavgaikwad: @djzager It may be possible to detect that based on error strings. At the time of implementation, I will see if there's a more sophisticated way to do it.

enhancements/crane-2.0/ownership-change/README.md
```diff
@@ -0,0 +1,227 @@
...
+  drop:
+    - KILL
+    - MKNOD
+    - SETGID
+    - SETUID
+```
+
+By dropping these capabilities, we ensure that
+- the Pods don't require special privileges
+- we do not need a separate SCC. pods will be admitted via existing admission policies of the cluster (SCC/PodSecurityPolicy/PodSecurityAdmission)
```
pranavgaikwad, line 88: @djzager hope this helps answers your questions

djzager (approved):

enhancements/crane-2.0/ownership-change/README.md
alaypatel07: I think you will get Forbidden http error and then run error string detection.  If no SCC/PSP/PSA is found that admits the pod, an example error from upstream documentation

> Error from server (Forbidden): error when creating "STDIN": pods "privileged" is forbidden: unable to validate against any pod security policy: [spec.containers[0].securityContext.privileged: Invalid value: true: Privileged containers are not allowed]

https://kubernetes.io/docs/concepts/security/pod-security-policy/

Regardless I think the more important thing to call out is, the Pod API will be our primary way knob of determining the least privileged admission mechanism. 

In order to make this solution robust, we need to find a way to make the SecurityContext configurable for advanced users who need to run with more privileges and they have to coordinate with admin to craft the right SCC at cluster level. In other words, advanced use cases will be solved by documentation and CLI overrides but the generic case will be the most restrictive SCC.

Also, it is my understanding, the list of dropCapabilities is for the `restricted` SCC it might be different in different Cluster environments, but it is reasonable to start with the mentioned default list and change it in the future, assuming it is trivial to do so

enhancements/crane-2.0/ownership-change/README.md
```diff
@@ -0,0 +1,227 @@
...
+
+#### Story 2 [Kubernetes]
+
+As a Kubernetes user, I need a way to fix ownership of files migrated with state transfer such that I can migrate existing volumes into a new storage that does not automatically update file ownership using `chown`.
+
+### Implementation Details/Notes/Constraints [optional]
+
+#### Exceptions to SCC
+
+We are only relying on SecurityContext fields to ensure right SCC / PodSecurityPolicies are attached. It is possible that the environments may have multiple conflicting SCC / PodSecurityPolicies and the Rsync Pods may end up using something other than `restricted`. It is likely that the SCC/PodSecurityPolicies admitting the Rsync Pods doesnt have the same `fsGroup`/`runAsUser` policies as `restricted`. In such cases, we cannot guarantee that the Rsync Pods will be admitted at all. The solution to such problems cannot be generalized, and it will have to be solved on a case-by-case basis.
```
alaypatel07, line 150: +1 on calling it out. 

I wonder, if such SCC exists at the cluster level, other services will run into it as well and will be a more generic problem than transfer-pvc failing, but unsure until we get a concrete example.

enhancements/crane-2.0/ownership-change/README.md
```diff
@@ -0,0 +1,227 @@
...
+
+Following Pros/Cons were considered before ruling out the alternative:
+#### Complexity
+
+Adding a new subcommand takes more efforts than simply adding a new option to `rsync` subcommand. Also, creating a new temporary Pod implies added surface for debugging that Pod. Therefore, final solution is better than the alternative in terms of complexity.
+
+#### Speed/Performance
+
+By fixing ownership during Rsync transfers ensures that the total time it takes to fix ownership is overlapped with rsync transfer time. In case of Staging/Cutover setting, this time is also spread across multiple runs making it even faster during the final Cutover. This is similar to the [performance issue](https://github.com/kubernetes/kubernetes/issues/69699) in `kubelet` due to automatic `chown` of volumes at mount time.
```
alaypatel07, line 223: How does this change with the proposed workflow?

enhancements/crane-2.0/ownership-change/README.md
```diff
@@ -0,0 +1,227 @@
...
+
+On Kubernetes clusters, kubelet updates file ownership automatically whenever `fsGroup` field is specified in the PodSecurityContext. It makes sure that the GID specified in `fsGroup` is added to Supplemental Groups of the container process. Also, it runs `chown` recursively on the volume mount before exposing it in the Pod. However, this is not true for _all_ storage types. Automatic `chown` does not work in following cases:
+- [Kubernetes >= 1.19] whenever a CSI driver opts out recursive modifications by setting `fsGroupPolicy` to `None`
+- [Kubernetes < 1.19] whenever a CSI volume does not specify `fsType` OR when `fsType` is present but volume is not ReadWriteOnce
+- for shared storages such as NFS or in-tree storages that do not explicitely change the permissions
```
alaypatel07, line 35: One this to explicitely call out:

Assume that users want to migrate to shared storage where uid should be as selected in the range, but GID should be 5555 for example, the user could create a custom SCC to allow for this and override uid/gid. This will create the files with right uid/gid but require admin privilege which is a acceptable solution since the storage is shared IMHO

alaypatel07 (approved):
/lgtm

This covers all the scenarios I had in mind and allows for advanced user options through CLI overrides and documentation.

enhancements/crane-2.0/ownership-change/README.md
pranavgaikwad: The point I am trying to make here is that we do not need to run chown explicitely after transfer. The total time overlaps with Rsync transfer time with proposed workflow.

PR #64, opened 2022-04-04
Redesigning plugin release

JaydipGabani:


djzager:
cc @shawn-hurley @alaypatel07

enhancements/crane-2.0/plugin-release/README.md
```diff
@@ -45,19 +47,28 @@ We are trying to define a standardized release process for plugins including way
...
 
+The `crane-plugins` repository will only include a list of plugins available within `crane` and respective location of that repository, instead of storing release information and manifests of every released plugins within `crane-plugins`.
+
+`crane-plugins` will act as an index that points to all plugin github repository, so there will be only one file `index.yaml` in `crane-plugins` and that will look something like - 
+
+```
+OpenShiftPlugin: <url-path to crane-plugin-openshift/manifests>
+ImageStreamPlugin: <url-path to crane-plugin-imagestream/manifests>
+```
+
```
djzager, line 71: I do not believe that it is a requirement for the actual plugin manifests to be managed in the `crane-plugin-openshift` repository. Crane-plugins may be the perfect place to store these things, but what is different now, is that the `OpenShiftPlugin` wherever it is pointing is itself an index of all versions discoverable to `crane`.

 What is important now versus then, is that we want the individual plugins to be an index themselves.

JaydipGabani: One of the core motivations behind this change is to decouple each plugin with central repo and the ability to allow plugin authors to release without making changes in `crane-plugins`. 

I believe having just an index file in the respective plugin makes the whole process more complex than it needs to be. i.e
`crane-plugins -> <plugin> -> crane-plugins`.

djzager: What I had in mind was that you have

`${CRANE_PLUGINS_REPO_ROOT}/index.yaml` that looks like:

```yaml
# Should we be considering an APIVersion for these? ie.
apiVersion: konveyor.io/v1alpha1
kind: PluginIndex

plugins:
# Example staying crane-plugins
- name: OpenShiftPlugin
  path: https://github.com/konveyor/crane-plugins/raw/main/plugins/OpenShift/index.yaml
# Example letting the plugin owner manage it
- name: ImageStreamPlugin
  path: https://github.com/konveyor/crane-plugin-imagestream/raw/main/manifests/index.yaml
```

And each of these index YAML looks like:

```
# Should we be considering an APIVersion for these? ie.
apiVersion: konveyor.io/v1alpha1
kind: Plugin

versions:
- name: OpenShiftPlugin
  shortDescription: OpenshiftPlugin
  description: this is OpenshiftPlugin
  version: v0.0.1
  binaries:
    - os: linux
      arch: amd64
      uri: https://github.com/konveyor/crane-plugin-openshift/releases/download/v0.0.1/amd64-linux-openshiftplugin-v0.0.1
    - ...
optionalFields:
  - flagName: "strip-default-pull-secrets"
    help:     "Whether to strip Pod and BuildConfig default pull secrets (beginning with builder/default/deployer-dockercfg-) that aren't replaced by the map param pull-secret-replacement"
    example:  "true"
  - flagName: "pull-secret-replacement"
    help:     "Map of pull secrets to replace in Pods and BuildConfigs while transforming in format secret1=destsecret1,secret2=destsecret2[...]"
    example:  "default-dockercfg-h4n7g=default-dockercfg-12345,builder-dockercfg-abcde=builder-dockercfg-12345"
  - flagName: "registry-replacement"
    help:     "Map of image registry paths to swap on transform, in the format original-registry1=target-registry1,original-registry2=target-registry2..."
    example:  "docker-registry.default.svc:5000=image-registry.openshift-image-registry.svc:5000,docker.io/foo=quay.io/bar"
- name: OpenShiftPlugin
  shortDescription: OpenshiftPlugin
  description: this is OpenshiftPlugin
  version: v0.0.2
  binaries:
    - os: linux
      arch: amd64
      uri: https://github.com/konveyor/crane-plugin-openshift/releases/download/v0.0.1/amd64-linux-openshiftplugin-v0.0.2
    - ...
optionalFields:
  - flagName: "strip-default-pull-secrets"
    help:     "Whether to strip Pod and BuildConfig default pull secrets (beginning with builder/default/deployer-dockercfg-) that aren't replaced by the map param pull-secret-replacement"
    example:  "true"
  - flagName: "pull-secret-replacement"
    help:     "Map of pull secrets to replace in Pods and BuildConfigs while transforming in format secret1=destsecret1,secret2=destsecret2[...]"
    example:  "default-dockercfg-h4n7g=default-dockercfg-12345,builder-dockercfg-abcde=builder-dockercfg-12345"
  - flagName: "registry-replacement"
    help:     "Map of image registry paths to swap on transform, in the format original-registry1=target-registry1,original-registry2=target-registry2..."
    example:  "docker-registry.default.svc:5000=image-registry.openshift-image-registry.svc:5000,docker.io/foo=quay.io/bar"
```

sseago: So the location of the plugin-specific sub-repo is just a path. It could be a link into the crane-plugins github repo. It could be a link into the plugin's own github repo. It could even be a local file path (although this would only be the case if the plugin repo itself was a local file, perhaps something downloaded/installed in a container image, etc.). As I understood things from earlier discussions, the point is to decouple the plugin's release management from the plugin repo's main file. The plugin-specific sub-repo could live in either place, and the software won't care one way or another. There's a maintenance/policy question as to where it *should* live, but that's a separate conversation and can be had on a per-plugin basis, as well as an overall "what makes the most sense?" basis.  Perhaps the enhancement could list a preference, but it certainly won't implement a requirement. In particular, third-party plugin repos will also be using this design for their own plugin implementation, and we aren't telling them where they're supposed to put their plugin-specific sub-repos.

enhancements/crane-2.0/plugin-release/README.md
```diff
@@ -21,7 +23,7 @@ superseded-by:
 
 # Release process for plugins and keeping Crane-plugins upto date with all releases
 
-This document proposes a solution to setup a release cycle for all the `Crane` plugins managed by us. Alongside the approaches to keep `crane-plugins` (all the manifests files for the plugins) upto date.
+This document proposes a redesign of release process and changes in structure of `crane-plugins` and all the `Crane` plugins managed by us a little in order to achieve this simplicity.
```
sseago, line 26: Since this is an update to the design of the plugin repo file format, specific repos such as `crane-plugins` might serve as examples of repos that will make use fo this format, but the existence of any specific plugin repo is not required here. A crane CLI user might *not* use crane-plugins at all and might use their own local plugin repository instead.

sseago: Actually, looking at this again, it's less clear whether this enhancement's primary purpose is the release process change or the plugin repository design change. If we're assuming that the repository design change is already done (or handled separately), then we don't need to worry about the repo format change here -- we can assume that the links to the plugin-specific sub-repo files will already work. But there are two different tasks here. FIrst, enhance the way the plugin-manager works to allow it to pull detailed release information on a plugin from a linked sub-repository, and second, update the release process to make use of this new feature.

shawn-hurley: I think that the purpose of this document should be to describe the full process that is being updated. This should be focused on IMO 
1. The overall UX and goals that we are trying to achieve.
2. The reasons for the change.

enhancements/crane-2.0/plugin-release/README.md
```diff
@@ -36,7 +38,7 @@ Not needed
 
 ## Summary
 
-We are trying to define a standardized release process for plugins including ways to keep the `crane-plugins` repository upto date with all the plugins releases.
+We are trying to make release process simpler and decouple `crane-plugins` from individual plugin repository to make sure independent release of each plugin is as simple as it can be. The changes/approach proposed in this document will make it possible to release a plugin without making changes in `crane-plugin` repository so we can release individual plugins with minimum changes possible.
```
shawn-hurley, line 41: Is the goal to decouple the index from the individual plugin repo? I think that we should state something about a pointer or an abstraction?

enhancements/crane-2.0/plugin-release/README.md
```diff
@@ -45,19 +47,28 @@ We are trying to define a standardized release process for plugins including way
 
 ## Motivation
 
-With the release of crane-lib and crane, there is possible need to release compatible plugins. To make the `crane plugin-manager` work properly, we need to have all the `manifest` files in the default `crane-plugin` repository  upto date. Since there are multiple repositories included (crane-plugins, repositories for each plugin) the release process becomes complicated. Goal of this document is to define a streamlined release process for plugins and a way to make sure that `crane-plugins` repository remains upto date with released plugins.  
+As of now to release a plugin we need to have at least two PRs and need to make changes in two different repository (`crane-plugins` and individual plugin repository). This document aims to reduce the number of steps needed to release a plugin. As of how the current design is implemented, every author needs to be able to contribute to `crane-plugins`, by decoupling `crane-plugins` and individual plugin repos we can also eliminate the need of having access for each plugin author to be able to contribute to `crane-plugins`.
```
shawn-hurley, line 50: Considering that we are updating the original document there will be no history of what was before. 

I think there are two options strike through the original stuff throughout the document so that we have a visible history alternatively make a new doc that describes the full process. Alternatively updating this section to not mention the old design might be sufficient as we have not reached 1.0

enhancements/crane-2.0/plugin-release/README.md
```diff
@@ -45,19 +47,28 @@ We are trying to define a standardized release process for plugins including way
...
 
 ### Goals
 
-- We have a standardized process to release plugins
```
shawn-hurley, line 52: Is this really no longer a goal of the release process?

enhancements/crane-2.0/plugin-release/README.md
shawn-hurley: I am a +1 on what Scott and David said. We want to handle both cases (and very importantly the local file path).

enhancements/crane-2.0/plugin-release/README.md
```diff
@@ -106,26 +117,18 @@ jobs:
...
           tag: ${{ github.event.inputs.version }} 
 
-      - name: Checkout crane-plugins
-        uses: actions/checkout@v2
-        with:
-          repository: ${{ github.repository_owner }}/crane-plugins
-          token: ${{ secrets.PLUGIN_RELEASE }}
-
-      - name: Updating index file and adding manifest
+      - name: Adding manifest
```
shawn-hurley, line 127: This seems to update the manifest file, but I think that it should also update the index file for the plugin (at least make it optional)

enhancements/crane-2.0/plugin-release/README.md
```diff
@@ -152,23 +155,24 @@ jobs:
...
 
 ```
 
-The above workflow raises a PR against [crane-plugins](https://github.com/konveyor/crane-plugins) that updates the `index.yml` adding the latest release of the plugin and adds a `manifest` file for the plugin at appropriate location within `crane-plugins`.
+The above workflow raises a PR against [crane-plugin-<plugin-name>](https://github.com/konveyor/crane-plugin-openshift) that updates the `\manifests` adding the manifest yaml for latest release of the plugin.
 
-The next step would be to review this PR and merge it to the `main` branch of `crane-plugins`.
+The next step would be to review this PR and merge it to the `main` branch of `crane-plugin-<plugin-name>`.
 
-And the final step after testing the changes of `main` branch would be to release these manifest files in to `crane-cli` branch, for that we would need to cherry pick a tested commit and change `url` of manifest files in `index.yml` (i.e changing from `main` to `crane-cli` in index.yml for every entry)
```
shawn-hurley, line 171: We still need something to update the index.

enhancements/crane-2.0/plugin-release/README.md
```diff
@@ -199,19 +203,16 @@ N/A
 
 ## Drawbacks
 
-1. We need to update `release.yml` file whenever `metadata` changes for a plugin (i.e optional field, description, etc)
-2. Manual merge for any PR against `crane-plugin` is required (at the time of multiple plugin release at the same time it may require additional attention to resolve conflicts amongst PRs)
+1. Plugin author has to make sure that there are no bugs and gaps in the new plugin release that is intended.
```
shawn-hurley, line 206: We need to make sure that it is easy for a user to determine WHERE a bug could be coming from.

This may be a follow on PR but if I cant figure out which plugin stopped working I am just going to blame Crane and we (the Crane team) will have to become traffic cops for the issues of plugins.

enhancements/crane-2.0/plugin-release/README.md
```diff
@@ -199,19 +203,16 @@ N/A
...
 
 ## Alternatives
 
-To overcome the drawback of the solution below are other approaches. The below are the alternatives for the process of manually raising PR against `crane-plugins` and not the alternatives of whole solution proposed. We want to have a release workflow for the plugin.
+To overcome the drawback of the solution below are other approaches.
 
 ### Alternative 1
 
-We can set up a webhook in plugin repository that sends a request to a CI (potentially configured in `crane-plugins`) after release, this event will causes CI to update the `index.yml` file and add a manifest file for the plugin in the main branch.
+The current implementation of crane plugins allow us to test all the plugin while they are not GA but released through `main` branch and then after making everything sure we make them GA through`crane-lib` branch
```
shawn-hurley, line 214: This will be confusing phrasing once this update is the "current" implementation I think.

enhancements/crane-2.0/plugin-release/README.md
JaydipGabani: @djzager This will mean keeping manifests and index within `crane-plugins` just a little restructuring, and sill upon releasing any plugin we would need to change the respective manifest file in `crane-plugins` (but in this approach, we won't modify `index.yaml`). Is this assumption correct? and in this case, we need to update `index.yaml` only when a brand new plugin gets added, is that right?

I am struggling to understand what benefit will this have since pertaining to a release the only change we are making is not updating the index file. Can you clarify that for me? The assumption is we want to get rid of the `crane-cli` branch and want to handle everything from the `main` branch.

cc: @sseago @shawn-hurley

shawn-hurley (commented):
VISACK from me

enhancements/crane-2.0/plugin-release/redesign.md
```diff
@@ -0,0 +1,371 @@
...
+- name: OpenShiftPlugin
+  path: https://github.com/konveyor/crane-plugins/raw/main/plugins/OpenShift/index.yaml
+
+```
+
+There will be a single manifest file for each plugin. Each of the manifest file of respective plugin will include all the information of every release of the plugin. Reformated manifest file would look something like,
+
+```
+apiVersion: konveyor.io/v1alpha1
+kind: Plugin
```
sseago, line 80: I wonder whether we should call this `PluginManifest` rather than `Plugin` (and `PluginManifestIndex` instead of `PluginIndex`)

sseago (commented):
Looks good overall -- now that we've added the `kind` field to the manifest files, I had a question on naming it, though (see the comment inline).

enhancements/crane-2.0/plugin-release/redesign.md
```diff
@@ -0,0 +1,371 @@
...
+### Non-Goals
+
+Not applicable
+
+## Proposal
+
+The `index.yaml` in `crane-plugins` will only include mapping of a plugin with respective location of manifest (url/file path).
+
+```
+apiVersion: konveyor.io/v1alpha1
```
djzager, line 67: nit.
```suggestion
apiVersion: crane.konveyor.io/v1alpha1
```

like `crane.konveyor.io` that we have in mtrho-operator.

enhancements/crane-2.0/plugin-release/redesign.md
```diff
@@ -0,0 +1,371 @@
...
+plugins:
+- name: OpenShiftPlugin
+  path: https://github.com/konveyor/crane-plugins/raw/main/plugins/OpenShift/index.yaml
+
+```
+
+There will be a single manifest file for each plugin. Each of the manifest file of respective plugin will include all the information of every release of the plugin. Reformated manifest file would look something like,
+
+```
+apiVersion: konveyor.io/v1alpha1
```
djzager, line 79: nit.
```suggestion
apiVersion: crane.konveyor.io/v1alpha1
```

like `crane.konveyor.io` that we have in mtrho-operator.

enhancements/crane-2.0/plugin-release/redesign.md
JaydipGabani: Makes sense to me

enhancements/crane-2.0/plugin-release/redesign.md
```diff
@@ -0,0 +1,371 @@
...
+  - "@shawn-hurley"
+  - "@sseago"
+  - "@djzager"
+approvers:
+  - "@shawn-hurley"
+  - "@sseago"
+  - "@djzager"
+creation-date: 2021-11-05
+last-updated: 2022-04-04
+status: provisional|implementable|implemented|deferred|rejected|withdrawn|replaced
```
djzager, line 15:
```suggestion
status: implementable
```

enhancements/crane-2.0/plugin-release/redesign.md
```diff
@@ -0,0 +1,371 @@
...
+see-also:
+  - "N/A" 
+replaces:
+  - "N/A"
```
djzager, line 19: This should reference the old one.

enhancements/crane-2.0/plugin-release/redesign.md
```diff
@@ -0,0 +1,371 @@
...
+superseded-by:
+  - "N/A"
+---
+
+# Release process for plugins and keeping Crane-plugins upto date with all releases
+
+This document proposes a redesign of release process as well as re-structure of `crane-plugins` a little to accommodate simplified process of plugin release. The motivation behind achieving this simplicity is to make release process as simpler as it could be for a plugin author.
```
djzager, line 26:
```suggestion
This document proposes a redesign of release process as well as re-structure of `crane-plugins` a little to accommodate simplified process of plugin release. The motivation behind achieving this simplicity is to make release process simpler for the plugin author.
```

sseago (approved):

enhancements/crane-2.0/plugin-release/redesign.md
```diff
@@ -0,0 +1,371 @@
...
+Not needed
+
+## Summary
+
+We are trying to make release process simpler and decouple the need of updating index file with individual releases of each plugin so the changes needed in `crane-plugins` to include new releases of any plugins could be limited to respective plugin's manifest file. Alongside making the process simpler to accommodate this simplicity, we need to restructure `crane-plugins`.
+
+<b> Relevant repositories </b>
+
+- [crane-plugins](https://github.com/konveyor/crane-plugins): repository where all the plugin manifest files are present
+- [crane-plugin-<plugin-name>](https://github.com/konveyor/crane-plugin-openshift): a plugin repository where individual plugin code lives. (For example, openshift plugin)
```
djzager, line 46: Not 100% certain what you are going for here, maybe you want to move the link to the "for example, OpenShift plugin"?

enhancements/crane-2.0/plugin-release/redesign.md
shawn-hurley: Can someone explain to me what the extra information is adding vs what it was before?

djzager: I would like to avoid drawing out this conversation too long. But I will say that the first things that came to mind when I saw this thread were:
* What is a PluginManifest?
* How is a PluginManifest different from a Plugin?
* (like @shawn-hurley's comment) What new information is being conveyed by "Manifest"?

Looking at the wiki for [Manifest files](https://en.wikipedia.org/wiki/Manifest_file), I just don't follow what we are trying to convey here with manifest. I am sorry for not getting it, could you help me understand?

sseago: @djzager A PluginManifest *describes* a plugin. A Plugin actually manipulates things. It feels wrong to describe this manifest as a Plugin, since this is just the file format that the plugin manager consumes so that it knows how to find plugins. This yaml has always been called a manifest -- it's just that until this enhancement, we've never named it in the format itself.

sseago: Basically we've always called this a manifest, but the format doesn't currently have a "kind" element in it. Or, more accurately, until this enhancement, the manifest combined the index and manifest in one, since we had one file for everything.

djzager: Isn't the assumption that a GVK of `crane.konveyor.io/v1alpha1:Plugin` imply that it is the definition of a Plugin (a description of a plugin if you will) and not an actual plugin that does work or manipulates things?

My position is that calling it a `PluginManifest` is redundant because, of course these YAML files don't do actual work.

shawn-hurley: +1 to what @djzager said.

While this is a Manifest file, the "kind" should be what it describes (i.e. Backup vs. BackupDefinition or BackupManifest in velero.)

djzager: @JaydipGabani This enhancement looks good to me, as I see it resolving this conversation is all that remains.

djzager (approved):

PR #68, opened 2022-05-09
Propose quiesce subcommand in crane

pranavgaikwad:


enhancements/crane-2.0/quiesce-workloads/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Summary
+
+Quiesce _ClusterTask_ in Crane uses `kubectl scale` command to bring workloads down by setting replicas of workloads to 0. The _ClusterTask_ cannot be used by CLI users. To enable Cutover migrations using Crane, the CLI users need a way to quiesce down workloads in the source namespace.
+
+## Motivation
+
+### Goals
+
+- Enable first class cutover migration flow for Crane CLI users
```
shawn-hurley, line 45: What are the steps that the crane CLI does for this now?

enhancements/crane-2.0/quiesce-workloads/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+- Allow users to quiesce up/down workloads in source namespace
+
+- Provide a single subcommand to bring down all supported workloads at once
+
+- Enable quiescing of DaemonSet resources which is currently not possible using `kubectl`
```
shawn-hurley, line 51: This is by design right? how do we know whatever we do is safe for the user's workload?

shawn-hurley (changes requested):
I am not personally convinced that this is needed, that it is the right thing, or that it would fit into the philosophy of sharp tools. 

Can you write out, today, what the User experience is for this and then what the user experience would be after the fact.

I am concerned about calling this anything other than "scale" or something similar.

pranavgaikwad:
After discussion with @alaypatel07 @djzager @eriknelson @shawn-hurley we are rejecting this enhancement. Highlights from the discussion:
- Quiescing applications correctly depends on the nature of the application. It is a problem that cannot be reliably solved with a subcommand. The "quiesce" subcommand will only work for best cases and will leave room for hard to solve corner cases.
- Crane is meant to be a collection of sharp tools which works flawlessly for anything Kubernetes. The quiesce subcommand will only work on a very small subset of workload resources in Kubernetes. It doesn't fit into the philosophy of sharp tools. 
- Implementing a subcommand like quiesce in Crane gives users a false hope that it will intelligently quiesce their applications. It is not true. Quiesce subcommand discussed in this enhancement is not intelligent. It is simply taking resource replicas down.

Conclusion: 
- Merge this enhancement as "rejected"
- Work on a follow up enhancement PR which describes a potential "scale" subcommand  instead. It will simply attempt to scale down the resources. The enhancement will compare the user experience of scale command and compare it with equivalent bash scripts that achieve the same result.

djzager (approved):

shawn-hurley (approved):
Thanks for the discussion and for summary.

PR #69, opened 2022-05-16
Propose crane scale subcommand

pranavgaikwad:


enhancements/crane-2.0/scale-workloads/README.md
```diff
@@ -0,0 +1,183 @@
...
+  done
+done
+```
+
+### Implementation Details/Notes/Constraints [optional]
+
+Scaling up daemonsets with the bash script is a little tricky. The temporary annotation that stores the pre-scaling value of annotation needs to be converted from string back to JSON before patching the DaemonSet. 
+
+The bash script also doesn't handle errors well. Some extra logic is needed to incorporate error handling in the above script.
+
```
djzager, line 183: Need the alternatives section to make sure we have clear wording for what we considered as a possibility and why we rejected it.

djzager (commented):
Few comments.

enhancements/crane-2.0/scale-workloads/README.md
```diff
@@ -0,0 +1,183 @@
...
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [x] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] User-facing documentation is created
+
+## Summary
+
+For cutover migrations in Crane, users are expected to quiesce their workloads. It is to ensure that new data is not written to Persistent Volumes during migration. This enhancement proposes a new subcommand `scale` in Crane CLI. It will help users automate scaling down/up resources in the source namespaces. The subcommand will work on known resources in Kubernetes.
```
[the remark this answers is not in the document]
alaypatel07: +1. spent some time this morning to do this ugly hack, need to prettify it but the following works:

``` bash
$ for group in $(k get --raw /apis/ | jq -r '.groups[].name'); do
    preferredGroupVersion=$(k get --raw /apis/$group | jq -r .preferredVersion.groupVersion); 
    scaleEndpoint=$(k get --raw /apis/$preferredGroupVersion | jq '.resources[]|select(.name | endswith("/scale"))|.name'); 
    [ -n "$scaleEndpoint" ] && echo $preferredGroupVersion $scaleEndpoint;
  done
apps/v1 "deployments/scale" "replicasets/scale" "statefulsets/scale"
```

Note: my env is minikube so it is missing the openshift scale endpoints

enhancements/crane-2.0/scale-workloads/README.md
```diff
@@ -0,0 +1,183 @@
...
+
+## Motivation
+
+Crane UI provides a _ClusterTask_ that internally runs `kubectl scale` command to bring down applications in the source namespace. CLI users are expected to use `kubectl scale` themselves to scale down individual resources. `kubectl scale` does not work with resources like Jobs, CronJobs and DaemonSets. The user experience can be improved by providing a similar `scale` command in Crane which will work on all resources in source namespaces at once.
```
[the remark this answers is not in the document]
shawn-hurley: What is the reason that these resources do not have a scale sub resources? Have we determined that, even though we can do something, are there edge cases that we have not considered?

enhancements/crane-2.0/scale-workloads/README.md
```diff
@@ -0,0 +1,183 @@
...
+
+### Goals
+
+- Provide an automated way to scale standard Kubernetes workload resources in source namespaces
```
shawn-hurley, line 47: There is a standard way, it is to use the `/scale` subresource. From what I can tell the goal is to add handling of resources that do not conform to the standard to look like they can?

enhancements/crane-2.0/scale-workloads/README.md
```diff
@@ -0,0 +1,183 @@
...
+### Non-Goals
+
+- Quiesce user workloads intelligently 
+
+## Proposal
+
+A new subcommand `scale` will be introduced in Crane CLI. The `scale` command can be used to scale down Deployments, DeploymentConfigs, Statefulsets, ReplicaSets, Jobs, CronJobs and Daemonsets by running:
+
+```bash
+> crane scale down --context <source_context> --namespace <namespace>
```
shawn-hurley, line 60: is context a required field?

pranavgaikwad: @shawn-hurley Yes, assuming we only want to do this on source cluster. We do not want users accidently executing this on a wrong cluster.

enhancements/crane-2.0/scale-workloads/README.md
pranavgaikwad: @shawn-hurley Yes, the idea is to include Jobs,CronJobs and DaemonSets too

enhancements/crane-2.0/scale-workloads/README.md
pranavgaikwad: @djzager ACK

enhancements/crane-2.0/scale-workloads/README.md
```diff
@@ -0,0 +1,183 @@
...
+```
+
+To scale the resources back to original state:
+
+```bash
+> crane scale up --context <source_context> --namespace <namespace>
```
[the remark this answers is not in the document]
pranavgaikwad: ACK

enhancements/crane-2.0/scale-workloads/README.md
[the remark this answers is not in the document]
pranavgaikwad: Thanks for your inputs folks. I will modify the enhancement to consider `/scale` endpoint instead. 

@alaypatel07 that's an interesting idea.

jmontleon (approved):

enhancements/crane-2.0/scale-workloads/README.md
shawn-hurley: What happens if they do this on the wrong cluster in practice?

pranavgaikwad: @shawn-hurley they would scale down the applications on destination if they are running.

shawn-hurley: and what would be the problem with this? I don't see why we have an extra required flag for something that doesn't seem like it would be a big deal.

pranavgaikwad: @shawn-hurley Back in the day, working with pvc-migrate, I have seen mixups happen in kube contexts. So my first thought is to make contexts explicit everywhere. Totally open to making this optional.

pranavgaikwad: @shawn-hurley @djzager I'll make it optional

enhancements/crane-2.0/scale-workloads/README.md
```diff
@@ -0,0 +1,195 @@
...
+- Quiesce user workloads intelligently 
+## Proposal
+
+A new subcommand `scale` will be introduced in Crane CLI. The `scale` command can be used to scale down Deployments, DeploymentConfigs, Statefulsets, ReplicaSets, Jobs, CronJobs and Daemonsets by running:
+
+```bash
+> crane scale down --context <source_context> --namespace <namespace>
+```
+
+The `scale down` subcommand will discover all resources in the source namespace. If the discovered resource implements scale subresource, it will store the value of field specified by `specReplicas` in an annotation. It will set the replicas value to 0. This will be repeated for all resources in the namespace. Finally, the subcommand will wait until all Pods that have ownerReferences set are Terminated. Additionally, if the discovered resource is a Job or CronJob, the subcommand will store the value of `spec.suspend` field in the annotation and set its value to `True` in order to scale them down. For Daemonset resources, the subcommand will set a unique dummy nodeSelector on the DaemonSet Pods in order to force un-schedule them. It will store the existing value of nodeSelector in the annotation.
```
shawn-hurley, line 66: One last minor point, you should instead of use the `Replicas` use the scale subresource to 0

shawn-hurley (approved):
if we instead use the scale subresource I approve this PR.

PR #71, opened 2022-06-15
Proposal to change namespace labels and DVM approach to keep up with upcoming PSA changes in 4.12

JaydipGabani:


enhancements/crane-1.x/redesign-mtc-4.11-4.12/README.md
```diff
@@ -0,0 +1,107 @@
...
+- `pod-security.kubernetes.io/enforce: privileged`
+- `pod-security.kubernetes.io/audit: privileged`
+- `pod-security.kubernetes.io/warn: privileged`
+
+
+#### For the worklods running in user namespace
+
+Rsyn pods run outside of `openshift-migration` namespace (in user namespaece) and need to run as root (with uid 0) to preserve UID/GID while migrating files. Running rsync pods as root with uid 0 will no longer be possible with default pod profile being `restricted` and without adding required labels to the user namespace to allow escalation to `priviledge`. Adding labels in user namespace is not reasonable. So now there are two possible way to tackle this situation as below.
+
+##### Not run rsync pods as root with uid 0
```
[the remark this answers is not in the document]
shawn-hurley: I think it is worth a quick look over the existing drivers and determining the most used and what they do.

shawn-hurley (changes requested):
While I generally agree with the Not run as the default, 

we need an escape hatch for the users.
we need to do the due diligence to determine what type of customers/users would be impacted and would need to make a change.

enhancements/crane-1.x/redesign-mtc-4.11-4.12/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+Rsync pods need to run as root to preserve UID/GID of the migrating files. We can choose not to preserve UID/GID and simply run the rsycn operator with `restricted` profile. We can acknowledge that with this change direct and inderect migration will defer in this aspect that with direct mogration UID/GID will not be maintained anymore, and with indirect migration UID/GID of migrating files will be maintained as is.
```
pranavgaikwad, line 88: As I said above, not preserving UID/GID wont be an accurate statement to make. They can still be preserved by running the Rsync Pods with correct UID/GIDs. OpenShift SCCs mutating capabilities can be leveraged to automatically assign the Pods right UIDs. This is explained in detail here : https://github.com/konveyor/enhancements/tree/master/enhancements/crane-2.0/ownership-change#proposal

pranavgaikwad (changes requested):

enhancements/crane-1.x/redesign-mtc-4.11-4.12/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+**Pros:**
+- Easy to implement
+
+**Cons:**
+- Not preserving UID/GID of files that are getting migrated
+
+## Alternatives
+
+Alternative solution to not running rsync pod as root with uid 0 is to adapt something similar to how restic works right now. Theoratically, we should be able to run rsync pod within `openshift-migration` namespace and spin up a new pod (similar to stunnel pod or stage pod) that mounts the volume in user namespace. Then through mount propogation we could access the volume through rsync pod and run rsync operations with running as root and preserving UID/GID of migrating files.
```
sseago, line 98: As with restic, this pod won't be mounting the volume directly -- it uses mount propagation to access volumes mounted on the node.

enhancements/crane-1.x/redesign-mtc-4.11-4.12/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+**Pros**:
+- UID/GID of the files migrated are getting preserved and there is no differences in terms of migration in direct and inderect migration.
+  
+**Cons**:
+- Difficult to implement
+- Changes current and proven design tremandously, might unearth some unknown edge cases 
+- Needs to be tested in the environment like non VPC ibm cluster where RWO is a little different and only one pod is allowed to mount a volume.
```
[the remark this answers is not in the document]
sseago: The RWO issue doesn't really come up here, since we're not adding a second mounting pod -- the openshift-migration rsync pod won't be mounting the PVC, it will be running on the same node as the mounting dummy pod and accessing the node-mounted volumes via mount propagation.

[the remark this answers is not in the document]
sseago: (so as long as Restic works without RWO violations, this will too)

enhancements/crane-1.x/redesign-mtc-4.11-4.12/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+### Goals
+
+1. Determine the changes that are needed to keep working in OpenShift 4.12 and onwards
+
+## Proposal
+
+The proposed solution includes two changes:
+1. Adding required labels to `openshift-migration` namespace to allow the worklods within this namespace to run with priviledge profile.
+2. For the workloads running in user namespace such as `rsync` pods, we have two alternates and of which our proposed solution is to not run the pod in priviledge mod and migrate files without preserving UID/GID.
```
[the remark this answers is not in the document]
JaydipGabani: Updated to reflect accuratly

shawn-hurley (commented):
Looking good, I do like the changes and the optionality makes sense. 

1. What happens when a user says do root, but it won't run. how do we tell the user this going to be a problem
2. Can we document the times when you need to keep the GID/UID and when we don't. I think it will be hard for a user to know without some explanation from us.

pranavgaikwad (changes requested):
I have some concerns around capabilities and new options to be introduced. It'd be good to call the details out for options.

enhancements/crane-1.x/redesign-mtc-4.11-4.12/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+We can also run rsync pod with specific UID/GID that is not root by setting SecurityContext of Rsync Containers:
+```
+securityContext:
+  runAsUser: 1000680000
+  runAsGroup: 1000680000
+```
+These UID/GID have to be within the allowed range so that we do not violate SCC or PSA. In order to allow custom values of UID/GID we might have to introduce some fields in DVM or migration CR.
+
+We can ask if users want to preserve the UID/GID data and wants to run rsync pods as root via setting a field may be in DVM or migration. If they want to we can add labels needed for root access in destination user namespace. By default, we do not add those labels in the destination namespace.
```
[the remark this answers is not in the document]
shawn-hurley: We should not do this automatically, they should be responsible for doing this.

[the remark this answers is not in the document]
JaydipGabani: Agree, I have updated the doc that talks about how to do this

enhancements/crane-1.x/redesign-mtc-4.11-4.12/README.md
```diff
@@ -0,0 +1,129 @@
...
+    - SETUID
+```
+
+We can also run rsync pod with specific UID/GID that is not root by setting SecurityContext of Rsync Containers:
+```
+securityContext:
+  runAsUser: 1000680000
+  runAsGroup: 1000680000
+```
+These UID/GID have to be within the allowed range so that we do not violate SCC or PSA. In order to allow custom values of UID/GID we might have to introduce some fields in DVM or migration CR.
```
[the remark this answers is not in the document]
JaydipGabani: done

enhancements/crane-1.x/redesign-mtc-4.11-4.12/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+##### Not run rsync pods as root with uid 0
+
+Rsync pods need to run as root to preserve UID/GID of the migrating files. We can choose to run the rsync operation with `restricted` profile. In this case rsync pods will run with the UID/GID provided by SCC that are within the range of allowed IDs of migrating namespace, and the files would have the same UID/GID as rsync pod. We can acknowledge that with this change direct and indirect migration will defer in this aspect that with direct migration UID/GID will not be maintained anymore, and with indirect migration UID/GID of migrating files will be maintained as is. We would also need to strip some capabilities from rsync pod to make sure it can be run with `restricted` profile.
```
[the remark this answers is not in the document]
JaydipGabani: I believe we already have checks for verifying if Rsync pods are running or not [here](https://github.com/konveyor/mig-controller/blob/master/pkg/controller/directvolumemigration/task.go#L338). So that should take care of the situation if pods don't come up or are in error state.

enhancements/crane-1.x/redesign-mtc-4.11-4.12/README.md
```diff
@@ -0,0 +1,165 @@
...
+- [PSA-autolabeling](https://github.com/openshift/enhancements/blob/master/enhancements/authentication/pod-security-admission-autolabeling.md)
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] User-facing documentation is created
+
+## Open Questions
```
shawn-hurley, line 48: Are these still open?

enhancements/crane-1.x/redesign-mtc-4.11-4.12/README.md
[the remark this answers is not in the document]
shawn-hurley: I think I need to be more clear, I believe the guidance should be that they do this. I don't want to make this a way to escalate myself for some period of time because we went and did something when they just click a box.


if the box is clicked we should verify the namespace has the correct labels.

enhancements/crane-1.x/redesign-mtc-4.11-4.12/README.md
```diff
@@ -0,0 +1,165 @@
...
+To implement the above proposal, the following changes are needed.
+
+- Introducing the following fields in migration CR
+
+```shell script
+// MigMigrationSpec defines the desired state of MigMigration
+type MigMigrationSpec struct {
+	...
+	// If set True, run rsync operations with escalated privileged, takes precedence over setting RunAsUser and RunAsGroup
+	RunAsRoot bool `json:"runasroot,omitempty"`
```
shawn-hurley, line 130: I think that false and not set are different, this should be a pointer.

enhancements/crane-1.x/redesign-mtc-4.11-4.12/README.md
```diff
@@ -0,0 +1,165 @@
...
+	RunAsUser *int64 `json:"runasuser,omitempty"`
+
+	// If set, runs rsync operations with provided group id. This provided user id should be a valid one that falls within the range of allowed GID of user namespace
+	RunAsGroup *int64 `json:"runasgroup,omitempty"`
+}
+```
+
+- Changing the way destination namespace and rsync pods are getting created
+    - The default behavior is to run rsync with minimal capabilities and `restricted` SCC with default UID/GID.
+    - If `RunAsRoot` is true, then create/patch destination namespace with needed labels to run the rsync pod as root. Run the migration as it would run as of now. After migration is complete, patch the namespace to remove added labels.
```
shawn-hurley, line 142: check if this has been done, if yes cool, then attempt to run as root. DO NOT update the namespace automatically IMO.

enhancements/crane-1.x/redesign-mtc-4.11-4.12/README.md
JaydipGabani: Nope, now the only question open is to determine use cases for rsync as root and non root

enhancements/crane-1.x/redesign-mtc-4.11-4.12/README.md
```diff
@@ -0,0 +1,159 @@
...
+	RunAsUser *int64 `json:"runasuser,omitempty"`
+
+	// If set, runs rsync operations with provided group id. This provided user id should be a valid one that falls within the range of allowed GID of user namespace
+	RunAsGroup *int64 `json:"runasgroup,omitempty"`
+}
+```
+
+- Changing the way destination namespace and rsync pods are getting created
+    - The default behavior is to run rsync with minimal capabilities and `restricted` SCC with default UID/GID.
+    - If `RunAsRoot` is set true by the user, then check if the destination namespace has the needed labels to run the rsync pod as root. Run the migration as it would run as of now. If the labels are not present in the destination namespace, then run rsync without root and warn User about the same in DVM CR.
```
pranavgaikwad, line 136: @JaydipGabani Will we set labels automatically? If yes, we need to inform the user that we are doing so.

enhancements/crane-1.x/redesign-mtc-4.11-4.12/README.md
```diff
@@ -0,0 +1,159 @@
...
+    - If `RunAsRoot` is not set and `RunAsUser` & `RunAsGroup` are provided then add this information to rsync pod definition to make sure rsync pod run with user provided UID/GID.
```
pranavgaikwad, line 137: @JaydipGabani What happens when only one of `RunAsUser` or `RunAsGroup` is provided?

enhancements/crane-1.x/redesign-mtc-4.11-4.12/README.md
```diff
@@ -0,0 +1,159 @@
...
+    
+- Update UI to ask users while running direct migration from final or state migration plan to run rsync as root or with specific UID/GID.
```
pranavgaikwad, line 139: I think we should remove anything UI specific from this enhancement. Right now, it isn't clear to me where it will belong in the UI as this is a migration specific setting and not at a plan level.

pranavgaikwad (changes requested):
Overall great improvement over initial versions. I still have some questions before we can merge this. Thank you for your patience @JaydipGabani

enhancements/crane-1.x/redesign-mtc-4.11-4.12/README.md
JaydipGabani: No we are not, @shawn-hurley suggested we should not be doing that, instead we ask user to create destination namespace with needed labels to run rsync pods with root. We will not in any case add those labels as of now.

enhancements/crane-1.x/redesign-mtc-4.11-4.12/README.md
JaydipGabani: Now there are no open questions left.

enhancements/crane-1.x/redesign-mtc-4.11-4.12/README.md
JaydipGabani: @pranavgaikwad we set the provided fields only and run rsync with the provided fields.

pranavgaikwad (approved):

enhancements/crane-1.x/redesign-mtc-4.11-4.12/README.md
```diff
@@ -0,0 +1,178 @@
...
+	RunAsGroup *int64 `json:"runasgroup,omitempty"`
+}
+```
+
+- Changing the way destination namespace and rsync pods are getting created
+    - The default behavior is to run rsync with minimal capabilities and `restricted` SCC with default UID/GID.
+    - If `RunAsRoot` is set true by the user, then check if the destination namespace has the needed labels to run the rsync pod as root. Run the migration as it would run as of now. If the labels are not present in the destination namespace, then run rsync without root and warn User about the same in DVM CR.
+    - If `RunAsRoot` is not set and `RunAsUser` & `RunAsGroup` are provided then add this information to rsync pod definition to make sure rsync pod run with user provided UID/GID.
+    
+#### Need of running rsync with different profiles
```
shawn-hurley, line 135: Note: we will want to make sure that we add this to our documentation and release notes about the change.

shawn-hurley (approved):
Thank you for taking the time to do the research

PR #72, opened 2022-06-22
Enhancement: Run KRM function from Crane

MundraAnkur:


enhancements/crane-2.0/run-functions/README.md
```diff
@@ -0,0 +1,143 @@
...
+Crane subcommand will follow this synopsis:
+```
+crane fn run [DIR] [flags] [-- args]
+```
+> The above subcommand will be used for transforming resources in a directory using functions. If a function fails, the process is aborted and the resource files will be left unchanged.
+ 
+#### Args:
+
+  *DIR:*
+> Path to the local directory containing resources. Defaults to the current working directory.
```
djzager, line 72: This would probably work on the `--export-dir` like you see from `crane transform`

enhancements/crane-2.0/run-functions/README.md
```diff
@@ -0,0 +1,143 @@
...
+		List of storage options to enable reading from the local filesystem.
+	  
+	  --network:
+		If enabled, container functions are allowed to access network.
+		By default it is disabled.
+  	
+	  --output, o:
+		If specified, the output resources are written to provided location,
+		if not specified, resources are modified in-place.
+		<OUT_DIR_PATH>: output resources are written to provided directory.
```
djzager, line 99: In keeping with `crane`'s philosophy, I believe that we want to *avoid* modifying files in place. We should create a new output directory to hold the results.

pranavgaikwad (changes requested):
@MundraAnkur Great work! Overall looks good, I have some questions (mostly for my own understanding)

enhancements/crane-2.0/run-functions/README.md
```diff
@@ -0,0 +1,143 @@
...
+title: run-krm-functions-from-crane
+authors:
+  - "@amundra"
+reviewers:
+  - "@jmatthew" 
+  - "@dzager"
+  - "@shurley"
+  - "@ernelson"
+  - "@dymurray"
+  - "@spampatt"
```
shubham-pampattiwar, line 11: NIT: @spampatt <-> @shubham-pampattiwar :)

enhancements/crane-2.0/run-functions/README.md
MundraAnkur: So, the default should be an export directory, and instead of mentioning `DIR` in command, we can utilize it under flag **--export-dir**?

enhancements/crane-2.0/run-functions/README.md
```diff
@@ -0,0 +1,143 @@
...
+$ crane fn run -i docker.io/example.com/set-namespace:v0.1 -o path/to/dest -- namespace=crane
+```
+
+### User Stories
+
+#### Story 1
+As a crane user, I would like to run a single KRM function with crane subcommand using function image as an argument and other command-line flags, such that the function is applied against the input resources and produces the transformed output. 
+
+#### Story 2
+As a crane user, I would like to run KRM functions **without** the crane cli subcommand, such that the given function is applied against the input resources and produces the transformed output.
```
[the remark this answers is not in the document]
MundraAnkur: So, we don't have to depend on cli commands to run the functions.  We can have a module where we pass in some kind of configuration file containing the resource directory, all the functions that we want to run, and any additional arguments that are needed for each function to run.

```
apiVersion: v1
kind: FunctionConfig
metadata:
  name: fn-config1
spec:
  input-resource-dir: testdata/
  functions:
    - image: quay.io/fnproject/fn-test-utils:latest
      env:
        FN_TEST_ENV: fn-test-env
        FN_TEST_ENV_2: fn-test-env-2
      mount:
        - type: bind
          src: path/to/fn-test-utils
          dst: /fn-test-utils
          readOnly: true
    - image: set-namespace:v0.2
      network: true
      data:
          namespace: guestbook
  output-resource-dir: testdata/transform/
```
Something like this can be passed to a function and it will execute the functions and store the transformed resources.

enhancements/crane-2.0/run-functions/README.md
```diff
@@ -0,0 +1,143 @@
...
+
+## Proposal
+- We can add a subcommand in the crane cli that can run KRM functions with the arguments. The underlying functionality can be built on top of [kyaml](https://pkg.go.dev/sigs.k8s.io/kustomize/kyaml) package that provides libraries to run containerized function images.
+ 
+- We can also create a library which will perform the similar tasks as a subcommand. It might be helpful to automate the execution of functions as we don't need to depend on cli commands to execute functions.
+
+### Command
+Crane subcommand will follow this synopsis:
+```
+crane fn run [DIR] [flags] [-- args]
```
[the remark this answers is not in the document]
MundraAnkur: @pranavgaikwad  @shubham-pampattiwar  I'm not really sure what the correct answer to your questions is but I guess initially we are trying to have a capability in the crane to handle KRM functions later we can expand this to utilize KRM functions internally in the subcommands.
 @djzager might be able to answer this more accurately

enhancements/crane-2.0/run-functions/README.md
```diff
@@ -0,0 +1,143 @@
...
+	
+  *args:*
+>	Arguments to pass to the function. The value can be in `key=value` format and come after the separator **'--'**
+
+  *Flags:*
+>
+    --image, i:
+		Container image of the function to execute i.e. *quay.io/krm-fn/set-annotation:v3.1*. 
+
+	  --as-current-user:
```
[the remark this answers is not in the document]
MundraAnkur: The flag is used to indicate whether the docker/podman container should use the uid and gid that run the command and if it is not specified container will run as "nobody"

enhancements/crane-2.0/run-functions/README.md
```diff
@@ -0,0 +1,143 @@
...
+
+#### Story 2
+As a crane user, I would like to run KRM functions **without** the crane cli subcommand, such that the given function is applied against the input resources and produces the transformed output. 
+
+
+#### Story 3
+As a crane user, I would like to run multiple functions in a predefined manner such that they are applied to input resources and produces a transformed output.
+
+## Alternatives
+- Alternative to creating an integrated solution to crane cli, we can leverage the already built solutions that provide similar functionality like [KPT](https://github.com/GoogleContainerTools/kpt) and [Kustomize](https://github.com/kubernetes-sigs/kustomize).
```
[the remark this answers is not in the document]
MundraAnkur: Actually, I'm not really sure why the alternative solution will not work cos, eventually, we are transforming resources in the local directory by applying functions and that can be done with the existing solution `kpt | kustomize`. 
The only thing that comes to my mind is when we want to use the functions internally without the cli commands, we need to have our own solution which will help us run functions in the pipeline without external help.

enhancements/crane-2.0/run-functions/README.md
```diff
@@ -0,0 +1,165 @@
...
+At present, we can not utilize the KRM functions in the crane. To leverage the benefits of KRM functions and aligning with the community standards, we are proposing to add the capability in the crane CLI to execute a function. 
+
+### Goals
+
+- Propose solution to handle execution of KRM functions in crane.
+- Implement the proposed solution.
+ 
+
+## Proposal
+- We will add a subcommand in the crane cli that can run KRM functions with the arguments. The underlying functionality can be built on top of [kyaml](https://pkg.go.dev/sigs.k8s.io/kustomize/kyaml) package that provides libraries to run containerized function images. We'll also create a library which will perform the similar tasks as a subcommand. It might be helpful to automate the execution of functions as we don't need to depend on cli commands to execute functions.
```
shawn-hurley, line 54: I am starting to get a little confused on the proposal. 

What is the purpose of the library that we would build? how would you automat something without using the CLI's? are you suggesting that someone builds a binary to run specific KRM functions?

enhancements/crane-2.0/run-functions/README.md
```diff
@@ -0,0 +1,165 @@
...
+> The above subcommand will be used for transforming resources in a directory using functions. If a function fails, the process is aborted and the resource files will be left unchanged.
+ 
+#### Args:
+	
+  *args:*
+>	Arguments to pass to the function. The value can be in `key=value` format and come after the separator **'--'**
+
+  *Flags:*
+>
+	  --image, i: (required)
```
shawn-hurley, line 70: why not just make this a positional arg? if I always have to have it adding `-i` could become tiresome.

enhancements/crane-2.0/run-functions/README.md
```diff
@@ -0,0 +1,165 @@
...
+  
+	  --env:
+		List of local environment variables to be exported to the container function.
+		The value can be in key=value format or only the key of an already exported environment variable..
+  
+	  --mount:
+		List of storage options to enable reading from the local filesystem.
+	  
+	  --network:
+		If enabled, container functions are allowed to access network.
```
shawn-hurley, line 85: is this an option in the library already?

enhancements/crane-2.0/run-functions/README.md
```diff
@@ -0,0 +1,165 @@
...
+$ crane fn run -i docker.io/example.com/set-namespace:v0.1 --transform-dir path/to/dest -- namespace=crane
+```
+
+### User Stories
+
+#### Story 1
+As a crane user, I would like to run a single KRM function with crane subcommand using function image as an argument and other command-line flags, such that the function is applied against the input resources and produces the transformed output. 
+
+#### Story 2
+As a crane user, I would like to run KRM functions **without** the crane cli subcommand, such that the given function is applied against the input resources and produces the transformed output.
```
shawn-hurley, line 131: I think that we need to flesh this story out some more, what does the user do?

enhancements/crane-2.0/run-functions/README.md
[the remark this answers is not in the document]
shawn-hurley: It would be nice to more fully flesh that alternative out. what would a user do while using a combination of crane/kpt/kustomize?

shawn-hurley (changes requested):
I think that some of the user stories make sense.


I would also like to discuss if we want to expose some options like network or as-current-user. I am not clear that these are needed and a justification or a use case for them might be helpful IMO.

enhancements/crane-2.0/run-functions/README.md
MundraAnkur: Yes, it's an option in the [library](https://pkg.go.dev/sigs.k8s.io/kustomize/kyaml@v0.13.7/fn/runtime/runtimeutil#ContainerSpec)

enhancements/crane-2.0/run-functions/README.md
MundraAnkur: Yeah, we can make it a positional arg.
If we support binaries in the future then it will be an optional arg

enhancements/crane-2.0/run-functions/README.md
```diff
@@ -0,0 +1,165 @@
...
+Crane subcommand will follow this synopsis:
+```
+crane fn run [flags] [-- args]
+```
+> The above subcommand will be used for transforming resources in a directory using functions. If a function fails, the process is aborted and the resource files will be left unchanged.
+ 
+#### Args:
+	
+  *args:*
+>	Arguments to pass to the function. The value can be in `key=value` format and come after the separator **'--'**
```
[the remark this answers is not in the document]
MundraAnkur: arg: app=todolist stage=production (set these labels)
crane fn run -i docker.io/example.com/set-labels:v0.1 -- **app=todolist stage=production**

enhancements/crane-2.0/run-functions/README.md
MundraAnkur: As I have mentioned [here](https://github.com/konveyor/enhancements/pull/72#discussion_r905335070), let's say we want to execute some predefined set of functions internally (for ex: when `crane transform` is executed, we have to some basic transformation tasks which are fixed e.g. clean-metadata, remove status, etc), we won't have to separately run `crane fn run clean-metadata: latest`, `crane fn run remove-status: latest` 

The library won't have any special functionality, it's just the same piece of logic that will additionally parse the config file containing functions and necessary details and execute them.

enhancements/crane-2.0/run-functions/README.md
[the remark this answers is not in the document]
MundraAnkur: If we are not running the functions under some other commands (e.g. `crane transform`) internally then all we need to do is run functions on the resources present locally and transform them.

Scenario:
1. User export the resources using `crane export`, then 
2. User transform the resources with `kpt | kustomize`  and store them in the transform directory, finally
3. User applies the transformed resources to the target using `crane apply`

**Let's discuss this and the proposal for a library(maybe just a few methods) to fill the gap in my understanding**

enhancements/crane-2.0/run-functions/README.md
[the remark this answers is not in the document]
shawn-hurley: hmm I think we have prior art here @sseago what is the current way we do this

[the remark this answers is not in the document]
sseago: @shawn-hurley so we currently have a way to pass args from the CLI to binary (and internal plugins -- basically anything implementing the Plugin interface). If we want to make use of that, we just need to work out a way to implement a KRMPlugin type that's analogous to the BinaryPlugin type -- then we'd get all the arg-passing from CLI for free.

enhancements/crane-2.0/run-functions/README.md
```diff
@@ -0,0 +1,174 @@
...
+As a crane user, I would like to run multiple functions in a predefined manner such that they are applied to input resources and produces a transformed output.
+
+
+## Alternatives
+Alternative to creating an integrated solution to crane cli, we can leverage the already built solutions that provide similar functionality like [KPT](https://github.com/GoogleContainerTools/kpt) and [Kustomize](https://github.com/kubernetes-sigs/kustomize).
+
+At present, user can perform the following steps to utilize KRM functions in crane:
+ 1. User export the resources using `crane export`, then 
+ 2. User transform the resources with `kpt | kustomize`  and store them in the transform directory, finally
+ 3. User applies the transformed resources to the target using `crane apply`
```
djzager, line 169: :grin: `crane apply` only takes json patches (from `crane transform`) and "apply"s those patches to the original manifests (from `crane export`) to give you new manifests that could be `kubectl apply`ed to another cluster.

In this proposed alternative, `crane` would only be useful for the resource discovery piece (or `crane export`), and the rest would need to be handled with `kpt | kustomize`.

djzager (commented):
All of my concerns have been addressed. I'll leave the `lgtm` to others who have requested changes.

Thank you for your efforts.

pranavgaikwad (approved):
/lgtm

PR #73, opened 2022-06-23
Proposal to create a KRM Function Manager

JonahSussman:


enhancements/common/krm-function-manager.md
```diff
@@ -0,0 +1,321 @@
...
+- [ ] User-facing documentation is created
+
+## Open Questions
+
+1. What is the exact format of the outputted data?
+2. How could the system be extended to include the management of entire pipelines?
+
+## Summary
+
+This enhancement proposes a new tool, called **Kaffine** to help manage the discovery, installation, removal, and updating of KRM functions. The tool acts as a buffer between the low-level management of different KRM function versions and the high-level usage of them in an application.
```
shubham-pampattiwar, line 38: +1 on the name `Kaffine` ;)

enhancements/common/krm-function-manager.md
```diff
@@ -0,0 +1,321 @@
...
+
+## Motivation
+
+Currently, when creating an application that utilizes KRM functions, there is no standardized way to manage the versioning these functions. While it is possible to manage the images of the functions manually, it is incredibly un-ergonomic to do so. The simplest solution to avoid managing versions is to use the `latest` tag when developing the application. Unfortunately, this produces three major undesirable effects:
+
+1. **Lack of reproducibility**: A developer using `latest` might start out with `v1` on their system. However, when they go to deploy on a different system, `latest` might pull a higher version with breaking changes.
+1. **Security vulnerabilities**: Many enterprise solutions will simply not tolerate blindly updating to the latest version of a technology before it is proven.
+1. **Dependant on container technology**: While KRM functions utilize container technology for now, if that requirement is ever relaxed to include raw binaries, relying on tags will simply not work.
```
shubham-pampattiwar, line 46: +1 on this motivation point

pranavgaikwad (commented):
I like the idea a lot. Good job on putting this all together. Great enhancement!
 
Before commenting on details, I want to make sure I understand the purpose of this tool: 
1. How does this tool relate to Crane? and potential Crane KRM function runner? 
2. If a separate tool is to built, I believe the best place it belongs to is upstream Kubernetes community. Am I wrong or you already planning to propose upstram?

enhancements/common/krm-function-manager.md
```diff
@@ -0,0 +1,321 @@
...
+
+## Summary
+
+This enhancement proposes a new tool, called **Kaffine** to help manage the discovery, installation, removal, and updating of KRM functions. The tool acts as a buffer between the low-level management of different KRM function versions and the high-level usage of them in an application.
+
+## Motivation
+
+Currently, when creating an application that utilizes KRM functions, there is no standardized way to manage the versioning these functions. While it is possible to manage the images of the functions manually, it is incredibly un-ergonomic to do so. The simplest solution to avoid managing versions is to use the `latest` tag when developing the application. Unfortunately, this produces three major undesirable effects:
+
+1. **Lack of reproducibility**: A developer using `latest` might start out with `v1` on their system. However, when they go to deploy on a different system, `latest` might pull a higher version with breaking changes.
```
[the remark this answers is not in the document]
JonahSussman: That's correct. However, let's say that `v2` comes out with enhanced bug fixes and the developer uses a tool that [structures their function calls similar to KPT](https://kpt.dev/book/04-using-functions/01-declarative-function-execution). One function might be called multiple times in many different locations, each with its own `v1` tag, leading to a refactoring headache. The problem gets compounded the more a particular function is used.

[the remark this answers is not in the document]
shubham-pampattiwar: The developer could use a KRM function to replace the tags, right ? (where you pass the resourceList of kind `kptfile` and mutate the image value)

enhancements/common/krm-function-manager.md
```diff
@@ -0,0 +1,321 @@
...
+Alice has in mind some functions that she wants to use that are created by Group Alpha. She adds the repo their repo via `kaffine config add repo group-alpha.example.com`. Then she searches which functions are available in the repo.
+
+```
+$ kaffine search -l group-alpha
+KIND      NAME          DESCRIPTION           LATEST  INSTALLED
+function  alpha-foo-er  A function that foos  v2.0.0     ---
+function  alpha-bar-er  A function that bars  v1.0.3     ---
+function  alpha-baz-er  A function that bazs  v3.1.0     ---
+
+$ kaffine install alpha-foo-er
```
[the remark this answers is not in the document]
JonahSussman: I was actually hoping to get feedback on the best way to approach this. I see two ways `Kaffine` could perform this action.

1. `Kaffine` will only manage references to the images. It passes along a YAML file to the function consumer where it handles downloading and installing the container.
2. `Kaffine` downloads and installs the container, perhaps even giving it some extra tagging information so it's local to the project. It passes along a YAML file with the local install of the container referenced.

Option 2 is significantly more complicated and obtrusive than 1. Additionally, if we're not careful, I could see how we'd end up with a `node_modules`-style bloat situation. Thus, for ease of "opting-in," I envision it working like option 1 for now.

JonahSussman:
@pranavgaikwad Thank you! With regards to your questions:

1. Crane, to my understanding, has a variety of OpenShift plugins that could be converted to KRM functions. This tool will be able to manage the versioning of each function used. Thus, the runner can focus on just running the functions. (Additionally, at some point in the future I envision being able to download and install entire pipelines with Kaffine, which will each require their own dependencies to manage)
2. I have not considered the possibility of proposing upstream. I would love to hear others' thoughts on this!

enhancements/common/krm-function-manager.md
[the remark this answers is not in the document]
JonahSussman: That could certainly be a possibility! I think the real power of a tool like this starts to show when you take into account the ease with which you can update functions. A single `update` command and Kaffine will handle the entire process of disovering updates and swapping the versioning for you.

enhancements/common/krm-function-manager.md
djzager, line 38: Primarily though, `Kaffine` is concerned about discovery and updating right? I don't see a whole lot of need to install/remove containers (unless, as you said, we use binaries instead). So the biggest draw is the ability to know when new versions of a function are available...right?

enhancements/common/krm-function-manager.md
```diff
@@ -0,0 +1,321 @@
...
+- Be able to add or remove repositories of KRM functions to the tool
+- Search the repositories for functions and install them
+- Update the installed functions
+- Easily opt-in or opt-out of the system
+
+### Non-Goals
+
+- Running KRM functions
+- Managing how applications use functions
+<!-- - Actually installing or uninstalling containers -->
```
djzager, line 68: I agree that we probably shouldn't be installing/removing container images. Maybe `docker|podman pull` but removing them doesn't seem like something we want to start.

enhancements/common/krm-function-manager.md
```diff
@@ -0,0 +1,321 @@
...
+
+## Proposal
+
+### User Stories
+
+#### Story 1: Creating a new application
+
+Alice has in mind some functions that she wants to use that are created by Group Alpha. She adds the repo their repo via `kaffine config add repo group-alpha.example.com`. Then she searches which functions are available in the repo.
```
djzager, line 76: This is the first I've seen the idea of a KRM Function repo defined. Am I missing it or do we need some exposition on what exactly I would find at `group-alpha.example.com`?

enhancements/common/krm-function-manager.md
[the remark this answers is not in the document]
djzager: I wonder if it would make more sense if it was a declaration, I want to know about updates to `alpha-foo-er` function.

enhancements/common/krm-function-manager.md
```diff
@@ -0,0 +1,321 @@
...
+Bob wants to check if there are any updates to the functions he has installed.
+
+```
+$ kaffine list
+KIND      NAME          DESCRIPTION           INSTALLED
+function  alpha-foo-er  A function that foos  v1.0.0
+function  alpha-bar-er  A function that bars  v0.0.3
+function  alpha-baz-er  A function that bazs  v2.1.0
+
+$ kaffine update --check
```
djzager, line 138: Looks similar to `dnf update-check`, nice. So are the variations something like:

```
# Check for updates
kaffine update --check

# Update a single function
kaffine update alpha-foo-er

# Update all
kaffine update --all
```

enhancements/common/krm-function-manager.md
djzager: Looks like it is later in the document, I think you should consider either:
1. Putting some of the technical description/notes before user stories so we have context when we get here.
2. Adding a glossary

djzager: Later, you describe the repos as having both a uri and a name...but here it is only the uri. Should it be `kaffine config add repo ${name} group-alpha.example.com`?

enhancements/common/krm-function-manager.md
```diff
@@ -0,0 +1,321 @@
...
+function  alpha-foo-er  A function that foos  v2.0.0
+function  alpha-bar-er  A function that bars  v1.0.3
+function  alpha-baz-er  A function that bazs  v2.1.0
+```
+
+### Implementation Details/Notes/Constraints
+
+#### Notes
+
+A repo has 2 main parts: a URI and a name. The URI is the location of the repo. It can be a server somewhere or a path on the local filesystem. The name is a user-friendly name of the repo. If two repos conflict in naming, use the URI instead.
```
djzager, line 162: I think we should add repos the same way we add remotes with git...kaffine could force the names to be unique.

enhancements/common/krm-function-manager.md
```diff
@@ -0,0 +1,321 @@
...
+group-alpha/alpha-baz-er v2.1.0 -> v3.1.0
+Checked for updates without installing.
+
+$ kaffine update alpha-foo-er
+group-alpha/alpha-foo-er v1.0.0 -> v2.0.0
+
+$ kaffine update alpha-bar-er
+group-alpha/alpha-bar-er v0.0.3 -> v1.0.3
+
+$ kaffine list
```
djzager, line 151: Should `kaffine list` show what repo they come from? I'm not certain but figured I should ask.

enhancements/common/krm-function-manager.md
```diff
@@ -0,0 +1,321 @@
...
+#### CLI Commands
+
+**Search**
+
+```
+kaffine search
+kaffine search <function-name>
+```
+
+Searches all repositories for the function with the specified name. If no name is provided, it’s functionally the same as: `kaffine search '*' -r`. If output is piped, `-o yaml` is automatically selected. Otherwise, `-o default` is the default.
```
djzager, line 178: I personally would suggest we avoid detecting (and reacting to) whether `kaffine` is being piped. Users can add `-o yaml` themselves to pipe to follow-up commands.

enhancements/common/krm-function-manager.md
djzager: I was hoping this portion would explain what the structure of a repo is but I'm not seeing it.

djzager (commented):
A few comments. Nice work by the way. I appreciate the detail.

enhancements/common/krm-function-manager.md
shawn-hurley: install is basically writing down the version and place for a KRM function, worth remembering that you can use KRM functions as binaries as well as containers.

enhancements/common/krm-function-manager.md
```diff
@@ -0,0 +1,321 @@
...
+1. What is the exact format of the outputted data?
+2. How could the system be extended to include the management of entire pipelines?
+
+## Summary
+
+This enhancement proposes a new tool, called **Kaffine** to help manage the discovery, installation, removal, and updating of KRM functions. The tool acts as a buffer between the low-level management of different KRM function versions and the high-level usage of them in an application.
+
+## Motivation
+
+Currently, when creating an application that utilizes KRM functions, there is no standardized way to manage the versioning these functions. While it is possible to manage the images of the functions manually, it is incredibly un-ergonomic to do so. The simplest solution to avoid managing versions is to use the `latest` tag when developing the application. Unfortunately, this produces three major undesirable effects:
```
shawn-hurley, line 42: nit: the versioning of these functions.

enhancements/common/krm-function-manager.md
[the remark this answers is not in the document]
shawn-hurley: I think it is worth noting that the goal is not to make a developer do things like replacing tags. we are making a porcelain command to make this easier and more ergonomic.

enhancements/common/krm-function-manager.md
shawn-hurley: raw binaries is a thing already IIUC.

enhancements/common/krm-function-manager.md
```diff
@@ -0,0 +1,321 @@
...
+
+1. **Lack of reproducibility**: A developer using `latest` might start out with `v1` on their system. However, when they go to deploy on a different system, `latest` might pull a higher version with breaking changes.
+1. **Security vulnerabilities**: Many enterprise solutions will simply not tolerate blindly updating to the latest version of a technology before it is proven.
+1. **Dependant on container technology**: While KRM functions utilize container technology for now, if that requirement is ever relaxed to include raw binaries, relying on tags will simply not work. 
+
+The other option is to lock down every instance of the function's usage to a specific version. While this solves some of the issues with `latest`, this method has issues of its own:
+
+1. **Lack of version mobility**: Developers will have to update every single reference to the function to the newer version, causing a refactoring headache.
+1. **Security vulnerabilities (still)**: A newer version of the function may be released with critical updates and bugfixes
+
```
shawn-hurley, line 52: Adding a third one, which is auditability. If this is defined in one place you don't accidentally use two different versions etc.

enhancements/common/krm-function-manager.md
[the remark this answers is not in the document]
shawn-hurley: +1 @JonahSussman I agree on the use here of 1

+1 also on using make-believe stuff just to get the feel of the use by a user. 

I think for technical discussions, lets have that on that part of the enhancement.

shawn-hurley (changes requested):
I think this enhancements motivation and user stories sections are really good and really tells us what/why and how for a user. I think we now have to get into the technical details and make a second specifically for this. We are very much on the right track but need to:
1. We need to see the links or mentions of the krm function catalog kep's.
2. We need to describe how the installation is saved on disk.
3. We need to describe how we save/know about a repo and what information is stored where.


I would also suggest that we consider what we **need** for the UX and what are nice to-haves and potentially move some of them to a follow-up. there are a lot of options for a lot of subcommands I am worried about the timing to complete all of this.

JonahSussman:
@shawn-hurley @djzager I made some changes regarding the structure of how Kaffine gets its information. I agree that the concept of a "repo" was ill-defined and instead swapped it out with another technology proposed by [KEP-2906: Kustomize Function Catalog](https://github.com/kubernetes/enhancements/tree/master/keps/sig-cli/2906-kustomize-function-catalog). If you could review these changes and tell me what you think, I'd greatly appreciate it!

jwmatthews:

PR #74, opened 2022-07-11
Progress reporting for trasnfer-pvc

pranavgaikwad:


enhancements/crane-2.0/progress-reporting/README.md
```diff
@@ -0,0 +1,140 @@
...
+Unlike MTC (where logs were polled at certain intervals by the controller), Crane works on a log stream directly. The working is similar to `kubectl logs` command. This reduces (perhaps, eliminates) loss of log lines and delay. As a result, we can provide transfer progress with much more accuracy, speed and detail as compared to MTC. Additionally, the subcommand works on only one PVC at a time. It makes it easier to work on logs in isolation.
+
+### Goals
+
+> Provide better progress information about ongoing transfer
+    > Details like transfer speed, number of files transferred, total data transerred, completion percentage, errors, list of failed files 
+
+> Separate output and error streams
+
+> Optionally, write structured progress output to a file for automations to consume
```
djzager, line 52: I'm thinking these should be bullet points instead of quotes?

enhancements/crane-2.0/progress-reporting/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+### Separating errors
+
+Since Kubernetes `Stream()` API does not separate std error and output, the rsync logs are printed to stdout. Errors in rsync logs are easy to parse using well known error keywords. It is possible to separate between output and errors. `failedFiles` and `errors` fields will be printed to standard error.
+ 
+### Dumping list of failed files
+
+It is possible that rsync fails to transfer some files. Reasons could be bad file, i/o errors, permission issues, storage timing out, among others. In case of large number of files, having a list of failed files will help diagnose problems faster. A structured output file can also be leveraged by automation for reporting, retries, manual copying etc. Users may also find files that they don't necessarily care about, e.g. lost+found, dead symlinks, device files. To achieve that, the subcommand will add `--output=<file>` option. When enabled, a list of failed filenames will be written to an output file.
+
+### User Stories [optional]
```
djzager, line 127:
```suggestion
### User Stories
```

enhancements/crane-2.0/progress-reporting/README.md
```diff
@@ -0,0 +1,140 @@
...
+		return err
+	}
+	rsyncLogStreamReader := RsyncLogStreamReader{
+		r: reader,
+	}
+    // Copy() copy custom log message to stdout
+    _, err = io.Copy(os.Stdout, rsyncLogStreamReader)
+	if err != nil {
+		return err
+	}
```
djzager, line 84: Would we be writing anything to stderr?

enhancements/crane-2.0/progress-reporting/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+```golang
+func (r RsyncLogStreamReader) Read(b []byte) (n int, err error) {
+	// parse logs and generate our own log output
+	stdout, stderr := r.generateProgressLog(string(buf[:n]))
+	...
+}
+```
+
+This will allow us to generate progress information as and when logs are available on stream instead of processing all logs at a time. This improves performance, and reduces delay in progress updates. The actual progress will look like:
```
djzager, line 95: Do we intend to dump the raw logs somewhere?

enhancements/crane-2.0/progress-reporting/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+### User Stories [optional]
+
+#### Story 1
+
+As a user, I want to see detailed transfer progress in CLI so that I don't have to grok log lines manually.
+
+#### Story 2
+
+As a user, I want to see a list of files that failed transfer so that I can use it to manually copy those files over.
```
djzager, line 135: I do not believe we currently have the idea of creating the tunnel for the user to aide them in such a manual transfer...should we?

djzager (changes requested):
A few nits. I really like this idea. There are several pieces that are not quite lining up for me...
1. We are parsing stderr/stdout, are we putting anything on stderr? This proposal talks about detecting stderr vs stdout but it's not clear if we are writing this stuff to disk somewhere or what.
2. Where are we saving the list of files that failed transfer? Stdout ... some file on disk, it isn't clear.
3. What happens to the original unprocessed logs? Saved to disk optionally, in the `kubectl logs` if the user really wanted them for debugging.

Clarity around :point_up: points would be really helpful.

enhancements/crane-2.0/progress-reporting/README.md
pranavgaikwad: @djzager We currently don't. I dont know whether we should or not. Something to discuss.

enhancements/crane-2.0/progress-reporting/README.md
pranavgaikwad: @djzager thanks for pointing that out. Updated.

pranavgaikwad:
@djzager Updated document to answer your questions. By default, normal progress logs will be printed on stdout, errors on stderr. Errors will include actual error messages returned by rsync along with file names. If a file output is requested, both progress logs and errors will be written to a file in structured format (mostly JSON). I didn't consider the next steps (what to do with that output file) in the scope of this enhancement.

enhancements/crane-2.0/progress-reporting/README.md
jmontleon: Maybe a debug mode where the client pod just sleeps so they can run commands manually and investigate issues on the server side to try and understand why it's falling apart, whether that be corruption, permissions, etc.

jmontleon:
Does this change anything about the way we do things in pipelines when running in cluster? I can't see why, but thought I'd ask.

jmontleon:
Any reason we should (or shouldn't) have a quiet or silent mode that only prints errors?

djzager (approved):
I like the idea of having a `--quiet` mode and structuring the output of failed files. Nice work.

PR #75, opened 2022-07-12
Improve export behavior

jmontleon:


enhancements/crane-2.0/export-group-mapping/README.md
```diff
@@ -0,0 +1,126 @@
...
+- Deal with known situations of cohabitating resources in a way that better
+facilitates moving to newer clusters.
+
+### Non-Goals
+
+- While preferred versions may also come into play with the modernization, this enhancement is focused on contending with resources represented by multiple API groups.
+
+## Proposal
+
+- Export all API groups by enabling the feature in Velero Discovery
```
djzager, line 65: Would this get us the `extensions/v1beta1` and the `apps/v1` in the 3.11 scenario in addition to `authorization.openshift.io` and `authorization.k8s.io` in the rbac case?

jmontleon: Not even just the two, all of the available:
```
Deployment_apps_v1beta1_mediawiki_mariadb.yaml
Deployment_apps_v1beta2_mediawiki_mariadb.yaml
Deployment_apps_v1_mediawiki_mariadb.yaml
Deployment_extensions_v1beta1_mediawiki_mariadb.yaml
```
The preferred version for apps is v1
```
      "preferredVersion": {
        "groupVersion": "apps/v1",
        "version": "v1"
      }
```
We'll probably want to add some logic to check if it's the preferred version and dump just the one, rather than also v1beta1 and v1beta2. I assume this would be true for other APIs as well and could get rather ugly cleaning up.

djzager: yeah. it wasn't clear on the first review that this would get ALL of them...not sure if it's possible to word it in a way to get it through my thick skull but should consider if it's possible :sunglasses:

jmontleon: Mentioning filtering out the non-preferred versions in the update I'm working on now.

enhancements/crane-2.0/export-group-mapping/README.md
```diff
@@ -0,0 +1,124 @@
...
+                namespace = "clusterscoped"
+        }
+-       return strings.Join([]string{obj.GetKind(), namespace, obj.GetName()}, "_") + ".yaml"
++       return strings.Join([]string{obj.GetKind(), obj.GetObjectKind().GroupVersionKind().GroupKind().Group, namespace, obj.GetName()}, "_") + ".yaml"
+ }
+```
+
+- Filter out versions that don't come from the groups preferred version
+
+- White out undesirable duplicate resources from groups deprecated groups, such as `extensions` and `authorization.openshift.io`
```
djzager, line 106: I don't think it's worth another round of review, but what we mean here is transform plugin(s) will be responsible for filtering out/whiting out these other versions.

djzager (approved):
I'm good with this plan as is. One comment only if others request changes to be more explicit about what's happening on `export` and what is happening on `transform`.

PR #77, opened 2022-07-15
Image migration support

djzager:


jmontleon (approved):
Seems sensible to me. My concerns/confusion were addressed in a slack conversation.

enhancements/crane-2.0/image-sync/README.md
```diff
@@ -0,0 +1,214 @@
...
+* Migrate OpenShift Internal Images
+* Re-Implement direct image migrations, `skopeo sync`, or `oc image mirror`.
+    These tools are readily available to handle images.
+
+## Proposal
+
+### Crane `skopeo-sync-gen` subcommand
+
+Introduce a new subcommand to crane, `skopeo-sync-gen`, that will look for and
+evaulate all `export`ed ImageStreams to determine if they images (and the
```
shawn-hurley, line 125: Nit: they/their

enhancements/crane-2.0/image-sync/README.md
```diff
@@ -0,0 +1,214 @@
...
+into consumable user actions.
+
+#### New Pipeline
+
+Currently, the UI import wizard recognizes when there are PeristentVolumeClaims
+(PVCs) to be migrated and asks the user for additional input for handling this
+data. We will update the import wizard to also recognize wihen there are
+ImageStreams to migrate. If the user wishes to migrate the images, they will be
+asked for the public URL of the cluster registry as this can not be reliably
+obtained via `oc registry info --public` (appears to only work on 4.11+ at the
```
shawn-hurley, line 176: On openshift 4, there should be a status field on the ClusterOperator for the registry operator that will give you the URL if exposed.

shawn-hurley (approved):
Except for some little nits, this is a great digest of the problem and how to solve it in a modular way that is extensible for users big +1 from me!

The only note, what does the crane slopes-sync command do if not on openshift?

djzager:
> The only note, what does the crane slopes-sync command do if not on openshift?

I mentioned [here](https://github.com/konveyor/crane-runner/issues/56#issuecomment-1185803340) that I believe we should be good citizens and make this generic so that it can be consumed by others to perform `skopeo sync` in cluster...but the priority is to make sure it works for us in this application.

It isn't immediately obvious how `skopeo-sync` tekton ClusterTask would be useful outside of our usecase, stranger things have happened.

enhancements/crane-2.0/image-sync/README.md
djzager: I guess what we could do is a type of graceful degradation.

1. `oc registry info --public` if this fails try
2. equivalent of `oc get route default-route -n openshift-image-registry --template='{{ .spec.host }}'` to get the route. If this fails try
3. `oc get route docker-registry -n default --template='{{ .spec.host }}'` to get the route (I think this is the 3.11 spot). If this fails, the task fails.

enhancements/crane-2.0/image-sync/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+### Crane Runner `crane-skopeo-sync-gen` ClusterTask
+
+Implement a new `crane-skopeo-sync-gen` ClusterTask that executes the new
+subcommand and saves the output to a file in a
+[Workspace](https://tekton.dev/docs/pipelines/tasks/#emitting-results).
+
+This is covered by [this GitHub Issue](https://github.com/konveyor/crane-runner/issues/55).
+
+### Crane Runner `skopeo-sync` ClusterTask
```
djzager, line 151: I'm wondering if we should call this `crane-skopeo-sync` to avoid any confusion about who introduced this and/or who it is for.

djzager: Calling it `crane-skopeo-sync`

enhancements/crane-2.0/image-sync/README.md
djzager: Updating to drop the user ask, associated tickets have been updated to reflect this.

PR #78, opened 2022-08-17
Proposal for end to end integration tests for forklift

rokkbert:
This is supposed to serve as a basis for discussing how we can achieve automated e2e tests for forklift.

### DIFF
--- a/enhancements/common/integration_tests.md
+++ b/enhancements/common/integration_tests.md
@@ -1,0 +1,175 @@
+---
+title: neat-enhancement-idea
+authors:
+  - "@janedoe"
+reviewers:
+  - TBD
+  - "@alicedoe"
+approvers:
+  - TBD
+  - "@oscardoe"
+creation-date: yyyy-mm-dd
+last-updated: yyyy-mm-dd
+status: provisional|implementable|implemented|deferred|rejected|withdrawn|replaced
+see-also:
+  - "/enhancements/this-other-neat-thing.md"  
+replaces:
+  - "/enhancements/that-less-than-great-idea.md"
+superseded-by:
+  - "/enhancements/our-past-effort.md"
+---
+
+# Neat Enhancement Idea
+
+This is the title of the enhancement. Keep it simple and descriptive. A good
+title can help communicate what the enhancement is and should be considered as
+part of any review.
+
+The YAML `title` should be lowercased and spaces/punctuation should be
+replaced with `-`.
+
+To get started with this template:
+1. **Pick a domain.** Find the appropriate domain to discuss your enhancement.
+1. **Make a copy of this template.** Copy this template into the directory for
+   the domain.
+1. **Fill out the "overview" sections.** This includes the Summary and
+   Motivation sections. These should be easy and explain why the community
+   should desire this enhancement.
+1. **Create a PR.** Assign it to folks with expertise in that domain to help
+   sponsor the process.
+1. **Merge at each milestone.** Merge when the design is able to transition to a
+   new status (provisional, implementable, implemented, etc.). View anything
+   marked as `provisional` as an idea worth exploring in the future, but not
+   accepted as ready to execute. Anything marked as `implementable` should
+   clearly communicate how an enhancement is coded up and delivered. If an
+   enhancement describes a new deployment topology or platform, include a
+   logical description for the deployment, and how it handles the unique aspects
+   of the platform. Aim for single topic PRs to keep discussions focused. If you
+   disagree with what is already in a document, open a new PR with suggested
+   changes.
+
+The `Metadata` section above is intended to support the creation of tooling
+around the enhancement process.
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] User-facing documentation is created
+
+## Open Questions [optional]
+
+This is where to call out areas of the design that require closure before deciding
+to implement the design.  For instance, 
+ > 1. This requires exposing previously private resources which contain sensitive
+  information.  Can we do this? 
+
+## Summary
+
+The `Summary` section is incredibly important for producing high quality
+user-focused documentation such as release notes or a development roadmap. It
+should be possible to collect this information before implementation begins in
+order to avoid requiring implementors to split their attention between writing
+release notes and implementing the feature itself. 
+
+A good summary is probably at least a paragraph in length.
+
+## Motivation
+
+This section is for explicitly listing the motivation, goals and non-goals of
+this proposal. Describe why the change is important and the benefits to users.
+
+### Goals
+
+List the specific goals of the proposal. How will we know that this has succeeded?
+
+### Non-Goals
+
+What is out of scope for this proposal? Listing non-goals helps to focus discussion
+and make progress.
+
+## Proposal
+
+This is where we get down to the nitty gritty of what the proposal actually is.
+
+### User Stories [optional]
+
+Detail the things that people will be able to do if this is implemented.
+Include as much detail as possible so that people can understand the "how" of
+the system. The goal here is to make this feel real for users without getting
+bogged down.
+
+#### Story 1
+
+#### Story 2
+
+### Implementation Details/Notes/Constraints [optional]
+
+What are the caveats to the implementation? What are some important details that
+didn't come across above. Go in to as much detail as necessary here. This might
+be a good place to talk about core concepts and how they relate.
+
+### Security, Risks, and Mitigations
+
+**Carefully think through the security implications for this change**
+
+What are the risks of this proposal and how do we mitigate. Think broadly. How
+will this impact the broader OKD ecosystem? Does this work in a managed services
+environment that has many tenants?
+
+How will security be reviewed and by whom? How will UX be reviewed and by whom?
+
+Consider including folks that also work outside your immediate sub-project.
+
+## Design Details
+
+### Test Plan
+
+**Note:** *Section not required until targeted at a release.*
+
+Consider the following in developing a test plan for this enhancement:
+- Will there be e2e and integration tests, in addition to unit tests?
+- How will it be tested in isolation vs with other components?
+
+No need to outline all of the test cases, just the general strategy. Anything
+that would count as tricky in the implementation and anything particularly
+challenging to test should be called out.
+
+All code is expected to have adequate tests (eventually with coverage
+expectations).
+
+### Upgrade / Downgrade Strategy
+
+If applicable, how will the component be upgraded and downgraded? Make sure this
+is in the test plan.
+
+Consider the following in developing an upgrade/downgrade strategy for this
+enhancement:
+- What changes (in invocations, configurations, API use, etc.) is an existing
+  cluster required to make on upgrade in order to...
+  -  keep previous behavior?
+  - make use of the enhancement?
+
+## Implementation History
+
+Major milestones in the life cycle of a proposal should be tracked in `Implementation
+History`.
+
+## Drawbacks
+
+The idea is to find the best form of an argument why this enhancement should _not_ be implemented.
+
+## Alternatives
+
+Similar to the `Drawbacks` section the `Alternatives` section is used to
+highlight and record other possible approaches to delivering the value proposed
+by an enhancement.
+
+## Infrastructure Needed [optional]
+
+Use this section if you need things from the project. Examples include a new
+subproject, repos requested, github details, and/or testing infrastructure.
+
+Listing these here allows the community to get the process for these resources
+started right away.
--- a/enhancements/common/integration_tests.md
+++ b/enhancements/common/integration_tests.md
@@ -1,55 +1,25 @@
 ---
-title: neat-enhancement-idea
+title: automated integration tests for forklift
 authors:
-  - "@janedoe"
+  - "@rokkbert"
 reviewers:
   - TBD
-  - "@alicedoe"
+  - TBD
 approvers:
   - TBD
-  - "@oscardoe"
-creation-date: yyyy-mm-dd
-last-updated: yyyy-mm-dd
-status: provisional|implementable|implemented|deferred|rejected|withdrawn|replaced
+  - TBD
+creation-date: 2022-08-17
+last-updated: 2022-08-17
+status: provisional
 see-also:
-  - "/enhancements/this-other-neat-thing.md"  
+  -   
 replaces:
-  - "/enhancements/that-less-than-great-idea.md"
+  - 
 superseded-by:
-  - "/enhancements/our-past-effort.md"
+  - 
 ---
 
-# Neat Enhancement Idea
-
-This is the title of the enhancement. Keep it simple and descriptive. A good
-title can help communicate what the enhancement is and should be considered as
-part of any review.
-
-The YAML `title` should be lowercased and spaces/punctuation should be
-replaced with `-`.
-
-To get started with this template:
-1. **Pick a domain.** Find the appropriate domain to discuss your enhancement.
-1. **Make a copy of this template.** Copy this template into the directory for
-   the domain.
-1. **Fill out the "overview" sections.** This includes the Summary and
-   Motivation sections. These should be easy and explain why the community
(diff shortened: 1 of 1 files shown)


enhancements/common/integration_tests.md
```diff
@@ -0,0 +1,188 @@
...
+* verifies that the migration succeeded
+
+Most of these can be done with command line tools but it would be good if the
+basic workflow would also be tested through the web-GUI using something like
+[cypress](https://www.cypress.io/) or [selenium](https://www.selenium.dev/).
+
+To meet the goal we need:
+* ressources and processes to provide the infrastructure
+  * kubernetes cluster to run forklift
+  * kubernetes cluster with kubevirt as migration target
```
ahadas, line 70: not necessarily as a separate cluster, we can deploy both mtv and cnv on the same cluster

enhancements/common/integration_tests.md
```diff
@@ -0,0 +1,188 @@
...
+### User Stories
+
+#### Test Release
+
+Before a release version is tagged the tests are triggered by the package maintainer
+to check for any problems that might have been overlooked.
+
+#### Test Merge
+
+If it's fast and cheap enough the tests can automatically run after a PR was merged.
```
ahadas, line 94: ideally they should run before the PR is merged

enhancements/common/integration_tests.md
```diff
@@ -0,0 +1,188 @@
...
+or [CRC](https://crc.dev/crc/))
+* creates (& modifies/deletes) providers for VMware and Red Hat Virtualization
+* creates (& modifies/deletes) appropriate network- and storage mappings
+* creates (& modifies/deletes) migration plans
+* migrates VMs
+* verifies that the migration succeeded
+
+Most of these can be done with command line tools but it would be good if the
+basic workflow would also be tested through the web-GUI using something like
+[cypress](https://www.cypress.io/) or [selenium](https://www.selenium.dev/).
```
ahadas, line 65: I'd say GUI tests are out of the scope of this effort - the GUI in k8s is just yet another client, and while it should definitely be tested (and @yaacov has some plans for that, I'd suggest syncing with him about tooling and such), it shouldn't be mixed with the tests of the backend. testing from the cli should be enough in this context

enhancements/common/integration_tests.md
```diff
@@ -0,0 +1,188 @@
...
+
+The goal is to have an automated job that builds or collects all forklift components
+and then:
+* deploys them to a cluster (e.g. [minikube](https://minikube.sigs.k8s.io/docs/)
+or [CRC](https://crc.dev/crc/))
+* creates (& modifies/deletes) providers for VMware and Red Hat Virtualization
+* creates (& modifies/deletes) appropriate network- and storage mappings
+* creates (& modifies/deletes) migration plans
+* migrates VMs
+* verifies that the migration succeeded
```
ahadas, line 61: one of the things that influenced kubevirt tests was the need for nested virtualization. so this is an important point - ideally we should run the vms and see that they function well, but the cost might be too high so if we're able to verify it differently (e.g., comparing disk hashes when importing from RHV or running virt-tools that check the integrity of the disks) it could simplify things for us

enhancements/common/integration_tests.md
```diff
@@ -0,0 +1,188 @@
...
+Having such tests, which take considerable effort to execute manually, would help
+to guard against regressions when changes are made or new features are implemented.
+They can also be extended as needed to verify specific bug fixes.
+
+They can serve the secondary purpose of documenting what the expected behaviour of
+the system is.
+
+### Goals
+
+The goal is to have an automated job that builds or collects all forklift components
```
ahadas, line 53: yes, ideally by getting the code from a single repo and building all the required components from it

enhancements/common/integration_tests.md
```diff
@@ -0,0 +1,188 @@
...
+on their own infrastructure to verify any work in progress version they are interested
+in.
+
+### Implementation Details/Notes/Constraints
+
+#### Open Questions
+
+* Can the tests run on GitHub?
+  * On a GitHub-hosted runner or somewhere else?
+* What is the best k8s-variant to test on?
```
ahadas, line 108: not sure what you mean here - plain k8s or okd? I see that kubevirt dropped all the okd lanes, makes sense to do the same - test with the 2-3 latest k8s versions

enhancements/common/integration_tests.md
```diff
@@ -0,0 +1,188 @@
...
+* Can/should the k8s-cluster[s] be instantiated dynamically by the
+test or be persistent somewhere?
+* What tool/framework should be used to write the tests and the setup scripts?
```
ahadas, line 111: we should align with the k8s ecosystem and use golang (and specifically ginkgo and gomega) and better be part of a repo that includes everything (so changes can be posted along with their tests when appropriate)

enhancements/common/integration_tests.md
```diff
@@ -0,0 +1,188 @@
...
+* Where can vSphere run?
+  * How much space does it need?
+  * License?
+  * Access control?
+* Where can oVirt run?
+  * How much space does it need?
+  * Access control?
```
ahadas, line 118: another question is whether we can use shared ovirt/vsphere environments - during warm migration we shut down the vm so one test might affect another test in that case. and so far we used the source environment in read-only mode during cold migrations but that might change when we'll migrate LUN disks..
deploying vsphere per execution would likely be too complex and should be easier with ovirt - but as we expect less changes in ovirt going forward and deploying ovirt per execution increases the likelihood for flakiness, I believe there's no clear cut (the other side of that is that people have not that great experience with using a shared ovirt environment, need to see why)

enhancements/common/integration_tests.md
```diff
@@ -0,0 +1,188 @@
...
+* Which GUI-test tool should be used?
+* How fast can/should the tests be?
+* When do the tests run, manually, automatically, periodically?
+* Where do the forklift packages come from? Are they specified when running
```
ahadas, line 122: they should be build as part of the tests, vendoring the dependencies should minimize the time it requires

ahadas (commented):
thanks for starting this @rokkbert , some initial feedback after a brief view

ahadas:
CC @michalskrivanek @liranr23 @bennyz @bkhizgiy @mnecas @yaacov @sjd78 @rszwajko @nyoxi

enhancements/common/integration_tests.md
```diff
@@ -0,0 +1,188 @@
...
+
+They can serve the secondary purpose of documenting what the expected behaviour of
+the system is.
+
+### Goals
+
+The goal is to have an automated job that builds or collects all forklift components
+and then:
+* deploys them to a cluster (e.g. [minikube](https://minikube.sigs.k8s.io/docs/)
+or [CRC](https://crc.dev/crc/))
```
ahadas, line 56: btw, I wouldn't rule out 'kind' for testing - kubevirt uses it in some of their tests, it should be faster..

enhancements/common/integration_tests.md
yaacov: it's also possible to split the tests, run tier0 tests that are fast and cheap enough on each PR, and the more complex tests can run on QE env that will run more time and will use more resources.

enhancements/common/integration_tests.md
yaacov: Yes, the forklift-ui is currently using [cypress](https://www.cypress.io/) for local e2e tests,( not on each PR ),
if the UI is light enough (e.g. running just the UI + light mock data) it may  be possible to run cypress on each PR:
https://docs.cypress.io/guides/continuous-integration/github-actions#Cypress-GitHub-Action

Note that running the e2e UI tests can also be done "headless" where the test environment simulate the browser pages and buttons without actually spinning up a graphic view.
https://docs.cypress.io/guides/guides/command-line
so running them looks an feel like any other CLI test, (e.g. run "./run-test.sh" )

I agree that tests you want to run on each PR should not test the UI, IMHO you should not want to block a backend feature because it breaks the current UI tests, IMHO it should be possible to have a new feature that is not compatible with current UI, and once it's merged to update the UI to work with it.

On the other hand, when QE tests the whole version, the UI tests must pass using current operator.

enhancements/common/integration_tests.md
rokkbert: I meant that if we start up the environment from the tests it would need to be very small, like CRC or minikube. Not sure about OKD.

enhancements/common/integration_tests.md
ahadas: it is an interesting case that we should keep support but there's a trade off for testing it - on the one hand, it's the more challenging scenario where you can't, e.g., set owner references or post entities due to lack of permissions, but it's expected to be less common and would require more things to configure. if we simplify the tests by 50% and yet cover 80% of the relevant scenario, I wouldn't make it twice as complicated for the other 20%..

enhancements/common/integration_tests.md
ahadas: yeah, or 'kind' or k3s - something that is based on k8s but is fast enough

enhancements/common/integration_tests.md
liranr23, line 61: do we wish to test `achieve` functionality (e.g should clean the importer pod)?

enhancements/common/integration_tests.md
```diff
@@ -0,0 +1,188 @@
...
+Most of these can be done with command line tools but it would be good if the
+basic workflow would also be tested through the web-GUI using something like
+[cypress](https://www.cypress.io/) or [selenium](https://www.selenium.dev/).
+
+To meet the goal we need:
+* ressources and processes to provide the infrastructure
+  * kubernetes cluster to run forklift
+  * kubernetes cluster with kubevirt as migration target
+  * vSphere, with VMs
+  * oVirt, with VMs
```
liranr23, line 72: i think that is most of the "hard work" other than having the cluster to use and migrate.
the VMs should differ and be customized per test / feature check.
That means we will need mock/envs that will be static or, we will have to create VMs from scratch. That a question we need to answer. As well for other setups (such as direct LUN in the future).
Moreover, when using single place we may create bottleneck when executing e2e tests concurrently.

enhancements/common/integration_tests.md
ahadas: you probably meant 'archive' here. the importer pod is created and should be cleaned up by CDI - did you mean the conversion pod?

liranr23: yup, thanks for the correction. i meant both. i think that from what i've seen they are both being in `completed` state but hanging there.

enhancements/common/integration_tests.md
michalskrivanek: +1 on a quick github runner test with kind. For a "proper" one, nightly or so, CRC perhaps? We do want to test full OCP environment with the "right" OLM and such.

enhancements/common/integration_tests.md
michalskrivanek: shared env might be all right. It can be updated/reinstalled/clean up asynchronously e.g. every night or week.
for vmw we will most likely use a shared one anyway or maybe https://github.com/goneri/deploy-vmware-ci ?

enhancements/common/integration_tests.md
michalskrivanek: I think all forklift parts are fairly small and fast enough to build

enhancements/common/integration_tests.md
michalskrivanek: creating bunch of vms is not a problem. I would go with real provider, "empty" VMS with various onfiguration won't take more than a minute or two to create.

enhancements/common/integration_tests.md
rokkbert: I think sharing environments that are not read-only is too risky. Two parallel tests breaking each other with false negatives would be an annoying waste of time.

enhancements/common/integration_tests.md
rokkbert: Certainly small in comparison to running the tests :)

enhancements/common/integration_tests.md
rokkbert: I didn't rule it out on purpose, I just wasn't aware of it.

enhancements/common/integration_tests.md
rokkbert: It might be good to also have at least one VM that is optimized for starting up fast, so we can actually start it after the migration and verify that it runs ok.

enhancements/common/integration_tests.md
ahadas: oh that didn't mean the importer pod is cleaned up properly - it just means that it's an issue out of the scope of forklift it that happens. @liranr23 could you please file an issue to CDI on that?

enhancements/common/integration_tests.md
ahadas: > It might be good to also have at least one VM that is optimized for starting up fast, so we can actually start it after the migration and verify that it runs ok.

I mentioned it [here](https://github.com/konveyor/enhancements/pull/78#discussion_r948444107) - this means we'll need nested virtualization and it would also take longer - not sure it is worth it for quick per-PR tests, maybe for more comprehensive ones?

enhancements/common/integration_tests.md
ahadas: we can also have something in between, e.g., a shared ovirt environment on which each run will provision its own vms first and clean them up at the end..

enhancements/common/integration_tests.md
ahadas: > it's also possible to split the tests, run tier0 tests that are fast and cheap enough on each PR

+1
yeah, besides the separation between upstream (forklift) and downstream (MTV) tests, it may make sense to split the upstream tests to quick ones and more comprehensive ones

>, and the more complex tests can run on QE env that will run more time and will use more resources.

I don't see why to use QE environments - they are internal environments and intended for downstream tests, we better use hardware we can expose outside

enhancements/common/integration_tests.md
liranr23: > we can also have something in between, e.g., a shared ovirt environment on which each run provisions its own vms first and cleans them up at the end..

this might be the approach, the downside of it is when it fails.
let's say you provisioned an environment.
now you create your vms to that test run, something failed (for tests we can make sure to continue into the teardown), but if the teardown fails we have leftovers.
then, when the 2nd run comes in it will fail to run, probably on setting up the vms.

rokkbert: I believe such problems can be avoided with randomized names, i.e. the 2nd run should only look at its own VMs and ignore any it doesn't recognize by name. Or can there be more side effects, which would not be covered by this?

enhancements/common/integration_tests.md
michalskrivanek: moving it to source code doesn't help, if you're building from clean it's just a bigger thing to clone.
building from clean takes 2:25 on my local laptop. That looks entirely acceptable

enhancements/common/integration_tests.md
liranr23: done, bz 2120565

enhancements/common/integration_tests.md
michalskrivanek: IMHO we should state that in the proposal already - two kind of tests
1) per-PR in github action
2) nightly/comprehensive taking longer, running periodically

enhancements/common/integration_tests.md
liranr23: no, i think it will solve it in the cost of recognizing the names, but we can always create prefix name and some generated suffix.

enhancements/common/integration_tests.md
mnecas: Just a suggestion for the tier0 tests: We could use the sqlite database inside the controller inventory controller to mock the inventory data. For example, we could test some large-scale inventories with specific use cases without having the provider for migrating.

rokkbert: Interesting idea, @mnecas , but how much is left to test after that? The validator-service?

jwmatthews:

I believe future VM migration work/questions would be surfaced in https://github.com/kubev2v

PR #82, opened 2022-09-21
feat: add an enhancement for a Konveyor CLI

HarikrishnanBalagopal:
Signed-off-by: Harikrishnan Balagopal <harikrishmenon@gmail.com>

### DIFF
--- a/enhancements/cli/README.md
+++ b/enhancements/cli/README.md
@@ -1,0 +1,141 @@
+---
+title: CLI for Konveyor
+authors:
+  - "@ashokponkumar,@HarikrishnanBalagopal"
+reviewers:
+  - ""
+approvers:
+  - ""
+creation-date: 2022-09-21
+last-updated: 2022-09-21
+status: implementable
+see-also:
+  - "N/A"
+replaces:
+  - "N/A"
+superseded-by:
+  - "N/A"
+---
+
+# CLI for Konveyor
+
+Create a unified CLI for the projects in the Konveyor community.
+
+## Release Signoff Checklist
+
+- [x] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+
+## Summary 
+
+At a high level this tool will 
+- improve discoverability of the various projects in Konveyor and the CLI or UI tools they provide.
+- provide a simple uniform process to install, invoke, update and uninstall multiple versions of all the tools.
+- optionally, we can provide static code analysis and recommendations on what tools to use.
+
+## Motivation
+
+- Currently there are a lot of projects under the Konveyor community. Several of these projects produce tools to help users transform their apps. However, the end-user the experience is still that of interacting with several completely different projects.
+- Recently Konveyor got accepted into CNCF as a Sandbox project. This gives us the opportunity to coalesce all the tools under a single umbrella.
+- A single CLI tool can help give the end-user a seamless experience similar to the ones provided by Kubectl, IBM Cloud CLI, AWS CLI, etc.
+
+## Non-goals
+
+- This tool is not meant to be a replacement for the UI/Hub workflow.
+
+## Proposal
+
+There are three key areas for this tool to work on:
+
+1. First, being one of the main ways users discover the tools. This will require the tools to be made availabe as plugins.
+2. Second, implement a plugin manager to manage the life-cycle of a plugin. This reduces the barrier to entry. Prototype: https://github.com/konveyor/cli
+3. Optional: Static code analysis engine to look at the user code and ask them a few question to decide which tools to recommend for their use case. This can look at stuff like programming languages and frameworks being used, databases being used, whether or not they have Dockerfiles, Kubernetes Yamls, Helm charts, Cloud Foundry yamls, Terraform yamls, etc.
+
+
+This enhancement will be followed by an in-depth dive into what challenges
+are involved in on-boarding new users and helping them test things out locally
+before committing themselves to a particular solution.
+
+
+The CLI also creates an appropriate directory structure to store all the plugins:
+
+```
+ $HOME
+ |- .konveyor
+ |  |- cache.yaml
+ |  |- plugins
+ |  |   |- <plugin-1-name>.yaml
+ |  |   |- <plugin-1-name>
+ |  |   |   └- v1.2.3
+ |  |   |       └- linux-amd64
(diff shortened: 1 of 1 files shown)


ashokponkumar:
@jwmatthews @rromannissen Here is the enhancement we had talked about. Let us know if more info needs to be put in.

jwmatthews (changes requested):
Thank you for capturing your thoughts in this enhancement, this greatly helps the community to understand and discuss.  

I ask that we hold off on moving this to implementation/review, and wait until we have learned more with the Unified Experience and can properly assess the needs we want from a CLI.

The reason I ask to hold on this is because I feel this approach is not yet aligned to the vision of the Unified Experience.  I think the technical elements of this can be iterated and we can find alignment in time, but I don't think now is the time to focus on the CLI.  

I am not convinced that we want to introduce a CLI focused on unifying all the projects in Konveyor community, I think that is unrelated to the vision of creating a Unified Experience for Konveyor....though I understand this is not obvious, I will explain my thoughts below.

My reasoning for suggesting the delay is I think our primary next step in Konveyor is to establish a Unified Experience for the Konveyor project as a whole.   I see the strategy for the Unified Experience being to create a user workflow centered on the Application Inventory delivered by the Hub.    We want the Hub to be the central entity.  We are not looking to form a collection of disparate tools and apply a light level of discovery/install/porcelain on top..we are looking to create a more unified workflow addressing the workflow of [konveyor/methodology](https://github.com/konveyor/methodology).

I see the Unified Experience being focused on:

1. Surface information about applications to a Decision Maker (static code analysis & assessments), storing info in the Application Inventory
2. Aid a Decision Maker to make informed decisions about their modernization/migration needs
3. Help a Decision Maker plan work via Migration Waves
4. Begin to facilitate the 'work' of modernizing/migration

I see 2 personas being involved for the Unified Experience
 - Decision Maker (Architect):   They will ultimately review information Konveyor surfaces and make informed decisions of what work to plan.  Additionally, they will seed projects for 'Developers' to work to handle identified issues.
 - Developer:  A developer will handle the source code changes to satisfy issues identified and allow the application to take advantage of the given target technology


I see the CLI enhancement as proposed would most likely address the 'Developer' persona for 'Step 4'.   i.e. it would help a developer doing the work to discovery tools, install, and use for aiding the work.

I see potential here to consider, but I don't think now is the time to focus on improving the developer persona, as I think the first priority is on the 'Decision Maker' persona.  For the 'Decision Maker' in 'Step 4', they will want to execute tools in a prescribed manner to seed many projects for many developers.  They are less concerned with a CLI tool for discovery and have a greater need for a non-interactive manner to execute tools as scale, likely via integration as an [addon](https://github.com/konveyor/tackle2-addon)

enhancements/cli/README.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Non-goals
+
+- This tool is not meant to be a replacement for the UI/Hub workflow.
+
+## Proposal
+
+There are three key areas for this tool to work on:
+
+1. First, being one of the main ways users discover the tools. This will require the tools to be made availabe as plugins.
```
shawn-hurley, line 51: I don't know how this plugin is built/released and/or what the API contract is between the plugins(do we mean Addons?) and the hub is. I believe that we need to solve these sets of problems before writing porcelain commands. 

The biggest concern would be designing the API's and interaction models based on a CLI that exists before we have the underlying pieces solved.

shawn-hurley (changes requested):
Besides the points that John brings up, we would need to have a much better understanding of:

1. What a plugin is and how does it interact with the hub?
2. What is the released artifact, and how do we release them from a community perspective? How does X plugin get in and Y does not? How or who is hosting the repo for artifacts?
3. How does one upgrade? is there some dependency model that we need to build?


I do think that some of these can be solved after getting some initial thing done, but I think we have too many unanswered questions.

enhancements/cli/README.md
HarikrishnanBalagopal: Thanks for the feedback, I understand that this is a low priority right now while the underlying pieces are being solved.
```
I believe that we need to solve these sets of problems before writing porcelain commands.
```
Not entirely sure what you mean by this, can you please clarify the exact change that is required?

Correct me if I am wrong but I think `porcelain` refers to plumbing commands that produce output in a format that is easily consumable by other tools. https://git-scm.com/book/en/v2/Git-Internals-Plumbing-and-Porcelain
The commands we implemented in the prototype so far are not meant as plumbing commands.
```
konveyor plugin list
konveyor plugin info <plugin-name>
konveyor plugin install <plugin-name>
konveyor plugin uninstall <plugin-name>
```
All of the commands produce log messages that are meant to inform the user who is invoking the command.

shawn-hurley:
I’m sorry, that was a bad choice of words.So what I mean by porcelain is a nice UX around the plumbing we need to build. Hope that helps?On Sep 23, 2022, at 12:02 PM, HarikrishnanBalagopal ***@***.***> wrote:﻿
@HarikrishnanBalagopal commented on this pull request.



In enhancements/cli/README.md:
> +
+## Motivation
+
+- Currently there are a lot of projects under the Konveyor community. Several of these projects produce tools to help users transform their apps. However, the end-user the experience is still that of interacting with several completely different projects.
+- Recently Konveyor got accepted into CNCF as a Sandbox project. This gives us the opportunity to coalesce all the tools under a single umbrella.
+- A single CLI tool can help give the end-user a seamless experience similar to the ones provided by Kubectl, IBM Cloud CLI, AWS CLI, etc.
+
+## Non-goals
+
+- This tool is not meant to be a replacement for the UI/Hub workflow.
+
+## Proposal
+
+There are three key areas for this tool to work on:
+
+1. First, being one of the main ways users discover the tools. This will require the tools to be made availabe as plugins.

Thanks for the feedback, I understand that this is a low priority right now while the underlying pieces are being solved.
I believe that we need to solve these sets of problems before writing porcelain commands.

Not entirely sure what you mean by this, can you please clarify the exact change that is required?
Correct me if I am wrong but I think porcelain refers to plumbing commands that produce output in a format that is easily consumable by other tools. https://git-scm.com/book/en/v2/Git-Internals-Plumbing-and-Porcelain
The commands we implemented in the prototype so far are not meant as plumbing commands.
konveyor plugin list
konveyor plugin info <plugin-name>
konveyor plugin install <plugin-name>
konveyor plugin uninstall <plugin-name>

All of the commands produce log messages that are meant to inform the user who is invoking the command.

PR #83, opened 2022-09-30
Analyzer Addons Reporting Format

pranavgaikwad:


enhancements/tackle/addons-report-format/README.md
```diff
@@ -0,0 +1,142 @@
...
+status: provisional
+see-also:
+  - "N/A"  
+replaces:
+  - "N/A"
+superseded-by:
+  - "N/A"
+---
+
+# Addons Report Format
```
shawn-hurley, line 24: I wonder if we need different types of add-ons. Each with their own defined API for the different phases of the modernization workflow?

enhancements/tackle/addons-report-format/README.md
```diff
@@ -0,0 +1,142 @@
...
+- [ ] User-facing documentation is created
+
+## Summary
+
+Addons will be executed via Tackle Hub for source code analysis, binary analysis and more. The output of addons is crucial. It will be used by Tackle to take further actions such as creating project management tasks, providing recommendations and remediations, among others. Since addons will be implemented for different purposes, they will perform different tasks. It is important for them to have a common output format. A common format will ensure that the addons integrate with Tackle Hub seamlessly. It will establish a contract between addons and the hub.
+
+### Goals
+
+- Propose a common output format for addons with following properties:
+  - human readable
```
shawn-hurley, line 42: Can you explain what this means as a goal? is this an optional thing?

enhancements/tackle/addons-report-format/README.md
```diff
@@ -0,0 +1,142 @@
...
+    // Technology related technology
+    Technology     Technology `json:"technology"`
+    // Incidents list of instances of violation found
+    Incidents      []Incident `json:"incidents"`
+    // Extras reserved for additional data
+    Extras         interface{}
+}
+
+// Category defines category of violation
+type Category int
```
shawn-hurley, line 85: Are we going to define the list of Categories? Or they going to be registerable and more generic? 

If this is the case, I wonder if it should be a string,

shawn-hurley (commented):
I do wonder if we need to discuss the input to analyzers. 

I think certain things, like AppID, etc., and any optional or changes to rules need to be communicated. 

I am missing the mapping, so it is hard to determine where each field is coming from and what that means for the feasibility of producing this output.

I would also like to consider making a schema for the Extras so that it can be verifiable and that a machine or UI can generate a friendly UI for the extra fields. 

I was also hopeful this would make filtering/searching based on these extras something that we can do.

enhancements/tackle/addons-report-format/README.md
djzager: I am in agreement with the idea that analyzers need their own defined API for input parameters and reporting results. While we don't yet have designs for having a defined API how an addon like tackle-test-generator would run, I see value in recognizing that not all addons are the same (require different inputs, have different format for outputs).

enhancements/tackle/addons-report-format/README.md
```diff
@@ -0,0 +1,142 @@
...
+## Proposal
+
+The addons will follow a common format for their output. It can be represented as JSON or YAML. For every addon Tackle will expect some common fields in their output. Based on capabilities of the addon, the addon can also add more fields. The format will thus have some common fields and some extra fields left for additional output of the addon. The additional output can potentially follow OpenAPI spec. 
+
+### Output Format
+
+Every addon will return a list of violations in their output. Each violation will be:
+
+```golang
+type Violation struct {
```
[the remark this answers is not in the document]
djzager: I would think the more meaningful information we could associate with a report, the better -- key word being meaningful. Makes me think of the github action runs [like this one for example](https://github.com/migtools/crane/actions/runs/3061120534) that has a unique id but also lets you know the sha it was run against.

enhancements/tackle/analyzers-report-format/README.md
```diff
@@ -0,0 +1,126 @@
...
+## Release Signoff Checklist
+
+- [x] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] User-facing documentation is created
+
+## Summary
+
+Addons will be executed via Tackle Hub for source code analysis, binary analysis and more. The output of addons is crucial. It will be used by Tackle to take further actions such as creating project management tasks, providing recommendations and remediations, among others. Since addons will be implemented for different purposes, they will perform different tasks. This enhancement only applies to analyzer addons. Since analyzers can be implemented for various different technologies, they all will work differently internally. However, all addons will be working towards a common goal in the modernization workflow. Therefore, it is important for them to have a common output format. A common format will ensure that the addons integrate with Tackle Hub seamlessly. It will establish a contract between addons and the hub.
```
shawn-hurley, line 37: nit: s/between addons/between analyzer addons/g

enhancements/tackle/analyzers-report-format/README.md
```diff
@@ -0,0 +1,126 @@
...
+  - assists Tackle take next steps
+
+### Non-Goals
+
+- Design APIs for serving the output
+- Propose data store to persist output
+
+## Proposal
+
+The analyzer addons will follow a common format for their output. It can be represented as JSON or YAML. Tackle will expect some common fields from every addon. These fields will assist tackle create project management tasks. Based on capabilities of the addon, the addon can also add more fields. The format will thus have some common fields and some extra fields left for additional output of the addon. The additional output can potentially follow OpenAPI spec.
```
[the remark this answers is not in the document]
shawn-hurley: I believe that we should consider removing the bit about open API until we decide where that open API should live, and we should make a full enhancement for it IMO.

enhancements/tackle/analyzers-report-format/README.md
```diff
@@ -0,0 +1,126 @@
...
+    // Extras reserved for additional data
+    Extras         json.RawMessage
+}
+
+// Incident defines instance of a violation
+type Incident struct {
+    // URI defines location in the codebase where violation is found
+    URI           string    `json:"uri"`
+    // StoryPoints defines expected story points for this incident
+    StoryPoints   int       `json:"storyPoints,omitempty"`
```
shawn-hurley, line 86: What happens if folks like t-shirt size pointing here like s,m,l.... etc. Do we believe that this is something to consider?

enhancements/tackle/analyzers-report-format/README.md
```diff
@@ -0,0 +1,126 @@
...
+    // Effort level of effort needed for the fix
+    Effort        string    `json:"effort,omitempty"`
```
shawn-hurley, line 88: Do we have a defined distinction between story points and effort?

enhancements/tackle/analyzers-report-format/README.md
```diff
@@ -0,0 +1,126 @@
...
+    // Description text description about the incident
+    Description   string    `json:"description"`
+    // QuickFix quick remediation if applicable
+    QuickFix      string    `json:"quickFix,omitempty"`
+    // ExternalLinks hyperlinks to external sources of docs, fixes
+    ExternalLinks []string  `json:"externalLinks"`
```
shawn-hurley, line 94: Would a helper text with an external link be a decent idea?

shawn-hurley (changes requested):
Just some thoughts and comments on fields

One last thing that does not need to be added is how version negotiation is determined. is it sent as a header?

enhancements/tackle/analyzers-report-format/README.md
pranavgaikwad: @shawn-hurley I took this field from windup report. I thought it was a text description of level of effort needed to mitigate the violation. But I was wrong. That field in windup report is actually just clone of StoryPoints field in text format. I am getting rid of that.

enhancements/tackle/analyzers-report-format/README.md
shawn-hurley: just asking the question

shawn-hurley: yeah maybe but I am wrong for worrying that an int is too restrictive, in that not every dev teams uses ints for estimating work

enhancements/tackle/analyzers-report-format/README.md
```diff
@@ -0,0 +1,132 @@
...
+  - "N/A"  
+replaces:
+  - "N/A"
+superseded-by:
+  - "N/A"
+---
+
+# Analyzers Report Format
+
+Tackle will run different addons to perform different tasks in the modernization workflow. Analyzers are one example of addons that take a set of rules & source code as input, analyze the source code and retur violations of those rules as output. The output is a violation report that helps Tackle take further actions. This document proposes a common format for the report of analyzer addons.
```
fabianvf, line 26:
```suggestion
Tackle will run different addons to perform different tasks in the modernization workflow. Analyzers are one example of addons that take a set of rules & source code as input, analyze the source code and return violations of those rules as output. The output is a violation report that helps Tackle take further actions. This document proposes a common format for the report of analyzer addons.
```

enhancements/tackle/analyzers-report-format/README.md
```diff
@@ -0,0 +1,132 @@
...
+
+The analyzer addons will follow a common format for their output. It can be represented as JSON or YAML. Tackle will expect some common fields from every addon. These fields will assist tackle create project management tasks. Based on capabilities of the addon, the addon can also add more fields. The format will thus have some common fields and some extra fields left for additional output of the addon. The additional output can potentially follow OpenAPI spec.
+
+### Output Format
+
+Every addon will return a list of violations in their output. Each violation will be:
+
+```golang
+type Violation struct {
+    // AnalysisID id of the analysis that generated this output
```
fabianvf, line 69: It may make more sense for this not to be included as part of our output data format, instead being tracked by the hub itself. Or at least acknowledge that this information must come from the hub and can't be provided by an analyzer addon

enhancements/tackle/analyzers-report-format/README.md
fabianvf: Would it make more sense to try and have effort level/complexity/etc provided as a few separate fields so that storypoints/tshirt sizing etc could be calculated later via user-configured formulae? At some point for these fields you have to provide some kind of concrete value or it's just not that useful, but maybe storypoints are too high-level/opinionated.

enhancements/tackle/analyzers-report-format/README.md
```diff
@@ -0,0 +1,132 @@
...
+    Extras         json.RawMessage
+}
+
+// Incident defines instance of a violation
+type Incident struct {
+    // URI defines location in the codebase where violation is found
+    URI           string    `json:"uri"`
+    // StoryPoints defines expected story points for this incident
+    StoryPoints   int       `json:"storyPoints,omitempty"`
+    // Description text description about the incident
```
fabianvf, line 87: Comment says `Description`, field is called `Message`

enhancements/tackle/analyzers-report-format/README.md
pranavgaikwad: @fabianvf Analysis will happen within the addon. If you have multiple different analyses for the same codebase, how would you differentiate between the output for them? Did you mean the API URL will have it implicitely?

fabianvf: yeah I was thinking the API url could have it implicitly, just because this is the only piece of data that can't come from inside an addon AFAIK. At least we should acknowledge that in the field description

enhancements/tackle/analyzers-report-format/README.md
fabianvf: @pranavgaikwad the reason I think it could make sense to have multiple fields is that story points are usually derived from some combination of approximations of complexity, amount of work, and risk (ie chance that there is additional unknown work to be done post-change). Different teams will determine story points differently using one or more of those metrics, so there could be value in providing those knobs separately and allowing the architect/project manager/etc to determine how they want their story points to be calculated from those more fundamental values.

shawn-hurley: Does this make life harder for the rule authors? Do we think (and I am inclined to lean this way) that a rule author knows the diff between complexity, amount of work, and risk?

enhancements/tackle/analyzers-report-format/README.md
jortel: Seems like this would be the Task ID which is injected by the Hub into the addon (pod) at runtime and thus known to the addon. If we want to correlate Violations/Incidents by analysis, Seems best for the addon to be responsible for populating the entire report. The Hub API won't know the task ID to inject it.
That said, I would imagine that the AnalysisID is something that the addon (adapter) Go binding _could_ inject but it would just be a convenience.

jortel:
The reported data seems reasonable to me but I don't know much about the data or how it will be used.  I don't know the requirements.  But, seems fine from the hub perspective.

enhancements/tackle/analyzers-report-format/README.md
```diff
@@ -0,0 +1,141 @@
...
+  - "N/A"  
+replaces:
+  - "N/A"
+superseded-by:
+  - "N/A"
+---
+
+# Analyzers Report Format
+
+Tackle will run different addons to perform different tasks in the modernization workflow. Analyzers are one example of addons that take a set of rules & source code as input, analyze the source code and return violations of those rules as output. The output is a violation report that helps Tackle take further actions. This document proposes a common format for the report of analyzer addons.
```
jwmatthews, line 26: What do you think about updating this enhancement to refer to Konveyor instead of Tackle?

enhancements/tackle/analyzers-report-format/README.md
shawn-hurley: I would love to have a discussion between UI/Core and us. And potentially even some UX designers if available.

My thought has been that multiple tasks could be executed for an application(s) at one time, all reporting back to be grouped for a given "analysis" run. 

This does seem to be a further thing that the hub and UI would need to know about, but wondering if we have thought about how many "analysis" addons can be tied together for a given point in time run and if this is valuable (I think it could be a good way to organize thoughts). 

The other point, is that eventually we will be making this call to the hub API, the thought right now is that we would report on the task level like `/analysis-addon-report/analysis-id/<id>` or something similar. 

I would like to consider this feature, but I assume this requires us to think about the UX past one analysis addon.

enhancements/tackle/analyzers-report-format/README.md
pranavgaikwad: Updated

enhancements/tackle/analyzers-report-format/README.md
```diff
@@ -0,0 +1,113 @@
...
+  - "N/A"  
+replaces:
+  - "N/A"
+superseded-by:
+  - "N/A"
+---
+
+# Analyzers Report Format
+
+Konveyor will run different addons to perform different tasks in the modernization workflow. Analyzers are one example of addons that take a set of rules & source code as input, analyze the source code and return violations of those rules as output. The output is a violation report that helps Konveyor take further actions such as generating reports, creating project management tasks, etc. This document proposes a format for the output of analyzer addons.
```
shawn-hurley, line 26: nit: source code -> source code or binaries?

enhancements/tackle/analyzers-report-format/README.md
```diff
@@ -0,0 +1,113 @@
...
+	Incidents []Incident `yaml:"incidents"`
+
+	// ExternalLinks hyperlinks to external sources of docs, fixes
+	Links []Link `yaml:"links,omitempty"`
+
+	// Extras reserved for additional data
+	Extras json.RawMessage
+
+	// Effort defines expected story points for this incident
+	Effort *int `yaml:"effort,omitempty"`
```
shawn-hurley, line 78: I thnk this should be on an incident, right?

shawn-hurley (changes requested):
Where are we putting:

1. Rules that ran, but had no matches
2. Rules that had errors
3. the tags that a given ruleset sets? 

I think that we need to add these things to fully understand the apis and contracts that we need to build.

enhancements/tackle/analyzers-report-format/README.md
pranavgaikwad: No since the incidents are just copies of the same Rule, we decided to keep it at Violation level. will update the comment

pranavgaikwad: @shawn-hurley lowest granularity for aggregation is at issue level. we do not envision breaking down the filters further

shawn-hurley: @jortel @mansam I think I agree with Pranav, but want to double check :)

pranavgaikwad:
@shawn-hurley added new sections to talk about tags & errors, please take a look when you can

enhancements/tackle/analyzers-report-format/README.md
```diff
@@ -0,0 +1,172 @@
...
+### Non-Goals
+
+- Design APIs for serving the output
+- Propose data store to persist output
+
+## Proposal
+
+The analyzer addons will follow a common format for their output. It can be represented as JSON or YAML. Konveyor will expect some common fields from every addon. These fields will assist konveyor create project management tasks. Based on capabilities of the addon, the addon can also add more fields. The format will thus have some common fields and some extra fields left for additional output of the addon. The additional output can potentially be OpenAPI spec.
+
+### Output Format
```
jortel, line 52: Please consider the output of the analyzer CLI be a single report document that contains:
- violations
- dependencies
- tags

For example:
```
type Report struct {
    RuleSets []RuleSet
    Dependencies []Dep
    Tags []string
}
```

enhancements/tackle/analyzers-report-format/README.md
```diff
@@ -0,0 +1,172 @@
...
+	Category *Category `yaml:"category,omitempty"`
+
+	// Labels labels on the rule
+	Labels []string `yaml:"labels,omitempty"`
+
+	// Incidents list of instances of violation found
+	Incidents []Incident `yaml:"incidents"`
+
+	// ExternalLinks hyperlinks to external sources of docs, fixes
+	Links []Link `yaml:"links,omitempty"`
```
jortel, line 74: Incident.URL is just the URL but here we have a Link object that contains both a URL and a title. What is the value of the Title?
Would it be better to be more consistent with links?

enhancements/tackle/analyzers-report-format/README.md
shawn-hurley: This could be a large amount of data is there a concern with this?

enhancements/tackle/analyzers-report-format/README.md
pranavgaikwad: @jortel Incident has a file URI. On the other hand, Links contain an actual hyperlink and an additional title field for displaying in the report.

enhancements/tackle/analyzers-report-format/README.md
jortel: I wonder if it would make sense to report effort on both:
Violation.Effort (sum of Incident.Effort)
Incident.Effort (effort for this incident).
Though I many misunderstand how effort is to be calculated.

enhancements/tackle/analyzers-report-format/README.md
jortel: Where does the Title come from?
How does the analyzer know how the UI wants to display it?

enhancements/tackle/analyzers-report-format/README.md
jortel: > @jortel @mansam I think I agree with Pranav, but want to double check :)

Seems reasonable based on my current understanding.

enhancements/tackle/analyzers-report-format/README.md
pranavgaikwad: @jortel Title comes from the Rule. Rule Authors provide them

enhancements/tackle/analyzers-report-format/README.md
```diff
@@ -0,0 +1,172 @@
...
+
+	// Extras reserved for additional data
+	Extras json.RawMessage
+
+	// Effort is a numeric value denoting level of effort needed to fix one occurrence of this issue
+	Effort *int `yaml:"effort,omitempty"`
```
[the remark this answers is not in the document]
pranavgaikwad: @jortel there's a 0 value for certain rules in windup. Some rules might not have the value specified, nil is for those cases

enhancements/tackle/analyzers-report-format/README.md
```diff
@@ -0,0 +1,172 @@
...
+Addons will be executed via Konveyor Hub for source code analysis, binary analysis and more. The output of addons is crucial. It will be used by Konveyor to take further actions such as creating project management tasks, providing recommendations and remediations, among others. Since addons will be implemented for different purposes, they will perform different tasks. This enhancement only applies to analyzer addons. Since analyzers can be implemented for various different technologies, they all will work differently internally. However, all analyzer addons will be working towards a common goal in the modernization workflow. Therefore, it is important for them to have a common output format. A common format will ensure that the analyzer addons integrate with Konveyor Hub seamlessly.
+
+### Goals
+
+- Propose a machine readable output format for analyzer addons
+
+### Non-Goals
+
+- Design APIs for serving the output
+- Propose data store to persist output
```
djzager, line 46: That part I'm missing is how the hub is going to consume this information. Maybe it's out of scope for the enhancement but we should probably mention that here.

Like, is the hub going to read it from the stdout of the pod, or is the addon going to hit some hub endpoint to report the results?

enhancements/tackle/analyzers-report-format/README.md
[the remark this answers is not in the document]
jortel: understood.  but what is the meaningful difference?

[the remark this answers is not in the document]
pranavgaikwad: @jortel On the other hand, Tags are part of Violations right now. Tags are not necessarily "issues". We made effort a pointer to distinguish the 0 value. I think you bring up a good point here. It's best to move Tags out of violations and make Violations 1:1 with issues. That way, we don't have instances where we don't set effort at all. @shawn-hurley wdyt?

enhancements/tackle/analyzers-report-format/README.md
pranavgaikwad: @djzager out of scope for this one, will add as a non-goal

enhancements/tackle/analyzers-report-format/README.md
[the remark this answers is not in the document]
shawn-hurley: There is a meaningful difference between a rule author not setting the effort when I think about the potential category. It is easier to disambiguate now than in the future on upgrade, and I believe it has minimal impact

enhancements/tackle/analyzers-report-format/README.md
jortel: The addon will POST the report to an endpoint (something like): /applications/1/analysis/report body=report.
The API is the integration point for everything. The addon uses the API to fetch whatever it needs from the inventory and uses the API to contribute back to the inventory.

shawn-hurley (approved):
For now the output being on a rule set and it being a list of rules sets makes sense to me. I am an plus one for now

aufi (approved):

PR #85, opened 2022-10-28
Tackle Integration with Jira

rromannissen:
Signed-off-by: Ramon Roman Nissen <rromannissen@gmail.com>


enhancements/tackle/tackle-jira-integration/README.md
```diff
@@ -0,0 +1,350 @@
...
+- Migrators, or the developers in charge of executing the actual changes in the applications, are often part of Third Party factories provided by partners or internal development teams belonging to the organization itself. It is fairly common for organizations to keep track of the work executed by such resources using their own task/issue trackers. This approach would avoid having to log this work twice with two different tools, and might even prevent some development teams from having to learn another additional tool.
+- This integration would enable Tackle to aggregate progress and status data about the overall migration initiative, opening the door for executive dashboards that could be consumed by a potential new senior management persona.
```
shawn-hurley, line 52: nit: Tackle/Konveyor

enhancements/tackle/tackle-jira-integration/README.md
```diff
@@ -0,0 +1,350 @@
...
+
+- Design and implement a generic integration layer with task/issue managers.
+  - Provide a way to create issues/tasks on the external system.
+  - Provide a way to consume information from the issues/tasks on the external system and display it in the Tackle UI.
+- Implement the integration with Atlassian Jira meeting the interface requirements of the integration layer.
+
+
+### Non-Goals
+
+- Implement a Jira addon of any kind.
```
shawn-hurley, line 64: Can you expand on this non-goal? 

Are you saying this integration should not look like an add-on and should be its own thing or something else? 

Can you help me understand why not being an addon is important to you?

enhancements/tackle/tackle-jira-integration/README.md
```diff
@@ -0,0 +1,350 @@
...
+### Personas / Actors
+
+
+#### Administrator
+
+The administrator for the tool that has access to some application-wide configuration parameters that other users can consume but not change or browse. Example: Jira configuration, Jira Credentials.
+
+#### Architect
+
+A technical lead for the migration project that can create and modify applications and information related to it. The Architects don’t need to have access to sensitive information, but can consume it. Example: Create an issue from an application without knowing the credentials of a given Jira instance.
```
shawn-hurley, line 78: Is there a security layer here that we need to make sure is covered? 

The use case is Architect A has access to applications 2 and 3 but not 3. They probably shouldn't be able to access the Jira instance for 3, right?

enhancements/tackle/tackle-jira-integration/README.md
```diff
@@ -0,0 +1,350 @@
...
+- *URL*: String
+- *Instance type*: Dropdown, single selection. Values:
+  - Server
+  - Data Center
+  - Cloud
+- *Credentials*: Dropdown, single selection. Will display a list of all Jira credentials available on the system.
+- *Enable insecure communication*: Toggle.
+
+
+![Jira Configuration - Create Instance](images/admin-jira-config-create.png?raw=true "Jira Configuration - Create Instance")
```
shawn-hurley, line 181: Nit: naming concern


Doesn't Jira call a running thing that you use an instance? could we re-name this to connection or something so that no one thinks we are spinning up a Jira Instance?

enhancements/tackle/tackle-jira-integration/README.md
```diff
@@ -0,0 +1,350 @@
...
+
+
+
+### Security, Risks, and Mitigations
+
+TBD
+
+## Design Details
+
+### Test Plan
```
shawn-hurley, line 328: We have this as an open question I do think we should figure this out before merging.

we should also add to this, how does changes in the integration code when changed, also run through the hub and the hub should run the implementation when it changes. IMO.

shawn-hurley (changes requested):
I have a few more questions

1. Is there a UX for someone not using the UI?
2. How does the hub talk to the integration code? Is this tree to the hub, or can the community contribute these out of tree?

enhancements/tackle/tackle-jira-integration/README.md
rromannissen: Jira has its own type of addons. By this I don't mean not implementing whatever we need to build as a Konveyor addon (that needs to be determined by the engineers), but not building any kind of Atlassian Jira addon ([although the seem to have changed their name to apps](https://marketplace.atlassian.com/addons/app/jira)) to make Jira aware of this integration, at least for the moment. One of the things that we discussed back when this was on the drawing board was the possibility of having a Jira addon that would create projects for Konveyor migration with a given set of issue types, workflows and configuration. That was deemed as overkill on the very first iteration, but could be explored in the future. 

Nevertheless, good catch, we should update the terminology to reflect the new name.

enhancements/tackle/tackle-jira-integration/README.md
rromannissen: Not really, the idea for this is to abstract Architects from using nominal Jira accounts. A service account is provided by the organization, and Architects can use it to access Jira indirectly. Regarding access to application 1 or 2 for an Architect, we don't have the concept of multi tenancy in Tackle. All architects are able to access all applications, so by extension it makes sense they are able to access their associated information in Jira as well, at least indirectly.

enhancements/tackle/tackle-jira-integration/README.md
rromannissen: Yeah, good catch. I'd say "Create Jira Instance Connection" could make more sense. How does that sound?

enhancements/tackle/tackle-jira-integration/README.md
rromannissen: I guess we should decide with @jwmatthews and @jortel if we want to get started replacing the terminology and refer to Tackle as Konveyor from now on.

enhancements/tackle/tackle-jira-integration/README.md
rromannissen: I've changed it and added a link to the Jira App marketplace to make it self explanatory.

enhancements/tackle/tackle-jira-integration/README.md
shawn-hurley: Sounds good for now. I do think that this is a feature that we should add to the backlog then. 

I see it being useful for managed services as well as Active Directory integration.

enhancements/tackle/tackle-jira-integration/README.md
shawn-hurley: +1 was a small concern and it think that would solve it!

shawn-hurley (approved):
Changes sound good, the only concern is testing plan but once that has something I am good!

enhancements/tackle/tackle-jira-integration/README.md
jwmatthews: +1 to replace terminology.   I am trying to do this with how I speak/write but I still slip up :) 

In future we should consider if we want to go further and change repository names as well.

jwmatthews:
LGTM, I read what's captured so far and makes sense.

rromannissen:
@jortel @mansam  we have some new commits from atSistemas to review and discuss.

rromannissen:
@joseacr @tprumbs I think the best approach here would be for you to share details on how to interact with the different flavors and versions of the Jira API. Creating a standalone service would require additional resources for the deployment of Konveyor/Tackle, and the team is considering to include the integration with issue trackers as part of the core. Another interesting topic to contribute to would be to define automated testing strategies when dealing with Jira. We don't know of there is any kind of mock API or a way to spin up Jira instances on demand, but we definitely need something for our automated end to end tests.

joseacr:
> @joseacr @tprumbs I think the best approach here would be for you to share details on how to interact with the different flavors and versions of the Jira API. Creating a standalone service would require additional resources for the deployment of Konveyor/Tackle, and the team is considering to include the integration with issue trackers as part of the core. Another interesting topic to contribute to would be to define automated testing strategies when dealing with Jira. We don't know of there is any kind of mock API or a way to spin up Jira instances on demand, but we definitely need something for our automated end to end tests.

I completely agree @rromannissen. We will bring our functional knowledge about JIRA and its integrations.

mansam (approved):

ibolton336 (approved):

PR #92, opened 2023-02-02
Custom Migration Targets for Konveyor enhancement

rromannissen:
Signed-off-by: rromannissen <rromannissen@gmail.com>

Through interaction with different focus user groups in the field, it has been discovered that configuring custom rules on each analysis run can be cumbersome and doesn't scale well when dealing with third party factories, as it requires them to understand some concepts from the Konveyor domain that could be abstracted away. The purpose of this enhancement is to provide an abstraction layer for these users, hiding away the complexities of custom rules configuration by exposing them as custom migration targets managed by advanced users.


hjochmann:
Support this, for our current customer project this kind of integration could be useful

jortel (approved):
LGTM

enhancements/custom-migration-targets/README.md
```diff
@@ -0,0 +1,325 @@
+---
+title: migration-waves
```
istein1, line 2: title: Custom Migration Targets

PR #94, opened 2023-02-06
Automated Tagging enhancement

### LINKED ISSUE #95: [RFE] Automated tagging of resources with Windup
One of the key assets in Konveyor is its tagging model that allows classifying applications in multiple dimensions. However, the task of tagging applications remains manual, which might not scale well when dealing with a large portfolio. The current enhancement aims at providing a way for addons to automatically tag applications based on the information they are able to surface. The starting point would be to leverage the technology stack information that Windup is able to collect during an analysis and translate that into tags that get automatically added to an application.

rromannissen:
Signed-off-by: rromannissen <rromannissen@gmail.com>

One of the key assets in Konveyor is its tagging model that allows classifying applications in multiple dimensions. However, the task of tagging applications remains manual, which might not scale well when dealing with a large portfolio. The current enhancement aims at providing a way for addons to automatically tag applications based on the information they are able to surface. The starting point would be to leverage the technology stack information that Windup is able to collect during an analysis and translate that into tags that get automatically added to an application.

Fixes #95


shawn-hurley:
Should this be its process and not explicitly tied to an analysis?

jortel:
I think we'll want to add a option (checkbox) to the analysis wizard to indicate the user wants the analysis addon to _tag_ the application.  Default: true.

rromannissen:
> I think we'll want to add a option (checkbox) to the analysis wizard to indicate the user wants the analysis addon to _tag_ the application. Default: true.

@jortel that is already included under the "Changes in the Analysis Configuration Wizard" section from the functional specification: "The analysis configuration wizard should include an additional option to allow users to configure whether they want the analysis to automatically tag the included applications or not. This will be displayed as an additional switch in the advanced options view from the wizard. Mockups for this are WIP and will be added as soon as they are ready."

rromannissen:
> Should this be its own process and not explicitly tied to an analysis?

@shawn-hurley that's the spirit of the enhancement: build something generic that any addon can leverage in the future, but deliver value on the first iteration with the current analysis addon AKA Windup.

rromannissen:
@PhilipCattanach we will need a list of all the tags available in Windup and their corresponding category. I know the list is available in the Red Hat downstream documentation here: https://access.redhat.com/documentation/en-us/migration_toolkit_for_applications/6.0/html/cli_guide/reference_material#tech-tags_cli-guide My guess is that if you were able to extract that, it shouldn't be hard to extract the categories as well, which would be required to seed Konveyor with that information as tag types(to be renamed to Categories as well) and tags.

jortel (approved):
LGTM

enhancements/autotagging/README.md
```diff
@@ -0,0 +1,172 @@
...
+- The list of technologies and their corresponding categories that Windup can currently identify will be used to seed the tag types/categories and tags to be shipped with Konveyor out of the box.
+- Custom rules can identify additional tags, so the Windup addon should be able to create tags if they are not already present in the system.
+- The source for tags coming from a Windup analysis will be "Analysis".
+- If an analysis is run on an application that was analyzed previously, the list of tags associated to the "Analysis" source will be recreated unless the user disables the automated tagging in the advanced options screen from the analysis wizard.
+- When running an analysis in the "Source + Dependencies" mode, the Windup CLI produces output for two differentiated applications, one being related to the source code of the application itself and the other to its dependencies. All technologies from both applications should be created as tags for the application that was analyzed. In other words, technologies coming from both the application source code and its dependencies should be taken into account and tagged accordingly.
```
shawn-hurley, line 140: These types of things need to be discussed from an API perspective with @djzager for the addons spec discussion IMO

shawn-hurley (changes requested):
I would like to wait to merge this. I don't know the API that is being proposed from an analyzer perspective.

enhancements/autotagging/README.md
rromannissen: I don't think these belong to the addon API discussion, as they are more of a specification of how the Windup adapter layer on the addon should work. In other words, these are the specifics of how a given addon works, and don't define the API in any way. In fact, no changes to the API should be required to address this from what I understand, but I would ask @jortel to keep me honest here.

djzager: Now I'm confused. In the summary:

> ...the task of tagging applications remains manual... [this] enhancement aims at providing a way for addons to automatically tag applications...

Are we saying that this enhancement is specifically about Windup tagging applications? Or is this enhancement about automatically executing addons in general that tag applications?

If it's the latter, I think it's fair to suggest that the changes required to support that kind of change would be of interest to addons.

djzager: Offline it seems like it was the former, that this enhancement is specifically about Windup automatically tagging applications (not executing the windup addon or any other addon automatically). If @rromannissen can confirm this understanding, then I agree that there should be no addon API changes required.

My comment for this enhancement would then be to make the distinction more clear in the summary.

jortel: Let's be careful with using the term _auto_ or _automatic_ here. The windup addon will simply parse the JSON output of Windup and use that information to ensure tags/categories exist and are associated with the application.  This can be done using the existing APIs. The only additional thing is UI will need to pass an additional option (based on user input) in the Task.Data to the windup addon to indicate that it should _tag_ the application.

rromannissen: Yeah, for the moment it's about having a way to represent tags coming from multiple sources and allowing addons to tag applications on demand. The automated execution of addons that tag applications can be discussed in future enhancements.

rromannissen: My use of *automatic* here comes from the fact that up until now, if users wants to tag an application, they have to do it manually through the UI. What this enhancement proposes is that, as a first step, after an analysis is completed some tags appear on the analyzed applications *automatically*, meaning with no human intervention aside from the configuration of the analysis parameters.

shawn-hurley (approved):

PR #111, opened 2023-03-16
Dynamic reports enhancement

pranavgaikwad:


enhancements/dynamic-reports/README.md
```diff
@@ -0,0 +1,192 @@
...
+* Description: A text description in markdown format
+* Rule ID: Rule for which the issue was created
+* Category: Helps define severity of issue
+* Effort: A numeric value denoting efforts needed to fix the issue
+* Hyperlinks: External links that provide additional information pertaining to the issue
+* Incidents: Occurrences of the issue across the source code. Each incident will have a file URI and a message that helps show information about that particular incident.
+
+### Dependencies
+
+Analyzers output dependencies for an application with each dependency having following properties:
```
rromannissen, line 59: We might need to add metadata that belongs to the domain of each dependency management system. For example, in Maven dependencies are determined by their GAV coordinates (Group, Artifact, Version, and optionally Packaging). Other systems like npm might work differently, so I imagine we need to distinguish between some core fields that are common to all Dependencies, and others that belong to each dependency management domain. My former OOP self is screaming inheritance here, but I won't get into details! :)

enhancements/dynamic-reports/README.md
```diff
@@ -0,0 +1,192 @@
...
+![components.png](./components.png)
+
+### Issues
+
+An issue is analogous to a _Violation_ created by the analyzer. They are created when analyzer finds a match for a Rule.
+
+Issues have following properties:
+
+* Description: A text description in markdown format
+* Rule ID: Rule for which the issue was created
```
shawn-hurley, line 51: do we need to change this to RuleSet.RuleID?

enhancements/dynamic-reports/README.md
shawn-hurley: +1 on adding group I think every dep system that I can think of has this and should be a common field.

enhancements/dynamic-reports/README.md
```diff
@@ -0,0 +1,198 @@
...
+
+* As a user, I want to be able to see trend of story points of issues belonging to "mandatory" category generated from the history of all analyses.
+
+* As a user, I want to be able to see trend of number of issues in each category from the history of all analyses.
+
+_See [historical data](#historical-data)_
+
+#### Application scope
+
+##### Dependencies
```
shawn-hurley, line 127: Do we want to add the stretch goal of being able to see the whole graph? I assume this would change the data that is captured and reported in someway and I wonder if we need to make sure that we can upgrade to do this when we add it in the future?

enhancements/dynamic-reports/README.md
```diff
@@ -0,0 +1,198 @@
...
+* As a user, I want to be able to see a list of Spring Beans in an application with details including Bean Name and Class.
+
+* As a user, I want to be able to see Hibernate session factories and entities in an application.
+
+
+### Implementation Details/Notes/Constraints
+
+#### Selection of applications
+
+Since there could be thousands of applications in the inventory, displaying all issues / dependencies for all applications in the inventory might not be ideal for user experience. In this enhancement, we are proposing that the users select a subset of applications via a separate filter for which they want to see the list of issues or dependencies. They can also select all applications, but we want that to be the choice of the user.
```
shawn-hurley, line 166: Is this a "may" or "must" select subset of applications?

enhancements/dynamic-reports/README.md
```diff
@@ -0,0 +1,198 @@
...
+
+When finding dependencies between two or more applications, it is implied that the dependencies will only be matched using their names. Note that dependencies have more fields than names such as versions, SHAs etc. 
+
+#### Technology filters
+
+Technologies are modelled as name of the technology along with a range of versions of the technologies, both being string values. In this enhancement, the proposed filter is based only on the name of the technology, and does not filter on version of the technology.
+
+#### Filters on multiple columns
+
+One or more columns can be combined together to form complex searches. When filtering by multiple columns, logical AND between different types of columns is implied. Within the same type of columns, however, selecting multiple values implies a logical OR. For instance, when I select "mandatory" for the category column, and "java" and "golang" for target technology, it is implied that a list of issues with "mandatory" category AND with target technologies "java" OR "golang" will be returned.
```
shawn-hurley, line 182: I wonder if an example URL that says the UI and Hub could comment on for this might be good? 

Are we going to implement something like graphql or will we use regular query strings?

shawn-hurley (commented):
I think this is an excellent start and does a good job of capturing the high-level needs. 

I think I have some questions on specifics if we are planning on coming back and updating this and creating either enhancements or prs of the underlying pieces, that also could make a ton of sense. just let me know and resolve the conversations :)

enhancements/dynamic-reports/README.md
```diff
@@ -0,0 +1,198 @@
...
+
+From analyzers perspective, Name field of a dependency would have fully qualified name for the dependency. It makes the most sense to have a generic "Name" field to accomodate all different kinds of dependencies out there for different technologies. As a result, implementing a filter based on Java package group names implies a full text search within names of dependencies.
+
+#### Current statistics
+
+It is possible that the analyses aren't available for all applications in the inventory. When selecting applications to generate statistics, we are only considering applications for which analyses data are available.
+
+#### Historical data
+
+Trends are generated from high level statistics of all analyses that happened so far. We want to store the computed statistics of every analysis and not the entire output. We want to limit the data points we store. But we are not defining a limit on the period of time for which the data stays. In this enhancement, we are saying all of the data from day 1 is available.
```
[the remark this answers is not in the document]
pranavgaikwad: @shawn-hurley I get what you're trying to say that we shouldn't be losing valuable data on purpose. But I also feel having too much data at user's disposal in the reports wouldn't be ideal from UX point of view. A good middle ground I feel would be - store most recent analysis data in Konveyor along with computed statistics, move raw data of previous runs to a secondary place but users don't see the raw old data in the report. This makes our life easy in getting the first iteration out, while not losing that data in case we decide to do more intelligent things in the future.

enhancements/dynamic-reports/README.md
pranavgaikwad: @shawn-hurley honestly, I don't know. @rromannissen any thoughts?

enhancements/dynamic-reports/README.md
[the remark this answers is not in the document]
shawn-hurley: From a high level that would be fine but does seem like we are agreeing that all data should be kept, just that it may not be available to the user through the UI. This makes sense to me can we more clearly codify that?

enhancements/dynamic-reports/README.md
```diff
@@ -0,0 +1,198 @@
...
+
+* As a user, I want to be able that belong to "mandatory" category broken down by their effort with associated story points for each level of effort.
+
+* As a user, I want to be able to see number of issues generated per Java package.
+
+###### Facts
+
+* As a user, I want to be able to see all tags generated for the application by the analyzer.
+
+* As a user, I want to be able to filter tags created by the analyzer by categories (tag types).
```
jortel, line 151: > As a user, I want to be able to see all tags generated for the application by the analyzer.
> As a user, I want to be able to filter tags created by the analyzer by categories (tag types).
Is this different than existing tag display/filtering?

pranavgaikwad: @jortel no. same as existing tag display

shawn-hurley:
I assume that either this is out of date or we need to merge. I think we need to get this up to date and merge if we are cutting the alpha. 

@pranavgaikwad

pranavgaikwad:
@shawn-hurley not terribly out-of-date. Just need to move some things to future, as we decided to reduce the scope. Will update.

jwmatthews:
@pranavgaikwad @shawn-hurley @jortel do you see further updates to this needed, or can we get an ACK and merge?

pranavgaikwad:
@jwmatthews I dont see any further updates.

shawn-hurley (approved):

PR #128, opened 2023-06-20
dotnet-provider

Chanakya-TS:
Enhancement for .Net provider for analyzer lsp

### DIFF
--- a/enhancements/dotnet-provider/README.md
+++ b/enhancements/dotnet-provider/README.md
@@ -1,0 +1,83 @@
+---
+title: dotnet-provider
+authors:
+  - "@Chanakya-TS"
+reviewers:
+  - ?
+approvers:
+  - ?
+creation-date: 2023-06-20
+last-updated: 2023-06-20
+status: provisional
+see-also:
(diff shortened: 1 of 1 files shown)


enhancements/dotnet-provider/README.md
```diff
@@ -0,0 +1,83 @@
...
+for the analyzer lsp to analyze .Net Framework applications.
+
+### Goals
+
+- Make an external provider for analyzer lsp that uses a pre-existing lsp such as Omnisharp Roslyn.
+- Create a sample set of rules for example applications to show the usage of the external provider.
+
+### Non-Goals
+
+- Make the set of rules that can be used for migration from .Net Framework to .Net Core.
```
djzager, line 53: This is a good non-goal. We should avoid any specific migration/modernization targets.

jwmatthews:
@djzager would you like to keep this open for further changes, or do you think its close to state we can merge?

I see @shawn-hurley had a suggestion to consider, not sure if it's been addressed yet.

djzager:
I think it should stay open so that we can modify/update it if it's something we want to target for release-0.4.

PR #132, opened 2023-06-29
:sparkles: Restructure application facts

pranavgaikwad:


enhancements/analyzer-lsp-app-facts/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+- Propose a new field in rule for explicitely indicating engine to generate additional information in addition to a tag
+
+- Propose a new field in the violations output for this information
+
+### Non-Goals
+
+While the final outcome of this will help generate "application facts" in dynamic reports more easily, I don't want to bring that into discussion here. 
+
+## Proposal
```
shawn-hurley, line 68: If this makes sense to complete this work, I am a huge plus one on the implementation-defined here.

shawn-hurley (changes requested):
I need help understanding the user outcome of this work and the actions that they would want to take.   The reason I think this is important is that we will be asking rule authors to make a choice on when to add these facts and when just tags are fine. For a rule author IMO to make this decision, they are going to need to understand what the expectation of the user to do with this information is. I am currently unaware of how we are involved in documentation or in code review.  I see there are two thoughts that a user can have about finding these things: 
1. There is a tag for something that should not be there, and therefore we would want to fix this. I feel like this still falls into violation.
2. There is a tag for something that should be there and good. I don’t see how this impacts any part of the modernization/migration unless there is something wrong with the config, in which case that should be a violation.

One thing we could consider is that “generate facts” implementation is just the default behavior and a rule author never has to decide, but the same question applies to the users of the output, what are they supposed to do with this information besides awareness of the location of why the tag is applied. If this information is important, and it could be, then I would just argue that it should be the default. 


I do think that this should be addressed in the goals for what the information is useful for.

jortel:
First, I don't have a good feel for what these facts are.  I'm not familiar.

I agree that facts should not be used when tags are a good fit to represent the same thing.  For example a fact of: language=Java or persistence=EJB should be a tag.

However, I have seen examples that suggest that facts generated by windup have complex values that _seem_ like information about the application that goes beyond classification (tags) and doesn't represent an issue to be remedied (violation).

For example: The JPA session factories, configurations and entities depicted [here](https://app.slack.com/client/T027F3GAJ/D04NPBQ2B1C). I don't see how this can/should be represented as tags not does it seem to describe a problem to be solved. It just seems like metadata (facts). But, again, I have no idea how this fact example is useful to the user as part of modernization.

pranavgaikwad:
After our discussion earlier, we are not going to consider application facts, it is descoped.

PR #157, opened 2024-01-23
:sparkles: Propose a test runner for analyzer-lsp rules

### LINKED ISSUE #156: [RFE] Allow rule authors to test / verify rules easily

pranavgaikwad:
Fixes #156

### DIFF
--- a/enhancements/analyzer-rules-testing/README.md
+++ b/enhancements/analyzer-rules-testing/README.md
@@ -21,7 +21,7 @@ status: implementable
 
 As of today, testing an analyzer rule involves creating a provider config file, running analysis using the rule against a sample application, finally verifying the results by inspecting the output file manually. 
 
-This enhancement proposes a new test runner for YAML rules. It will allow write tests for rules declaratively, run analysis using the rules against sample data, verify that the expected output in the test matches the actual output and finally, produce a result detailing passed / failed tests.
+This enhancement proposes a new test runner for YAML rules. It will enable rule authors to write tests for rules declaratively, run analysis using the rules against sample data, verify that the expected output in the test matches the actual output and finally, produce a result detailing passed / failed tests.
 
 ## Motivation
 
@@ -29,7 +29,7 @@ The biggest drawback with testing rules right now is that the output must be ver
 
 Assume that you're a rule author creating rules for a certain technology. You write your first rule, you run it against sample application, manually verify that your rule matched in the `output.yaml` file. You count the number of incidents, check if produced messages are correct. You check for any tags created if applicable. All good! You move onto writing your second rule, you run the rules file again and get an output file. This time, you ensure that the first rule still produces the same output as before in addition to verifying your new rule. In every iteration, you manually verify every rule in the ruleset. Ideally, we want the rule author to write a test first and then iteratively work towards perfecting their rule until it produces the desired output. It's necessary that they can write the test by hand and on top of that, they should be able to run a command to verify the results.
 
-In addition to the UX concern above, there is another important concern. As analyzer-lsp adds more and more providers, contribution to rulesets will increase as a result. For the long-term health of the project, it is essential that any new rules added to rulesets have associated tests and sample data. Manual testing process will only discourage rule authors from submitting tests.
+In addition to the UX concern above, there is another important concern. As analyzer-lsp adds more and more providers, contribution to rulesets will increase. For the long-term health of the project, it is essential that any new rules added to rulesets have associated tests and sample data. Manual testing process will only discourage rule authors from submitting tests.
 
 Lastly, the project has been mainly relying on rules from the Windup project which all have associated test cases. At some point, we will discontinue the Windup shim and move onto the new YAML format for good. What happens to the existing test cases in that case? We need to make sure we can convert the existing tests into this new format and continue to have coverage for those old rules.
 
@@ -65,9 +65,15 @@ tests:
       mode: source-only
       depLabelSelector: "!konveyor.io/source=open-source"
     hasIncidents:
-      count: 10
-      messageContains: "produced message should contain this text in all 10 incidents"
-      codeSnipContains: "removedFunctionCall()"
+      and:
+      - fileURI: "<file/uri/where/expected/incidents/should/be/present>"
+        exactly: 10
+        atLeast: 5
+      - fileURI: "<file/uri/of/expected/incident/with/custom/variables>"
+        locations:
+        - lineNumber: 8
+          messageContains: "incident on line number 8 should contain this"
+          codeSnipContains: "code snippet should contain this text"
     hasTags:
     - Category=tag1
 ```
@@ -81,10 +87,17 @@ tests:
     * analysisParams: Additional _optional_ parameters to configure analysis when running this test. By default, these won't be set.
       * mode: Analysis mode, defaulted by analyzer-lsp to provider default.
       * depLabelSelector: Dependency label selector, defaulted to nil.
-    * hasIncidents: The actual content of the incident which will be verified.
-      * count: Desired count of incidents, test will FAIL when total incidents don't match this number in the output.
-      * messageContains: String to search in the message, test will FAIL when any one or more incidents don't have this string in their message.
-      * codeSnipContains: String to search in the code snip, test will FAIL when one or more incidents don't have this string in their code snippets.
+    * hasIncidents: This defines how incidents should be verified. There could be multiple ways verification can be done for a test case. So this is a nested structure just like how we define _when_ block of a rule. It has exactly one _IncidentCondition_.
+      * _IncidentCondition_ is defined for a _fileURI_. It will apply to a specific file. In a given file, the incidents can be verified in one of several ways:
+        * _Simple Verification_: This simply checks if the given file contains exactly given number of incidents. Exactly one of the following parameters can be defined for this type of verification:
+          * count: there should be exactly this many incidents.
+          * atLeast: there should at least this many incidents, more are OK.
+          * atMost: there should at most this many incidents, less are OK.
+        * _Per incident verification_: This is defined on a per incident basis and takes a list of code locations as input. Each code location will take following parameters:
+          * lineNumber: incident should be found on this line
+          * messageContains: message should contain this string
+          * codeSnipContains: code snippet should contain this string
+      * _and_ and _or_ are two special type of _IncidentCondition_ that can be used to form a complex test.
     * hasTags: A list of tags to verify. The test will FAIL when any one or more tags in the list are not found in the output.
 
 Notice that there could be more than one _testCases_ in a test with different _analysisParams_. This is useful when you want to test the same rule but under different analysis parameters. 
@@ -97,6 +110,8 @@ A testcase passes when all of the following conditions are met:
   - _count_ is equal to the actual count of incidents found in the output
   - _messageContains_ is not defined OR (it is defined AND message is found in all incidents)
   - _codeSnipContains_ is not defined OR (it is defined AND code snip is found in all incidents)
+
+_If and / or are used for incident verification, the result of the top level condition will be evaluated based on results of all children accordingly_
 
 > Having a schema for the tests would be nice. It will allow users to auto-generate the boilerplate.
 
@@ -113,6 +128,8 @@ testConfig:
   - name: java
     location: /path/to/java/
 ```
+
+> When a specific provider config is defined in a test file, that config will take precedance over the ruleset level config
 
 #### Running the tests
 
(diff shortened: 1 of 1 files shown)


enhancements/analyzer-rules-testing/README.md
```diff
@@ -0,0 +1,293 @@
...
+### Proposal
+
+#### Writing a test for a rule
+
+Rule authors should should be able to "declare" a test for a rule in a YAML file:
+
+```yaml
+providers:
+- name: java
+  location: ./java-sample-data/
```
jmle, line 57: Just a thought: it might be a good practice to separate data for the different rulesets into separate folders, so that new data for a particular ruleset doesn't tamper with previous tests. Not sure if it could be enforced in any way... Maybe by enforcing some sort of folder structure for data?

enhancements/analyzer-rules-testing/README.md
```diff
@@ -0,0 +1,293 @@
...
+- _hasIncidents_ is not defined OR all of the following is true:
+  - _count_ is equal to the actual count of incidents found in the output
+  - _messageContains_ is not defined OR (it is defined AND message is found in all incidents)
+  - _codeSnipContains_ is not defined OR (it is defined AND code snip is found in all incidents)
+
+> Having a schema for the tests would be nice. It will allow users to auto-generate the boilerplate.
+
+#### Organizing test files in a ruleset
+
+The YAML file we described in the previous section will be saved in the same directory as its rules file. The name of this test file will be same as of the rules file appended with `_test` suffix. For instance, tests for a rules file `rules.yaml` will be stored in `rules_test.yaml` in the same directory. The analyzer-lsp engine will ignore any files that have `_test.yaml` during rules parsing.
```
jmle, line 105: Related to my other comment, maybe we should specify where and how to organize the data for each ruleset.

enhancements/analyzer-rules-testing/README.md
jmle: I think we should at least set a convention, ie:
```
ruleset
|- rule1.yaml
|- rule2.yaml
|- rule1_test.yaml
|- rule2_test.yaml
|- data
  |- rule1_test
    |- AClass.java
    |- a-dependency-1.2.jar
  |- rule2_test
    |- [...]
```
This would help rule developers keep things organised and homogeneous, and test data isolated among rulesets. WDYT?

pranavgaikwad: I don't feel too strongly about the convention. I think we should be at least putting everything under a `data` directory in a ruleset, certainly nothing outside the ruleset directory. But don't have any strong opinions how to structure it further. since we have a config variable to set any path, we cannot really enforce anything. depends on whatever rule authors think

shawn-hurley: What is the issue with allowing users to set this structure how they want in your mind, @jmle? I assume that this would work for some users; I could also see that for a multi-ruleset thing, maybe you have a single application you use, and everyone shares that.

I do not see a downside to allowing the convention to bubble up from the community. I assume that you have more thought and experience behind this coming from Windup, so I am hoping that you can educate us on the problems that this would solve

jmle: @shawn-hurley it's not super important, but in my experience developing rules for Windup, when test data gets mixed, a change or addition in some data might alter the results of the tests of another ruleset that you're not actively changing. I guess my thinking is also that tests should be as independent and isolated as possible, and in this case data and tests go together.

Say for example that you have in folder `eap7` a `data` folder with all data for the tests of that migration path. If you add an additional class for a test you're developing, that addition may inadvertently change the results of another test in the same folder, so you might need to change that test to accommodate the new results.

As I said it's not super important, so if you guys prefer to leave it to appear organically I'm cool :+1:

enhancements/analyzer-rules-testing/README.md
```diff
@@ -0,0 +1,293 @@
...
+- Discuss specific implementation details of the test runner
+
+### Proposal
+
+#### Writing a test for a rule
+
+Rule authors should should be able to "declare" a test for a rule in a YAML file:
+
+```yaml
+providers:
```
djzager, line 55: I've been trying to get up to speed on all the different fields and their location in the test files. Do you think we need to version these? I'm thinking like GVK `konveyor.io/v1alpha1/ruletest`, in case we need to change the format in the future.

enhancements/analyzer-rules-testing/README.md
```diff
@@ -0,0 +1,293 @@
...
+```sh
+--- rules_test.yaml    1/2 PASSED, 1/2 FAILED
+  - rule-1000          2/2 PASSED
+    - testCase[0]      PASS
+    - testCase[1]      PASS
+  - rule-1001          1/2 PASSED, 1/2 FAILED 
+    - testCase[0]      PASS
+    - testCase[1]      FAIL
+      - Unexpected message string in 8/10 incidents
+        - Log created in ./tmp/analyzer-test-099912/analysis.log
```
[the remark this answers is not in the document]
pranavgaikwad: @shawn-hurley Good point. One idea I can think of to get around this - in analyzer-lsp, we should print all error messages on stderr, the runner can then grab the stderr lines and place them in a separate file or print to output. 

I am not sure how to help debug the provider specific things further.

enhancements/analyzer-rules-testing/README.md
```diff
@@ -0,0 +1,293 @@
...
+> Having a schema for the tests would be nice. It will allow users to auto-generate the boilerplate.
+
+#### Organizing test files in a ruleset
+
+The YAML file we described in the previous section will be saved in the same directory as its rules file. The name of this test file will be same as of the rules file appended with `_test` suffix. For instance, tests for a rules file `rules.yaml` will be stored in `rules_test.yaml` in the same directory. The analyzer-lsp engine will ignore any files that have `_test.yaml` during rules parsing. 
+
+The providers can also be defined at a ruleset level in case users just want to re-use for all tests in a rulest. A new field `testConfig` can be introduced in `ruleset.yaml` file:
+
+```yaml
+name: my-ruleset
```
djzager, line 110: Changes like this make me wonder if we need to be versioning these.

enhancements/analyzer-rules-testing/README.md
```diff
@@ -0,0 +1,293 @@
...
+```
+
+#### Running the tests
+
+We are proposing a new CLI that takes the tests as input. 
+
+To run a single test file `rules_test.yaml`:
+
+```sh
+konveyor-analyzer-test --provider-settings provider_settings.json ./rules_test.yaml
```
djzager, line 124: I understand that the specifics of the runner are out of scope, but I would think being able to do `konveyor-analyzer-test --provider-settings settings.json ./rules/...` would be important.

djzager (commented):
This looks like a very promising start to me. Well detailed. It's hard for me to comment on some of the more technical pieces, like how a specific test for a rule should look, as I haven't spent a ton of time looking at rules. My biggest concerns are around UX and protecting ourselves from changes we may need to make in the future.

enhancements/analyzer-rules-testing/README.md
shawn-hurley: might be a little late, BUT I think it is small enough that we can do something smart in the absence of a version

enhancements/analyzer-rules-testing/README.md
shawn-hurley: > a change or addition in some data might alter the results of the tests of another ruleset that you're not actively changing.

This might be an odd question, but if this happens, wouldn't you want to know that the rule would pick up those changes? Then you either agree with that (assuming it is catching real things that you didn't expect the first time you wrote a test), OR you disagree and need to fix the rule.

I think that I see the value in isolating the data and making the tests independent is one way of doing things, and if we agree that this is the best way, to get the best rules, then I am all good doing it this way.

I am wondering if the case that I made for them to be shared leading to better rules has any merit from your experience, or if it doesn't make any sense :)

enhancements/analyzer-rules-testing/README.md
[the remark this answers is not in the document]
shawn-hurley: I think after our convo, that if we can filter the tests to run a single test case, and we are able to pull out the log file for the given provider/lsp such that it is not lost on container exit, that we should be ok for the first go.

enhancements/analyzer-rules-testing/README.md
pranavgaikwad: This is already covered below

enhancements/analyzer-rules-testing/README.md
```diff
@@ -0,0 +1,293 @@
...
+    hasIncidents:
+      count: 10
+      messageContains: "produced message should contain this text in all 10 incidents"
+      codeSnipContains: "removedFunctionCall()"
+    hasTags:
+    - Category=tag1
+```
+* providers: List of configs needed to setup providers. The runner will take a settings file at runtime and mutate it with values defined here for every run. This can also be defined at ruleset level. Values here will take precedance:
+  * name: Name of the provider to match in _provider_settings.json_
+  * location: Path to sample data for this provider
```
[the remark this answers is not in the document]
pranavgaikwad: @shawn-hurley renamed location to dataPath for more clarity

enhancements/analyzer-rules-testing/README.md
```diff
@@ -0,0 +1,327 @@
...
+
+### Non-Goals
+
+- Discuss specific implementation details of the test runner
+
+### Proposal
+
+#### Writing a test for a rule
+
+Rule authors should should be able to "declare" a test for a rule in a YAML file:
```
jmle, line 52: nit: "should should"

enhancements/analyzer-rules-testing/README.md
```diff
@@ -0,0 +1,327 @@
...
+
+The output is produced per test file aggregating total number of failed / passed rules at the top.
+
+Within a test file, it's further broken down into results of individual rules, with results of each test case for a rule.
+
+For any failed test case, the log and the output location is printed so that users can debug. _We will also grab the provider logs and place them in the output directory so that users can debug the providers themselves._
+
+Finally, a summary line will be printed showing overall statistics.
+
+Additionally a YAML output file will be generated which will be later used by the CI:
```
jmle, line 186: Wondering if we should also place the entire `violations.yaml` for the test run somewhere

jmle (approved):

enhancements/analyzer-rules-testing/README.md
```diff
@@ -0,0 +1,293 @@
...
+type IncidentsTest struct {
+	Count            int     `yaml:"count"`
+	MessageContains  *string `yaml:"messageContains"`
+  CodeSnipContains *string `yaml:"codeSnipContains"`
+}
+```
+
+_Mode_ and _DepLabelSelector_ are nil by default. So when they are not defined in the test, provider / engine defaults will be used.
+
+Also notice that _MessageContains_ and _CodeSnipContains_ checks in the testcase are nil by default. The only required field is _Count_. So at the minimum, the test will check for number of incidents for a rule. If any of these optional fields are defined, the test case will add those additional checks.
```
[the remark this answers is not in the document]
shawn-hurley: Yep I saw that LGTM

shawn-hurley (approved):
Approving assuming we agree on test case filter by name and that we can get agreement on the versioning that @djzager has brought up

enhancements/analyzer-rules-testing/README.md
djzager: > While, I kind of wish (we/I) would have versioned the thing from the start, I think we are in this situation now and have to deal. I don't think that we can have a versioned version in the test and not in the rulesets that exist today.
> 
> IDK thoughts?

I agree that, if one then, both rulesets and tests should be versioned. I also agree that rulesets w/o a version still carry with them an implicit version (although I don't know that it's necessarily `v1`). That indicates to me that this version conversation is outside the scope of this PR but one that we should prioritize for the v0.4.0 release. Namely:
1. What version are we at? `v1alpha1`, `v1beta1`, or `v1`. Any unversioned ruleset would implicitly carry this version.
2. Add that version info where appropriate.

pranavgaikwad: @shawn-hurley introducing v1 right now doesn't sound too bad, especially if we could assume anything without a version is v1.

enhancements/analyzer-rules-testing/README.md
```diff
@@ -0,0 +1,327 @@
...
+        - lineNumber: 8
+          messageContains: "incident on line number 8 should contain this"
+          codeSnipContains: "code snippet should contain this text"
+    hasTags:
+    - Category=tag1
+```
+* providers: List of configs needed to setup providers. The runner will take a settings file at runtime and mutate it with values defined here for every run. This can also be defined at ruleset level. Values here will take precedance:
+  * name: Name of the provider to match in _provider_settings.json_
+  * location: Path to sample data for this provider
+* tests: List of tests, each testing a distinct rule
```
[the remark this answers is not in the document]
pranavgaikwad: +1

enhancements/analyzer-rules-testing/README.md
shawn-hurley: I agree with @djzager lets have a discussion on versioning outside the scope of this pr, because it will effect quite a bit I think.

djzager (approved):

Issue #162, opened 2024-01-29
[RFE] Identifying misbehaving applications in a running cluster


pranavgaikwad:
The ask here is to enable analysis of running applications in an OpenShift / k8s cluster and identify applications that match certain rules. The rules identify problems in the running applications by looking at the runtime picture of them.

PR #163, opened 2024-02-07
Kubernetes provider module for the Konveyor analyzer

mansam:
Fixes https://github.com/konveyor/enhancements/issues/162

### DIFF
--- a/enhancements/analyzer-k8s-provider/README.md
+++ b/enhancements/analyzer-k8s-provider/README.md
@@ -82,7 +82,7 @@ category: optional
 message: "Low replica count"
 when:
   k8s-resource.rego_module:
-    policy: |
+    module: |
       package policy
       import data.lib.konveyor
       import future.keywords
@@ -106,6 +106,42 @@ when:
   k8s-resource.rego_expr:
     collection: deployments
     expression: item.spec.replicas < 2
+---
+ruleID: unmounted-claims
+effort: 1
+category: optional
+message:
+when:
+  k8s-resource.rego_module:
+    module: |
+      package policy
+      import data.lib.konveyor
+      import future.keywords
+
+      pvcs[pvc] {
+        some list in input.namespaces[_]
+        some item in list.items
+        item.kind == "PersistentVolumeClaim"
+        pvc := item
+      }
+      
+      mounted_claims[claim] {
+        some pod in data.lib.konveyor.pods
+        some volume in pod.spec.volumes
+        claim := volume.persistentVolumeClaim.claimName
+      }
+
+      incidents[msg] {
+        some pvc in pvcs
+        claim := pvc.metadata.name
+        not mounted_claims[claim]
+        msg := {
+          "apiVersion": pvc.apiVersion,
+          "namespace": pvc.metadata.namespace,
+          "kind": pvc.kind,
+          "name": pvc.metadata.name,
+        }
+      }
 ```
 
 Following after the initial two capabilities should be a third to inspect image manifests that belong to the application
@@ -122,10 +158,10 @@ that have read access to the relevant resources. The provider must tolerate limi
 
 ### Test Plan
 
-Functionality of the provider itself can be tested with a single-node cluster (seeded with known resources)
-and test rulesets that run each capability. It is of greater importance to ensure that each of the rulesets 
-that the provider runs are thoroughly tested. This will require configuration of resources on the cluster that
-are specific to the rules to be tested.
+Aside from the provider's cluster connection and resource gathering, all the provider's behavior could be tested
+with mock Kubernetes resources and simplified test rulesets that exercise each capability. To permit easy testing of the
+capabilities, and to make it possible for ruleset authors to test their rules, it needs to be possible to drop in an
+alternative k8s client (or resource collection abstraction) that is loaded with known resources.
 
 ## Open Questions
 
--- a/enhancements/analyzer-k8s-provider/README.md
+++ b/enhancements/analyzer-k8s-provider/README.md
@@ -7,7 +7,7 @@ reviewers:
 approvers:
   - TBD
 creation-date: 2024-02-05
-last-updated: 2024-02-05
+last-updated: 2024-03-04
 status: implementable
 ---
 
@@ -22,8 +22,9 @@ status: implementable
 
 ## Summary
 
-Create a Kubernetes provider for the Konveyor analyzer to perform in-cluster analysis of a live application and its environment. Initially this will include collection and analysis
-of resources in the application's home namespace, followed by analysis of image manifests.
+Create a Kubernetes provider for the Konveyor analyzer to perform in-cluster analysis of live applications and their deployment environments. This will include collection and analysis
+of a configurable list of namespace-scoped resources. Discovery and analysis of cluster-scoped resources and image manifests could be explored as a follow-up.
+The provider will conform to the Konveyor analyzer provider gRPC interface.
 
 ## Motivation
 
@@ -68,7 +69,7 @@ as its rules evaluation engine. Rego is expressive, easy to read, and well suite
 Leveraging OPA/Rego allows us to focus on implementing the specifics of cluster analysis and allows rules
 authors to write simple rules that are oriented to the data rather than the form in which it is stored (json, yaml, etc).
 
-The provider will initially surface two rule capabilities: One that is oriented towards ease-of-authorship in exchange for limited expressiveness (`rego.expr`), and one that permits freeform authoring of complex Rego modules (`rego.module`).
+The provider will initially surface two rule capabilities: One that is oriented towards ease-of-authorship in exchange for limited expressiveness (`rego_expr`), and one that permits freeform authoring of complex Rego modules (`rego_module`).
 Ideally, the simpler rule format will strike a balance that enables most rules to be written in it. To facilitate the use of the simpler rule capability,
 we will provide a built-in inventory of Rego rules that can be referred to by user-provided rules.
 
@@ -80,7 +81,7 @@ effort: 1
 category: optional
 message: "Low replica count"
 when:
-  rego.module:
+  k8s-resource.rego_module:
     policy: |
       package policy
       import data.lib.konveyor
@@ -102,7 +103,7 @@ effort: 1
 category: optional
 message: "Low replica count"
 when:
-  rego.expr:
+  k8s-resource.rego_expr:
     collection: deployments
     expression: item.spec.replicas < 2
 ```
@@ -115,6 +116,7 @@ pull the manifests and OPA can be used as the policy engine.
 
 In order to perform live cluster analysis, the provider will need to be furnished with cluster credentials
 that have read access to the relevant resources. The provider must tolerate limited permissions to the extent possible.
+`list` and `get` permission on the resource types to be inspected should be sufficient.
 
 ## Design Details
 
@@ -123,7 +125,13 @@ that have read access to the relevant resources. The provider must tolerate limi
 Functionality of the provider itself can be tested with a single-node cluster (seeded with known resources)
 and test rulesets that run each capability. It is of greater importance to ensure that each of the rulesets 
 that the provider runs are thoroughly tested. This will require configuration of resources on the cluster that
-are specific to the rules to be tested. 
+are specific to the rules to be tested.
+
+## Open Questions
+
+Should image manifest analysis be a capability of this provider? A big benefit of doing it in this provider would be that
+this provider is already able to discover what images are in use, and it would be able to analyze manifests in the context of the workloads using them.
+However, it is unclear how complex accessing the image manifests will be.
 
 ## Implementation History
 
--- a/enhancements/analyzer-k8s-provider/README.md
+++ b/enhancements/analyzer-k8s-provider/README.md
@@ -1,0 +1,130 @@
+---
+title: kubernetes-provider-module
+authors:
+  - "@mansam"
+reviewers:
+  - TBD
+approvers:
+  - TBD
+creation-date: 2024-02-05
+last-updated: 2024-02-05
+status: implementable
+---
+
+# Kubernetes provider module
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] User-facing documentation is created
+
+## Summary
+
+Create a Kubernetes provider for the Konveyor analyzer to perform in-cluster analysis of a live application and its environment. Initially this will include collection and analysis
+of resources in the application's home namespace, followed by analysis of image manifests.
+
+## Motivation
+
+Adding a Kubernetes provider to the Konveyor analyzer extends its utility beyond source/binary analysis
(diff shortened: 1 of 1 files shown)


enhancements/analyzer-k8s-provider/README.md
```diff
@@ -0,0 +1,130 @@
...
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] User-facing documentation is created
+
+## Summary
+
+Create a Kubernetes provider for the Konveyor analyzer to perform in-cluster analysis of a live application and its environment. Initially this will include collection and analysis
```
shawn-hurley, line 25: Analysis of image manifests seems like it maybe should be a separate provider.

I would assume that we would also want to support pulling from registries/searching a repository and other stuff like that. 

I would assume that we would want to pass credentials/proxies and everything for that, while I don't know if the Kubernetes provider. 

For now, I would assume that as a follow on we would solve this use case IMO.

enhancements/analyzer-k8s-provider/README.md
```diff
@@ -0,0 +1,130 @@
...
+of resources in the application's home namespace, followed by analysis of image manifests.
+
+## Motivation
```
shawn-hurley, line 28: Maybe I am an optimist, but is there a desire to scan all applications in all namespaces for the rules?

From the Hub/UI perspective, I understand that it makes sense to focus on the application. Why would we restrict the provider to a single namespace from the CLI or pure analysis engine perspective? Listing across namespaces is trivial so just wondering if we have considered it.

enhancements/analyzer-k8s-provider/README.md
```diff
@@ -0,0 +1,130 @@
...
+migration was successful and that the new deployment is adhering to best practices.
+
+### Goals
+
+* Enable users to perform live analysis of applications hosted in a Kubernetes cluster.
+* Enable users to write Kubernetes analysis rules in a concise, expressive, and easy-to-read format.
+
+### Non-Goals
+
+* Modification of live resources or automatic enforcement of rules.
```
shawn-hurley, line 43: +1 as a non-goal

enhancements/analyzer-k8s-provider/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Proposal
+
+The Kubernetes provider must be able to retrieve resources out of a live cluster, evaluate rules against those
+resources, and return the evaluation results with all the details necessary to act on them (resource coordinates, apiVersions, guidance, etc).
+It needs to be easy to write rules so that domain experts are more willing to contribute the time and effort to do so,
+and the rules language needs to be both expressive and familiar.
+
+### User Stories
```
shawn-hurley, line 52: Is there a 3rd user, a cluster admin, who wants to search for misbehaving applications in their cluster?

enhancements/analyzer-k8s-provider/README.md
```diff
@@ -0,0 +1,130 @@
...
+    expression: item.spec.replicas < 2
+```
+
+Following after the initial two capabilities should be a third to inspect image manifests that belong to the application
+being analyzed. This can likely be implemented by delegating to [Skopeo](https://github.com/containers/skopeo) or [containers/image](https://github.com/containers/image) to
+pull the manifests and OPA can be used as the policy engine.
+
+### Security, Risks, and Mitigations
+
+In order to perform live cluster analysis, the provider will need to be furnished with cluster credentials
```
shawn-hurley, line 116: How are we expecting to take credentials?

enhancements/analyzer-k8s-provider/README.md
```diff
@@ -0,0 +1,130 @@
...
+that have read access to the relevant resources. The provider must tolerate limited permissions to the extent possible.
```
shawn-hurley, line 117: I think we can safely say that we need `list` and `get` on the resources

enhancements/analyzer-k8s-provider/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Design Details
+
+### Test Plan
+
+Functionality of the provider itself can be tested with a single-node cluster (seeded with known resources)
```
shawn-hurley, line 123: Is there any limitation on the current e2e system that would prevent us from creating a new test for this provider?

@djzager

enhancements/analyzer-k8s-provider/README.md
```diff
@@ -0,0 +1,130 @@
...
+Leveraging OPA/Rego allows us to focus on implementing the specifics of cluster analysis and allows rules
+authors to write simple rules that are oriented to the data rather than the form in which it is stored (json, yaml, etc).
+
+The provider will initially surface two rule capabilities: One that is oriented towards ease-of-authorship in exchange for limited expressiveness (`rego.expr`), and one that permits freeform authoring of complex Rego modules (`rego.module`).
+Ideally, the simpler rule format will strike a balance that enables most rules to be written in it. To facilitate the use of the simpler rule capability,
+we will provide a built-in inventory of Rego rules that can be referred to by user-provided rules.
+
+#### Example Rules
+
+```yaml
```
shawn-hurley, line 77: Is there another example of using the policy?

enhancements/analyzer-k8s-provider/README.md
```diff
@@ -0,0 +1,130 @@
...
+details they need to successfully migrate or modernize their applications.
+
+### Implementation Details
+
+The Kubernetes provider will use Rego as its rules language, and will use the Open Policy Agent SDK
+as its rules evaluation engine. Rego is expressive, easy to read, and well suited to analysis of structured data such as Kubernetes resources.
+Leveraging OPA/Rego allows us to focus on implementing the specifics of cluster analysis and allows rules
+authors to write simple rules that are oriented to the data rather than the form in which it is stored (json, yaml, etc).
+
+The provider will initially surface two rule capabilities: One that is oriented towards ease-of-authorship in exchange for limited expressiveness (`rego.expr`), and one that permits freeform authoring of complex Rego modules (`rego.module`).
```
shawn-hurley, line 71: Why would the provider name be`rego` and not k8s or Kubernetes?

shawn-hurley (changes requested):
I have some questions and some things to think about. 

I think this is on the right track, though!

enhancements/analyzer-k8s-provider/README.md
mansam: The provider isn't intended to be named rego, but the capability is. Not sure if that makes more sense.

shawn-hurley: the syntax for the rules is <provider>.<cap> I think that you need to change this to:
kubernetes.rego_module or something similar

enhancements/analyzer-k8s-provider/README.md
mansam: The reason I think this capability is well suited to this provider is that by gathering resources from the cluster we are already discovering what images are in use and what resources they're associated with. As I understand it, Skopeo does not require doing a full pull of the image in order to inspect it. Credentials are a problem to be solved for sure. I'm not sure to what extent we can derive what we need from what's already pulled into the cluster.

Maybe this will turn out to be way more complicated than I'm anticipating, in which case it may make sense to do it as a separate provider that just happens to share a lot of overlapping implementation details with this one.

enhancements/analyzer-k8s-provider/README.md
mansam: That could be, but I'm not sure if the use case is meaningfully different from the first one in practice even though the motivation is different. I'm definitely open to discussing it further.

enhancements/analyzer-k8s-provider/README.md
mansam: What other examples would you like to see?

enhancements/analyzer-k8s-provider/README.md
shawn-hurley: I have a feeling that it will be slightly more complicated, and also usable when not using a Kubernetes cluster, for those reasons I do believe that we should consider moving it out.

With all of this said, I am not opposed to having them share code, but how much is shared in just usage of the library, I personally think it is ok to duplicate some code.

enhancements/analyzer-k8s-provider/README.md
shawn-hurley: Yeah we should add this to the enhancement, that way it is apart of any QE test plan and such, especially from the CLI.

enhancements/analyzer-k8s-provider/README.md
shawn-hurley: Probably one that needs to use the complicated nature of the module; AFAICT, this is just the long version of the shortcut below, or did I miss something?

enhancements/analyzer-k8s-provider/README.md
shawn-hurley: I think that a specific kube config is what I was looking for here; I think it is important because kubeconfig contains not only the user/token but also the coordinates of the cluster API, and we are **not** expecting to take a token or something and create a kubeconfig from the coordinates user/token information.

enhancements/analyzer-k8s-provider/README.md
mansam: That's fair, image analysis certainly has value outside of a cluster and I wasn't thinking about it that way before.

enhancements/analyzer-k8s-provider/README.md
mansam: Yeah, that's the long version of the shortcut for compare/contrast purposes. I'll come up with an additional example that uses the feature more completely.

enhancements/analyzer-k8s-provider/README.md
shawn-hurley: Now that I think about/see it on the demo, unless we expect to provide other capabilities focused on k8s, I would like to know if k8s-rego.cap is a better name for this provider. 

Unless we can think of things that would be added in the future, I worry about limiting the space in which other providers could work in k8s. if we take the whole provider name here.

enhancements/analyzer-k8s-provider/README.md
```diff
@@ -0,0 +1,130 @@
...
+to successfully migrate or modernize it.
+
+#### Story 2
+
+As a domain expert I want to write suites of analysis rules to furnish migrators with the
+details they need to successfully migrate or modernize their applications.
+
+### Implementation Details
+
+The Kubernetes provider will use Rego as its rules language, and will use the Open Policy Agent SDK
```
eemcmullan, line 66: Could possibly mention the use of analyzer-lsp `providerConfig` and `capability` to evaluate rules/resources.

enhancements/analyzer-k8s-provider/README.md
```diff
@@ -0,0 +1,130 @@
...
+as its rules evaluation engine. Rego is expressive, easy to read, and well suited to analysis of structured data such as Kubernetes resources.
+Leveraging OPA/Rego allows us to focus on implementing the specifics of cluster analysis and allows rules
+authors to write simple rules that are oriented to the data rather than the form in which it is stored (json, yaml, etc).
+
+The provider will initially surface two rule capabilities: One that is oriented towards ease-of-authorship in exchange for limited expressiveness (`rego.expr`), and one that permits freeform authoring of complex Rego modules (`rego.module`).
+Ideally, the simpler rule format will strike a balance that enables most rules to be written in it. To facilitate the use of the simpler rule capability,
+we will provide a built-in inventory of Rego rules that can be referred to by user-provided rules.
```
eemcmullan, line 73: @shawn-hurley could this call for a "shim" for rego rules -> our rules? Maybe this isn't worth the trouble if there will only be a limited number of these.

enhancements/analyzer-k8s-provider/README.md
aufi: I think this part is pretty important for real-world clusters usage. IIRC this was an issue with must-gather in previous migration projects Crane/Forklift when it didn't get `clusteradmin` role, but it created own user&namespace&pod for gathering which we'll not do in this this case, so `list/*`, `get/*` should he hopefully enough.

enhancements/analyzer-k8s-provider/README.md
mansam: Renamed to `k8s-resource` for now to better reflect what it does.

enhancements/analyzer-k8s-provider/README.md
mansam: Updated to mention that it will conform to the analyzer-lsp's gRPC provider interface (which would require that).

enhancements/analyzer-k8s-provider/README.md
shawn-hurley: I thought that @mansam had by hand moved over the ones that were written. I don't know if there are other larger `rego` projects that we would want to convert. 

I wonder if we have examples of those projects, I think it would make sense, I don't know them and don't know if it needed. Do you have any thoughts?

enhancements/analyzer-k8s-provider/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+In order to perform live cluster analysis, the provider will need to be furnished with cluster credentials
+that have read access to the relevant resources. The provider must tolerate limited permissions to the extent possible.
+`list` and `get` permission on the resource types to be inspected should be sufficient.
+
+## Design Details
+
+### Test Plan
+
+Functionality of the provider itself can be tested with a single-node cluster (seeded with known resources)
```
shawn-hurley, line 125: Is there a world, where we could have a "fake k8s client" that would allow us to run against a set of resources in a directory?

I think that this could be a follow-on especially when we start integrating the rules into the new yaml rule test that @pranavgaikwad is making.

shawn-hurley (approved):
For the most part, I think we are good to go, I would personally still like a complex example

enhancements/analyzer-k8s-provider/README.md
mansam: Yeah, I think we absolutely should look into doing that.

enhancements/analyzer-k8s-provider/README.md
shawn-hurley: This looks great! thank you!

enhancements/analyzer-k8s-provider/README.md
eemcmullan: If it's a small set of rules then I agree, no need. I'm not aware of any others.

enhancements/analyzer-k8s-provider/README.md
pranavgaikwad: @shawn-hurley +100

pranavgaikwad (approved):

jortel (approved):
Question:
Is there a use case for analyzing resources not deployed in a cluster.  For example, manifests either user defined or  generated in a git repository?
If so perhaps the provider could honor the `Location` instead of (or in addition) to the kubeconfig and namespaces.
This can be a follow up PR.

djzager (approved):
ACK

aufi (approved):

ibragins (approved):
LGTM

PR #165, opened 2024-02-14
Multi-language support for analysis

eemcmullan:


enhancements/multi-language support/README.md
```diff
@@ -0,0 +1,251 @@
...
+    ```
+    
+- As a user, I want to run analysis on my Java and TypeScript application against both default Java and TypeScript rules.
+
+    ```
+    kantra analyze --input=<my_java_ts_app> --output=<output_dir>
+    ```
+    
+    
+- As a user, I want to run analysis on my application that does not have a supported provider.
```
djzager, line 70: How do we handle the case where the user wants to run analysis on a Rust and TypeScript application (assume we have TypeScript provider but not Rust)?

enhancements/multi-language support/README.md
```diff
@@ -0,0 +1,251 @@
...
+### Goals
+
+- Create the ability to discover application languages in kantra to determine providers that are necessary to run analysis.
+- Enable all analyzer-lsp providers to be external (containing their own image and running independently). The analyzer-lsp Java provider will need to be moved out from the core analyzer flow.
+- Write and provide default rulesets for each supported provider.
+
+### Non-Goals
+
+- Further discussion of community providers and rulesets
+
```
djzager, line 42: I think it's fair to clearly state the idea of reaching out to a provider running somewhere else (via URL endpoint for example) and whether or not Kantra will support that use case or not.

djzager (commented):
This is a really good start. Thank you for putting this together.

enhancements/multi-language support/README.md
jmle: +1

enhancements/multi-language support/README.md
shawn-hurley: This is a good point, probably an override? I would think a mapping to to "noop" in the user config for that particular language so they don't have to add a `--override-lang <lang>` or something.

enhancements/multi-language support/README.md
shawn-hurley: I think it is fair to state that as a non-goal for this enhancement

enhancements/multi-language support/README.md
```diff
@@ -0,0 +1,251 @@
...
+apiVersion: v1
+kind: Pod
+metadata:
+  name: kantra
+spec:
+  containers:
+	command:
+	- /usr/local/bin/java-provider
+    args:
+    - "--provider-settings=/opt/settings.json", "--rules=/opt/rulesets/java"
```
shawn-hurley, line 203: Shouldn't the analyzer-lsp be sending the provider settings in the init config call?

shawn-hurley (changes requested):
An example of the pod that also runs the analyzer container would be helpful, right now, it looks like the pod is only running the provider

enhancements/multi-language support/README.md
```diff
@@ -0,0 +1,251 @@
...
+
+In kantra, we can run a podman pod with a container for the analyzer-lsp engine, as well as an additional container for each provider that is found during the initial discovery process.
+
+For each found provider, kantra will dynamically build a `providerSettings[]` config file at runtime. For providers supported by Konveyor, a `providerSettings` config will be kept in kantra for simple use. In the case of community providers, we will also require an additional flag to be set such as `--provider-settings` in which the user can set the appropriate config. 
+
+Alongside the `providerSettings`, a yaml file will also be dynamically created to define the podman pod and each provider container.
+
+```
+apiVersion: v1
+kind: Pod
```
shawn-hurley, line 195: @jortel @mansam I think that this should also match what y'all expect the task to run in the hub. I would think that these should match, with potentially different entry points and the hub probably has one more container (the add-on) (IIUC the structure of the hub correctly). 

Let me know what you think

enhancements/multi-language support/README.md
```diff
@@ -0,0 +1,251 @@
...
+  	hostPort: 8080
+    volumeMounts:
+    - name: java-volume
+      mountPath: /rulesets/java
+  volumes:
+  - name: java-volume
+    emptyDir: {}
+```
+
+If a user wishes to run a community provider, or override a supported provider, they must provide a yaml with the container definition. This can be passed in from a new flag such as `--unsupported-provider`. This configuration will then be added to the podman pod spec during kantra's runtime.
```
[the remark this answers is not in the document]
shawn-hurley: I wonder if we start with image and port, with the note that we could eventually need to add support for more stuff because you may need to mount something, or Env Var and maybe even a specific start command?

enhancements/multi-language support/README.md
```diff
@@ -0,0 +1,251 @@
...
+
+There will be some rules that can be applicable to multiple providers. For example: 
+
+```
+name: os/windows
+description: This is a ruleset for Windows operating system specific rules while migrating
+  to Linux operating system.
+```
+
+These multi-use rulesets can be found in `rulesets/common` and can use labels such as `konveyor.io/provider=go` and `konveyor.io/provider=java` to filter the common rules for relevant providers. For each running provider, their individual rulesets can be evaluated, as well as searching in `rulesets/common` for any rules with these provider labels.
```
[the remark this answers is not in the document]
shawn-hurley: I think the point is, that they are rules, that are specifically developed to be common and should live in a place where java ruleset author can rely on. If that makes sense?

enhancements/multi-language support/README.md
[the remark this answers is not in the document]
eemcmullan: We can do this and continue to let kantra deal with the volume mounts.

enhancements/multi-language support/README.md
```diff
@@ -0,0 +1,251 @@
...
+
+```
+kantra analyze --input=<rust_app> --output=<output_dir> --rules=<custom_rule_dir> --provider-settings=<community_provider_settings>
+    --unsupported-provider=<container_yaml>
+```
+
+### Default Rulesets for Supported Providers
+
+As of now, we have a set of default rules for the Java provider: https://github.com/konveyor/rulesets/
+We will want to add more default rulesets for each provider, such as `rulesets/golang` and `rulesets/java`. For delivery of these rulesets, they will be packaged in a known location in each providers' Dockerfile. In kantra, this/these path(s) will be set to `--rules`, unless this default behavior is disabled by the user.
```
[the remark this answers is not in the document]
eemcmullan: I can make this section less definitive then, and add the thought to 'open questions'.

enhancements/multi-language support/README.md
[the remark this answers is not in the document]
eemcmullan: @pranavgaikwad @shawn-hurley is this something we want to decide here? Or leave it generic for now, and resolve later?

enhancements/multi-language support/README.md
eemcmullan: For this case, I think we would require the user set a provider config for the "unsupported" provider (via a flag) that can be appended to the provider settings we will pass to analyzer-lsp. That will then run in its own container alongside the supported provider container.

enhancements/multi-language support/README.md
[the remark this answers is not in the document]
mansam: If conforming provider images were obligated to accept an environment variable for a port to bind to (something like `ENTRYPOINT provider --port ${PORT}`), then Kantra could just decide a port and further reduce the amount of stuff that needed to be provided by the user.

enhancements/multi-language support/README.md
```diff
@@ -0,0 +1,250 @@
...
+    "Aliases": ["golang"],
+    "Weight": 90.5,
+    "Frameworks": [],
+    "Tools": ["1.19"],
+    "CanBeComponent": true
+  }
+]
+```
+
+We can then complete our discovery process by grabbing each language `Name` to store in a slice for later use. Further details on how kantra will handle each discovered provider will be discussed below in the `kantra changes` section.
```
mansam, line 110: Alizer looks like a great way to discover the languages and frameworks that need to be analyzed. How will the discovered languages be mapped to specific providers? (I'm imagining that provider names and language names are not necessarily an exact match, and there could be multiple providers that support a given language).

enhancements/multi-language support/README.md
mansam: In the equivalent situation in the hub, the addon container is filling the role of the analyzer-lsp container, and the provider containers are alongside as they are here. I mentioned this below, but I think we should probably establish a convention for the provider's entrypoint and how the port should be provided so that any given provider image is started the same way as any other.

enhancements/multi-language support/README.md
```diff
@@ -0,0 +1,250 @@
...
+
+```
+kantra analyze --input=<rust_app> --output=<output_dir> --rules=<custom_rule_dir> --provider-settings=<community_provider_settings>
+    --unsupported-provider=<container_yaml>
+```
+
+### Default Rulesets for Supported Providers
+
+As of now, we have a set of default rules for the Java provider: https://github.com/konveyor/rulesets/
+We will want to add more default rulesets for each provider. For delivery of these rulesets, they will be packaged in a known location in each providers' Dockerfile. In kantra, this/these path(s) will be set to `--rules`, unless this default behavior is disabled by the user.
```
mansam, line 233: Should the concept of default rules be part of the provider contract? Provider writers could build the default rules into the image and have the provider load them without Kantra or any other caller needing to know where they are in the image. This would help reduce the distinction between supported and community providers since we wouldn't have to do any special-case setup of built-in rules for providers we know about, and built-in rules would automatically work for ones we don't. (We could still have a flag in the init config to turn off the default rules if we felt that was necessary.)

enhancements/multi-language support/README.md
shawn-hurley: I feel that *we* would need to own a mapping to known built-in providers.

shawn-hurley (changes requested):
I think we are missing the mapping of the languages to providers as @mansam points out. I think it has to solve a couple of problems
1. It needs to solve the default mapping ("Java" -> Java-provider, "yaml" -> yaml-provider, xml....). 
2. It needs to be overridable
3. It needs to be ok, to turn off a provider for a language, (I have Yaml code but I don't want the yaml provider to be used)
4. It needs to be able to ignore known unknown languages (JavaScript is an example right now).

I think it can be a whole new section

enhancements/multi-language support/README.md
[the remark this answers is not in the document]
eemcmullan: Good point

enhancements/multi-language support/README.md
```diff
@@ -183,15 +183,59 @@ type JavaScriptServiceClient struct {
...
+
+```go
+// used to map default provider to discovery language
+var defaultProvidersFromDiscovery = map[string]string{
+  JavaExternalProvider: "java",
+  PythonExternalProvider: "python"
+}
+```
+
+If a provider mapping is not found, kantra will look for `--unsupported-provider`.
```
shawn-hurley, line 226: I think that for now this will work, in the future we will want to:
1. Enable this to be some config file for the user to set once and leave it, so they don't have to do this every time
2. Once they do that, enable a "strict" option, that will fail the command if providers are not found.

I think that we can do this as a follow on, but I do worry about the UX of this without those quality of life enhancements.

shawn-hurley (approved):
I think that we must come back to how to enable folks on the CLI in the future, but this is a huge step forward and I think it is ready to start implementing

ibragins (approved):
LGTM

istein1:
LGTM

PR #167, opened 2024-02-20
:sparkles: propose .net external provider

djzager:


enhancements/dotnet-provider/README.md
```diff
@@ -0,0 +1,226 @@
...
+	return []provider.Capability{
+		{
+			Name:            "referenced",
+			TemplateContext: openapi3.SchemaRef{},
+		},
+	}
+}
+```
+
+The builtin provider's `xml` capability should make it possible for user's to
```
pranavgaikwad, line 126: I think we are missing text describing dependency capability here. I do feel that we should still have a dependency capability in future so that we can return the full tree as a first class output.

enhancements/dotnet-provider/README.md
```diff
@@ -0,0 +1,226 @@
...
+workaround this missing feature in the short term.
+
+### Service Client
+
+One novel concern that arose when interacting with the two Language Server
```
pranavgaikwad, line 131: Is this something that we can handle in provider.Init()

enhancements/dotnet-provider/README.md
```diff
@@ -0,0 +1,226 @@
...
+		}]
+	}
+]
+```
+
+**NOTE**: While we aren't fully articulating how this would be integrated into Konveyor's user experience, this should help with design considerations for future enhancements around Konveyor's multi-language support.
+
+## Open Questions
+* The Java provider had a `Location` field on it's `referenced` condition (see
+  [here](https://github.com/konveyor/analyzer-lsp/blob/main/provider/internal/java/provider.go#L125-L181)),
```
pranavgaikwad, line 223: I am fine with not implementing if this becomes a hurdle to meet the 0.4 timeline

aufi (approved):
Looks good to me as a first step for .net provider. Thinking if it would be helpful to have some overview on what is needed to have "minimum viable" working .net provider (have some plan on addressing non-goal items like ruleset creation)

enhancements/dotnet-provider/README.md
djzager: That's what I did in the POC https://github.com/konveyor/analyzer-dotnet-provider/pull/7/files#diff-2bb921fc6734aac7cc02498ebf71be40146b0698260e731b24c2ff7bea2a9f39R65

enhancements/dotnet-provider/README.md
djzager: Yeah. I marked the dependency analysis as a non-goal of the enhancement. I see now it's mentioned kind of strangely here...are you suggesting that I expand on this statement to make it clear that 1) we aren't doing dependency analysis now 2)we want to do it in the future?

shawn-hurley: I am a plus on on keeping it a non-goal, and not adding info about the future. Lets wait and see if we need to before committing to it?

djzager: I take that to mean I should just drop the whole bit about the `xml` capability. Is that right?

shawn-hurley: Yeah lets do that, and we can follow up with the dep provider for .Net

Issue #155, opened 2024-01-22
[RFE] Allow running providers selectively in analyzer addon


pranavgaikwad:
### Problem ###

We are anticipating changes in the analyzer-lsp base image. We will split up providers into their individual images. In the enhancement, we mainly want to discuss: 
- How exactly we will run the addon Pod with user selected language providers? An important problem here is sharing the source location. 
- How will the UI filter targets/rulesets appropriate for provider(s) that will be selected.
- How could someone use their own provider image? This is not just overriding an existing provider image, but can we make it so that users can also test with new language providers that are not yet in the UI? 
- How can we be sure that provider containers be terminated when the addon has completed?
- How will credentials be injected into provider containers? Example: maven settings file.
- In future, we might need to run .NET on a windows host remotely.

### Solution ###

See: [Hub Issue](https://github.com/konveyor/tackle2-hub/issues/599) for hub/addon details.

#### Filtering/Selection ####
Targets will be filtered and Providers will be selected by `Language` tag(s).  This will likely depend on the Application Discovery feature.

#### Addon Extensions ####

 An Addon Extension is primarily an optional container definition along with criteria defined by a new CR. Providers will be defined to the hub as an Addon Extension.  The first supported selection criteria will be by application tag matching.

#### Task Pod ####
The task pod will be configured with a _main_ container for the addon and a _sidecar_ container for each of the providers.  All containers will mount an `EmptyDir` volume into /shared for sharing files such as credentials and source code trees.

#### Custom Providers ####
Users may define new providers for testing using the UI by added a new Extension CR that is mapped to a new Language tag.  Then, add/replace the language tag so that extension (provider) is selected. 

Testing with the API is simpler.  The user would add the Extension, and submit an analysis tag using the API that references the extension explicitly.

Links:
- [Hub Issue](https://github.com/konveyor/tackle2-hub/issues/599)

Issue #170, opened 2024-04-09
[RFE] Automatic language and technology discovery for applications


rromannissen:
### What is the problem?

Getting insights about the language and technology stack of an application requires the user to run an analysis on it.

### Why is this a problem?

On large portfolios, having to manually schedule an analysis on all applications to start figuring out their language and technology stack could be cumbersome and somehow counterintuitive. The natural flow would be to first understand the language and technology stack of an application, and based on that, potentially have it aligned with an existing archetype to then run the analysis for the most suitable migration target based on that information. 

### Proposed solution

Automate the process of discovering first the language and then the technology stack for applications in the portfolio. When idle, the hub should automatically create language/technology stack discovery tasks for applications that don't have that information available yet in the portfolio. Certain considerations should be taken into account:

- Language/technology stack discovery should be triggered by certain events: application creation, application import, repository modification, credentials association... For example, if language/technology stack discovery failed on an application because the associated credentials are wrong, the system shouldn't try again until new credentials are provided. 
- Automatic discovery should happen with "spare cycles" available in the resources pool. Users shouldn't have to wait until an automatic discovery task is finished to run things like analysis. User requested tasks should always have preference over tasks executing in the background.
- Language/technology stack discovery should be implicit for application analysis. If for whatever reason one application doesn't have its language/technology stack information available yet, when trying to run an analysis on it, those operations will be executed before the actual analysis is executed.

jortel:
Recommend these fields for tasks to begin with:
- ID
- Kind
- State
- Priority
- Application

sjd78:
Related UI work:
  - https://github.com/konveyor/tackle2-ui/issues/1934

PR #173, opened 2024-04-25
:sparkles: Support multiple languages.

jortel:
Ref: #155

### DIFF
--- a/enhancements/multiple-provider/README.md
+++ b/enhancements/multiple-provider/README.md
@@ -1,0 +1,100 @@
+---
+title: neat-enhancement-idea
+authors:
+  - "jortel@redhat.com"
+reviewers:
+  - TBD
+approvers:
+  - TBD
+creation-date: 2024-04-25
+last-updated: 2024-04-25
+status: provisional|implementable|implemented|deferred|rejected|withdrawn|replaced
+see-also:
+replaces:
+superseded-by:
+---
+
+# Support Multiple Languages
+
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] User-facing documentation is created
+
+## Summary
+
+We are anticipating changes in the analyzer-lsp base image. We will split up providers into their individual images. In the enhancement, we mainly want to discuss: 
+- How exactly we will run the addon Pod with user selected language providers? An important problem here is sharing the source location. 
+- How will the UI filter targets/rulesets appropriate for provider(s) that will be selected.
+- How could someone use their own provider image? This is not just overriding an existing provider image, but can we make it so that users can also test with new language providers that are not yet in the UI? 
+- How can we be sure that provider containers be terminated when the addon has completed?
+- How will credentials be injected into provider containers? Example: maven settings file.
+- In future, we might need to run .NET on a windows host remotely.
+
+## Motivation
+
+Applications may be implemented by languages other than Java.
+
+### Goals
+
+- Determine which language providers to execute based on existing information about the application.
+- User can run analysis without knowing the languages.
+- Targets (rules) may be filtered by language.
+- Targets (rules) may be pre-filtered by the UI when language is known.
+- Language defaults to Java (until automated discovery feature added).
+
+### Non-Goals
+
+## Proposal
+
+#### Filtering/Selection ####
+Targets will be filtered and Providers will be selected by `Language` tag(s).  This will likely depend on the Application Discovery feature.
+
+#### Addon Extensions ####
+
+ An Addon Extension is primarily an optional container definition along with criteria defined by a new CR. Providers will be defined to the hub as an Addon Extension.  The first supported selection criteria will be by application tag matching.
+
+#### Task Pod ####
+The task pod will be configured with a _main_ container for the addon and a _sidecar_ container for each of the providers.  All containers will mount an `EmptyDir` volume into /shared for sharing files such as credentials and source code trees.
+
+#### Custom Providers ####
+Users may define new providers for testing using the UI by added a new Extension CR that is mapped to a new Language tag.  Then, add/replace the language tag so that extension (provider) is selected. 
+
+Testing with the API is simpler.  The user would add the Extension, and submit an analysis tag using the API that references the extension explicitly.
+
+### User Stories
+
+### Implementation Details/Notes/Constraints
+
+Constraints:
+- Must support running the analyzer (addon) and providers in:
+  - same container
+  - separate containers
+- Must support _remote_ providers running outside the cluster.
+- Must ensure all containers are terminated when analysis has completed.
+- Must provide the user with provider container logs.
+- Must provide credentials to through provider settings. Example: maven settings.
+- Must support TCP port assignment which is injected to the provider settings.
+- Must support provider specific settings.
+
+Details: [Hub Issue](https://github.com/konveyor/tackle2-hub/issues/599) for hub/addon details.
+
+### Security, Risks, and Mitigations
+
+## Design Details
+
+See: [Hub Issue](https://github.com/konveyor/tackle2-hub/issues/599) for hub/addon details.
+
+### Test Plan
+
+### Upgrade / Downgrade Strategy
+
+## Implementation History
+
+## Alternatives
+
+## Infrastructure Needed [optional]
+


shawn-hurley:
Can we copy the details over here so that we can comment on specific details

jortel:
> Can we copy the details over here so that we can comment on specific details

done.

enhancements/multiple-provider/README.md
```diff
@@ -0,0 +1,392 @@
...
+selector: tag:Language=Java
+
+Java or XML
+selector: tag:Language=Java || tag:Language=XML
+
+No language tag.
+selector: ! tag:Language=
+
+Java and Linux
+selector:  tag:Language=Java && tag:OperatingSystem=Linux
```
shawn-hurley, line 126: pretty concerned with a new format for selector, the concept of `tag:` does not exist, and does not make it clear what I can do with this. I am worried that we keep running into each other with concepts, and considering that the hub is another way of running the CLI at scale, I believe that we need to make the UX as cohesive as possible between the two.

I would prefer that either we make the selectors specific (for tag, whatever else you are thinking of selcting on) or we just start with selector and assume that it will only ever work off of tags

enhancements/multiple-provider/README.md
```diff
@@ -0,0 +1,392 @@
...
+``` 
+kind: Extension
+spec:
+  addon (string):
+  container (core.v1.Container):
+  selector (string):
+  metadata (object):
+```
+Fields:
+- `extension` - declares compatibility with addons.
```
shawn-hurley, line 140: nit: s/extension/addon/ I assume

enhancements/multiple-provider/README.md
```diff
@@ -0,0 +1,392 @@
...
+- `container` - defines the extension container.
+- `selector` - defines selector for inclusion in the addon task pod.
+- `metadata` - defines unstructured information.
+
+Changed _Addon_ CR (Spec):
```
shawn-hurley, line 145: I can not understand the changes the spec. 

Also if we are changing a k8s API, then we need to make sure that we understand what to do on upgrade with all the fields.

enhancements/multiple-provider/README.md
```diff
@@ -0,0 +1,392 @@
...
+
+#### Task Priority ####
+
+Tasks will be run in order of:
+- priority
+- id
+
+Pods are created with the appropriate _PriorityClass_ to ensure node scheduling matches. Task dependencies will be escalated as needed to prevent priority inversion. When (dep) tasks are already scheduled (Pending), the pod will be recreated referencing the updated priority class.
+
+When the nodes are saturated and task pods are phase=Pending, the task manager may reschedule (delete) task  with lower priority in an effort for higher priority pods to be run by the node scheduler.  PriorityClass preemption should handle this.
```
shawn-hurley, line 317: I am worried about anything that starts to try and reschedule things IIUC. If we are using PriorityClass already then leaving the pods in phase pending should be sufficient unless I am misunderstanding something.

jortel: Trust me, I pushed back on this as well.
I looked pretty hard at PriorityClasses.
A few problems with relying on them.
- they are a cluster resource that affect all pods in all namepaces.  I highly doubt any customer will let our operator install them (and should not).  We would need to convince customers to add the classes we need and define them on the tackle CR.  Customers would need to define a default (if one does not exist) to prevent our task pods from running at hiher priority than controllers.  This by itself seems to be a deal breaker.
- priority classes seem to only represent a preference. I cannot find a claim in the documentation that the node scheduler will postpone scheduling lower priority pods (for which there is a node with available resources) to clear capacity for pending higher priority pods requiring more resources than available.  This produces a priority inversion. In practice (what I observed)  was that in a resource bound cluster, when many _small_ pods are pending, _larger_, higher priority pods will are significantly delayed because it's easier to find small holes than large ones.

enhancements/multiple-provider/README.md
jortel: The incompatible changes are is part of a more holistic approach and better symmetry with extensions.  The additions are needed for dynamic selection of addons we'll almost certainly needed for platform discovery (future).  As for upgrade.  We can include a conversion web hook and version our API but:
- web hooks are a pain.
- the API is at alpha.1
- the only addon CRs are ours which are installed/updated by our operator.

That said, if folks feel strongly, we can include a [conversion web hook](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/). But I really don't think it's worth it.

enhancements/multiple-provider/README.md
jortel: Fixed.

enhancements/multiple-provider/README.md
jortel: I think the systemic use of a generic concept like _selector_ is a good thing and makes the UX more cohesive.  The analysis rules are clearly selected using a _label_ selectors.  Extensions, they are clearly _extension_ selectors (by nature of scope).  The same is true for addons.
I don't see how introducing a completely new term/convention for expressing how addons and extensions are selected makes for a better UX.  IMHO, it seems more confusing.
The tag: and platform: prefix provides for extensibility without changing the API and seems like a minor conceptual addition.

jortel: Also, there is not reason to expect that extensions will only be selected by tags and addons will only be selected by tags.  Given that both will need to be selected for source analysis, platform analysis and TBD.
For source analysis, the addon may be selected by tag.
For platform analysis, the addon will like be selected by platform (kind).
It's reasonable to anticipate the need the same flexibility for different types of discovery tasks.

The selector is a compact and extensible mechanism for supporting this and future needs.

enhancements/multiple-provider/README.md
shawn-hurley: After relooking I missed the `-` and `+` that you add that was not the -/+ from the diff, that is my fault I am sorry :)

If we are just adding fields and not changing fields or removing them, then we just need to handle the sane defaults (for the things that exist). 

I agree webhook should be a last resort and is not necessary unless we are changing a field. 

Field removal, on the other hand, is something that is not something we should do in one release. Instead, we should deprecate and deal with the consequences of them being left and removed in a future release. I know this is a pain, but is probably the right thing to do for our users. I think we need to come up with a plan here. 

I would suggest that we should follow a similar process for our APIs that the hub API is providing as well.

enhancements/multiple-provider/README.md
shawn-hurley: That is good information about priority classes; thanks for explaining that!

I want to push back on the need. IMO, the benefit is a premature optimization for a theoretical problem. From using Kantra to do language discovery, I am seeing times in milliseconds, and if we end up delaying by that time frame, it is hard for me to see the benefit as analysis is in the 10's of minutes, as far as I know. 

To be clear, I would agree that this is a problem that is needed if the delay is in the minutes/10's of minutes. 

Can we do this without the this, and come back if/when this does become an issue. Maybe we do something simple like an option to turn off automatic discovery for customers that do hit this issue as an escape hatch?

enhancements/multiple-provider/README.md
shawn-hurley: > The tag: and platform: prefix provides for extensibility without changing the API and seems like a minor conceptual addition.

I worry that this prefix will be expected in other selectors and will not exist; that is the confusion that I am worried about. This may not make sense as a worry, but I wanted to bring it up. 

I agree that the `<prefix>:` to selectors is a suitable mechanism for extending selectors as we move forward for "things" as you said. 

There are things that I would want to make sure that we have if we do go forward (and I think that the worry above is worth ignoring). 
1. we should have good validation, for if the <prefix> is valid. As you said, in the others it is assumed what it is and they are valid, but when we open it up, people can make mistakes that would be hard to track down (why isn't my thing being selected, OHH i am use language:Java instead of tag:Language=java as an example)
2. we should have a discovery mechanism to know what is valid, and because these selectors are fields in k8s resources, and we want them generic, there is no good way (see csi drivers IMO). Maybe this is a UX that we have to live with, and just make sure that we have good documentation. 

WDYT? does that make sense?

jortel: 1. Definitely. The implementation reports errors for an unknown selector prefix.  And, if the tag: selector references an unknown tag category.
2. Not sure I understand this exactly.  I'd anticipated the extension and addon controllers would validate the k8s 
 references in the CRs.  Is that what you meant?

enhancements/multiple-provider/README.md
jortel: The priority escalation features is simpler and only involves redeploying pods that are still pending.  This seems straight forward and simply ensures that the pods submitted to the node scheduler match task priority.  This is critical when a quota exists in the namespace.

To us, the preemption feature is the controversial one and what we pushed back on for these reasons:
- As you stated, It seems like a premature optimization.
- There is no way to determine which pods should be preempted. The pods blocked the analysis pod are just as likely to be belonging to other operators as our own. Even within our own, it impossible to tell.  In all cases it's a best-effort attempt using a sledge hammer.

For these reasons, I made the preemption feature optional.  Enabled with a setting.  I agree we should consider disabling by default.  Users can enable it should they experience problem described in the use case, they can enabled it.  However, we'll need to convince @rromannissen.

I am not opposed to making automated discovery enabled/disabled via setting as well. Again, we'll need to get with @rromannissen as to the default.

enhancements/multiple-provider/README.md
shawn-hurley: Awesome re: validation, can you add a bit about that in the enhancement and how that works? if it is there, and I missed it I am sorry. 

I was kind of hoping that we started to think of a "documentation" screen. Maybe this is outside the scope of this enhancement. One of the things that we are trying to do in Kantra is exposed via open API what providers give you for rules, and I was hoping that for this, we could do something similar.

enhancements/multiple-provider/README.md
jortel: I share the same philosophy here. However, I am all but certain there are no addon CRs in our namespace besides the (1) we install and manage.  Mainly because there is no way of running other from the UI.  Also, I would argue that an alpha1 version offer no guarantee against this kind of change.  It's only as a matter of practicality that I propose we don't do as your suggesting.
Leaving the resources and image fields in the spec introduces the opportunity for anomolies.  What if the image and resources are defined in both the (old) fields and the new container field?  Seems like we would be creating a problem complexity unnecessarily.

shawn-hurley: As discussed w/ Jeff offline:

Plan A: 
we remove the fields, but turn on strict enforcing. This will enable the users who were using the API that the fields that were removed are no longer valid. 

Plan B:
If there are issues with upgrade/downgrade and removing the fields, we will leave them in place but add validation that makes it so they can't be set, with a deprecation warning, and will hopefully be able to be removed in a future release. 

If anyone has thoughts let us know :)

enhancements/multiple-provider/README.md
shawn-hurley: I think that the outcome here is that we need to track two things:

1. A validation feature in the UI for when selecting the implicit selectors, that you can not use `<prefix>:` 
2. A validation feature in kantra to do the same.

@ibolton336 @sjd78 @pranavgaikwad let me know if you folks see issues with that.

enhancements/multiple-provider/README.md
```diff
@@ -0,0 +1,437 @@
...
+
+### Goals
+
+- Dynamically determine which addon to execute based on facts known about the application in inventory.
+- Dynamically determine which language providers to execute based on facts known about the application in inventory.
+- User can run analysis without knowing the languages.
+- Targets (rules) may be filtered by language.
+- Targets (rules) may be pre-filtered by the UI when language is known.
+- Language defaults to Java (until automated discovery feature added).
+- User initiated task execution (Example: analysis) must be prioritized.
```
djzager, line 59: I do think we need to talk in this enhancement about how we are going to deal with providers that support being run on different platforms (ie. `linux/amd64` and `windows/amd64`).

enhancements/multiple-provider/README.md
```diff
@@ -0,0 +1,437 @@
...
+  container (core.v1.Container):
+  selector (string):
+  metadata (object):
+```
+Fields:
+- `addon` - declares compatibility with addons.
+- `container` - defines the extension container.
+- `selector` - defines selector for inclusion in the addon task pod.
+- `metadata` - defines unstructured information.
+
```
djzager, line 149: Will need a way to expose the platform(s) an extension can run on.

enhancements/multiple-provider/README.md
```diff
@@ -0,0 +1,437 @@
...
+Fields:
+- `kind` - declares compatibility with a _kind_ of task.
+- `container` - defines the addon container.
+- `selector` - defines addon selector for task (kind).
+- `metadata` - defines unstructured information.
+- `schema` - _FUTURE_ defines the Task.Data schema.
+
+For symmetry and completeness, the `metadata` field is added and the separate addon fields for the container are replaced with the entire k8s _Container_.   This is mainly a general improvement while we're changing the CR.
+
```
djzager, line 170: Will need a mechanism to declare the platform/os an addon image can be run on.

djzager (commented):
This is very detailed. One concern that was brought up when we were talking about the .NET support effort was how to properly get the container running on the windows node. Assuming that we have analyzer-addon and dotnet-external-provider images that can be run on a windows node, we need a way to:
1. let the hub know that this extension/provider can be run on a specific os/platform
2. let the hub know that an addon can be run on a specific os/platform
3. mechanism for determining under which conditions the addon/provider will be run on a windows node.

enhancements/multiple-provider/README.md
jortel: Since the addon and extensions run on the same pod, the addon would need to selected by OS.  The addon selector could be used for this.  For example: `tag:OperatingSystem=Windows` would select the Windows version of the analyzer addon.

enhancements/multiple-provider/README.md
rromannissen: I think the confusion here is what would be introducing the potential delay for the user requested analysis. My understanding is that language discovery does indeed take milliseconds to execute, but automated discovery would also include the technology stack/insights discovery, which implies the execution of an analysis via LSP and can effectively take minutes to complete. The end goal for automated discovery is to surface language and technology stack information at scale throughout the entire portfolio, so applications get automatically associated to their corresponding archetypes ASAP. But that shouldn't imply that the user is unable to do anything until automated tasks are finished, as that would lead to the system being perceived as unresponsive. The problem and the request are stated in #170

shawn-hurley (approved):
@pranavgaikwad @eemcmullan @JustinXHale 

After talking with Jeff, there are some changes on the preemption that we are adding; after this, I am a +1 on this enhancement!

rromannissen (approved):
@JustinXHale we need to add a new dialog/step in the analysis wizard to ask users if they want to prioritize their analysis over any background tasks that might be running and warn about the potential consequences. No big deal, we can figure it offline and attach it to the enhancement afterwards. Aside from that, for me it's a +1 for the enhancement!

mansam (approved):

PR #174, opened 2024-05-03
:sparkles: application insights

pranavgaikwad:


shawn-hurley (commented):
What about chaining and these rules? do we have any sort of validation or rules around this?

enhancements/analysis-insights/README.md
```diff
@@ -0,0 +1,168 @@
...
+
+The main argument against showing this information as "issues" is that it clutters the view. A large proportion of users are only interested in issues that affect migration. Meaning, issues that have non-zero effort and have a well-defined category. We do not intend to change that behvior with this enhancement. 
+
+While keeping the violations or issues as-is, we propose a new section in the analyzer output called "insights". It will contain _violation_ type for each tagging rule that matched with fine grained incidents. The only difference between an issue and an insight will be that the insights will not have a category and effort.
```
rromannissen, line 48: I'm not sure about Insights not having a Category. I can imagine some rules author wanting to categorize insights in the same fashion that tags are classified into tag categories. For example, imagine that someone is writing rules that identify architecture components from the different layers of an application. It could make sense to allow that rules author to create categories such as "Frontend" or "Persistence" to organize the different insights. @agoncal @kthatipally please let me know your thoughts on this.

pranavgaikwad: @rromannissen those are still categories on tags, but not on incidents that get created. Does it make sense? But note that tag categories are not the same as issue categories

pranavgaikwad:
> What about chaining and these rules? do we have any sort of validation or rules around this?

@shawn-hurley chaining will work as usual. I don't think I understand your concern

enhancements/analysis-insights/README.md
rromannissen: What about adding tag metadata in the incident section of the analysis output instead of using the category field? That would be information that the Hub could use to categorize insights so users are allowed to filter them.

shawn-hurley:
Just want to know if it was supported and if we could make sure that this is added to the enhancement. I assumed it should, but wanted to ask.

enhancements/analysis-insights/README.md
shawn-hurley: Would the rule labels suffice for the filtration? We may need some mechanism to save this information in the hub, I am not aware if the labels are there are not, but it does feel like labeling the rules with "Frontend" and/or "Persistence" would also make sense in that case.

rromannissen: Since insights would be mostly derived from rules that use the _tag_ action, I think it makes sense to have a way to associate tags with the insights they produce. That would be a more straightforward UX than what I was thinking about with reusing the _Category_ field. Users would go into the Insights view and filter insights based on tags, which is already a familiar concept for them. I'm not sure about modeling this with anything else, as we could have potential mismatches between tags and labels/categories, and that would feel odd from a UX perspective.

enhancements/analysis-insights/README.md
```diff
@@ -0,0 +1,168 @@
...
+
+	Skipped []string `yaml:"skipped,omitempty" json:"skipped,omitempty"`
+}
+```
+
+The rules that satisfy the following conditions will generate incidents in the insights section of the output:
+
+1. Rules that use the _tag_ action to generate tags
+  * These rules will generate tags as well as insights
+2. Rules that don't have an effort and category defined
```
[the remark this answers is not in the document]
shawn-hurley:
```yaml
- customVariables: []
  description: RESTful Web Services @Context annotation has been deprecated
  effort: 0
  message: Future versions of this API will no longer support `@Context` and related
    types such as `ContextResolver`.
  ruleID: jakarta-ws-rs-00001
  when:
    java.referenced:
      location: ANNOTATION
      pattern: jakarta.ws.rs.core.Context
```

What tag would make sense for this rule? or is this not an insight? 

I am asking to understand what the impact of the proposal made would have on a concrete rule example, just to better understand.

enhancements/analysis-insights/README.md
pranavgaikwad: @rromannissen I feel like category field should not be re-used. We can add the labels containing the original tags on the incident and then in the UI, look at these labels and provide a filtering option. We can also show the value of the label in the table view.

enhancements/analysis-insights/README.md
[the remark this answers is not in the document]
rromannissen: I would add something like a generic "Deprecated API" tag to that rule, so the user is able to filter and find all deprecated APIs in a given application. If we want, we can add an additional more fine grained tag like "Jakarta JAX-RS Deprecated API". That could be a convention we adhere to in curated rules that come out of the box with Konveyor, but there would be no way to enforce it with custom rules development. The tag itself being applied to applications in the portfolio would also be useful to identify the ones that might require some actions in the future.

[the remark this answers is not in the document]
rromannissen: @pranavgaikwad regarding non-zero effort value on incidents, I agree with you. Insights shouldn't have effort at all, and incidents related to a violation should always have some magnitude or they wouldn't be issues at all.

enhancements/analysis-insights/README.md
kthatipally: Love the write-up. I agree with @rromannissen that categorizing insights is beneficial for providing customers with a holistic view. However, I'm of the opinion that our existing tag categories (e.g., caching, backend) are generic for the application and cannot trace back to the actual rules in the existing implementation. I love the idea of incorporating labels instead, as they offer a way for customers to filter insights and can essentially serve the same purpose as tags.

enhancements/analysis-insights/README.md
```diff
@@ -0,0 +1,168 @@
...
+	Unmatched []string `yaml:"unmatched,omitempty" json:"unmatched,omitempty"`
+
+	Skipped []string `yaml:"skipped,omitempty" json:"skipped,omitempty"`
+}
+```
+
+The rules that satisfy the following conditions will generate incidents in the insights section of the output:
+
+1. Rules that use the _tag_ action to generate tags
+  * These rules will generate tags as well as insights
```
kthatipally, line 145: @pranavgaikwad , Regarding the tagging rules, 
The existing tagging rules definition specifies that rules only include 'tag' as actions without additional tag metadata like messages or incidents. e,g (https://github.com/konveyor/analyzer-lsp/blob/005a232243a9682f2bd1f0dd18379b0c1494bc8b/rule-example.yaml#L88 ). Given the proposed implementation for generating insights and compatibility with existing rules, insights would likely be generated with empty messages and incidents for those rules. This isn't particularly helpful. Should we consider adding metadata (messages) to these tagging rules to enhance their utility?

enhancements/analysis-insights/README.md
```diff
@@ -0,0 +1,168 @@
...
+
+![](./AppInsights.png)
+
+The page is similar to issues page except theres no category or effort.
+
+## Implementation Details / Concerns
+
+### There can be too much data generated
+
+Although the new information is separated into its own section, the output can grow because incidents from tags can be repetitive. Right now, in this enhancement, we are not proposing a solution for reducing unnecessary data. But there's an ask from the community for a feature that [allows rule authors to suppress information](https://github.com/konveyor/rulesets/issues/70). I think that ties into here perfectly. In future, we can have a switch on the rule itself that allows tweaking what gets generated. For now, we already have incident limits to limit the total number of incidents. We do not expect to see output size increase exponentially with the proposed changes.
```
kthatipally, line 161: @pranavgaikwad , Are you referring to the repetitive incidents that occur when multiple tags are present in the rule? Eg:  https://github.com/konveyor/rulesets/blob/d930baf2645d6a7ef9486c4f81bc40d8abc3a6af/default/generated/eap6/81-seam-ui.windup.yaml#L22

pranavgaikwad: Not specifically when there are more than one tags, but yeah..this is a good example. There aer instances where you could see this with single tags too.

enhancements/analysis-insights/README.md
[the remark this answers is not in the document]
pranavgaikwad: @rromannissen @shawn-hurley I updated this section to make it clear what happens in each scenario based on what a rule defines.

shawn-hurley (approved):
LGTM

Issue #177, opened 2024-05-10
[RFE] Gradle support


jmle:

jmle:
Addressed through https://github.com/konveyor/enhancements/pull/178

PR #178, opened 2024-05-10
:sparkles: Gradle support

jmle:
See https://github.com/konveyor/enhancements/issues/177

### DIFF
--- a/enhancements/gradle-support/README.md
+++ b/enhancements/gradle-support/README.md
@@ -71,19 +71,20 @@ On the other hand, it is a widely used convention to include a [Gradle wrapper](
 
 #### Backwards compatibility
 As mentioned previously, Gradle often releases new versions. This increases the difficulty of having backwards [compatibility](https://docs.gradle.org/current/userguide/compatibility.html) with older versions of Gradle.
+Information about backwards compatibility will have to be added to the docs, potentially in the form of a table.
 
 Since Konveyor is a Migration & Modernization framework, we should make the enhancement as *backwards compatible as possible*. Projects that use older versions of Gradle are expected.
 
 ### Security, Risks, and Mitigations
 
 #### Constrained development environments
-Some users might work within environments constrained in different forms, usually for security reasons. While the usage of Konveyor within such environments doesn't imply a security risk, the enhancement should provide the possibility of adapting to them.
+Some users might work within environments constrained in different forms, usually for security reasons. While the usage of Konveyor within such environments doesn't imply a security risk, the enhancement should provide the possibility of adapting to them. The hub will probably have to adapt to working in these conditions with Gradle.
 
 ## Design Details
 
 ### Test Plan
 
-TBD when targeting release
+- e2e test to be added to the [tests repo](https://github.com/konveyor/go-konveyor-tests).
 
 ## Implementation History
 TBD
--- a/enhancements/gradle-support/README.md
+++ b/enhancements/gradle-support/README.md
@@ -31,7 +31,7 @@ As Konveyor's user base increases, the need to support Gradle has become evident
 ### Goals
 
 - The ability to analyze the code of Java Gradle projects
-- The ability to asses the dependencies of Java Gradle projects
+- The ability to identify the dependencies of Java Gradle projects
 - The ability to analyze the dependencies of Java Gradle projects
 - The ability to be as backwards compatible as possible in relation to Gradle versions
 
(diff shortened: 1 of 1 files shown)


enhancements/gradle-support/README.md
```diff
@@ -0,0 +1,89 @@
...
+While not as widely used as Maven, many Java projects use [Gradle](https://gradle.org/) as their build tool. It is therefore desirable for Konveyor to be able to analyze these projects, and for this ability to be out-of-the-box and transparent for the user.
+
+## Motivation
+
+As Konveyor's user base increases, the need to support Gradle has become evident. For instance, Gradle has been traditionally used for Android projects. Leaving Gradle out of scope for Konveyor would mean leaving a potentially big chunk of users unable to use the tool.
+
+### Goals
+
+- The ability to analyze the code of Java Gradle projects
+- The ability to asses the dependencies of Java Gradle projects
```
rromannissen, line 34: _assess_ is an overloaded word with a very specific meaning in the Konveyor domain. I would suggest replacing it with _identify_.

enhancements/gradle-support/README.md
```diff
@@ -0,0 +1,89 @@
...
+- Support for Groovy, Kotlin, or other programming languages usually associated with Gradle
+
+## Proposal
+
+The two main areas concerning support for Gradle are the analysis of the source code, done by Konveyor's rules engine through the usage of a language server, and the assessment of the dependencies of the project (also required for certain features of the rules engine).
+
+### User Stories
+
+- As a user, I want to be able to analyze the source code of my Gradle Java project.
+- As a user, I want to be able to get an assessment of the dependencies used by my Gradle Java project, both direct and transitive.
```
rromannissen, line 49: For the same reasons from my previous comment, I would suggest something like _As a user, I want to be able to identify the dependencies used by my Gradle Java project, both direct and transitive._

rromannissen (changes requested):
Aside from the overloaded terms, which should be changed, the enhancement looks good to me!

aufi (approved):
Looks good to me. Based on the enhancement text and grade docs, older gradle versions compatibility might be challenging, otherwise looks clear and understandable.

enhancements/gradle-support/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+#### Gradle wrapper
+Unlike Maven, which can be considered quite stable, Gradle often releases new versions and updates. It is not mandatory for a Gradle project to specify in its configuration or settings which Gradle version it needs.
+
+On the other hand, it is a widely used convention to include a [Gradle wrapper](https://docs.gradle.org/current/userguide/gradle_wrapper.html) in every Gradle project, which is able to download the required Gradle version and execute it automatically. Is this why this enhancement **will require the presence of the Gradle wrapper in every Gradle project being analyzed**.
+
+#### Backwards compatibility
+As mentioned previously, Gradle often releases new versions. This increases the difficulty of having backwards [compatibility](https://docs.gradle.org/current/userguide/compatibility.html) with older versions of Gradle.
+
+Since Konveyor is a Migration & Modernization framework, we should make the enhancement as *backwards compatible as possible*. Projects that use older versions of Gradle are expected.
```
shawn-hurley, line 75: I think that a table or something that shows the support for versions is going to be necessary. If we don't know now, that is fine, but I did want to call this out and make sure that we doc this.

enhancements/gradle-support/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+### Security, Risks, and Mitigations
+
+#### Constrained development environments
+Some users might work within environments constrained in different forms, usually for security reasons. While the usage of Konveyor within such environments doesn't imply a security risk, the enhancement should provide the possibility of adapting to them.
```
shawn-hurley, line 80: Does gradle have a settings file or something similar w/ auth/z for contacting a repository of the dependencies. If it does I assume the hub will have to add support.

enhancements/gradle-support/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Design Details
+
+### Test Plan
+
+TBD when targeting release
```
shawn-hurley, line 86: I think a line here about adding an e2e test, on the hub's go-konveyor-test repo will be worth calling out.

shawn-hurley (approved):
LGTM, for the most part, I think there are three things that we may want to add or clarify, but in spirit, this is good to go!

pranavgaikwad (approved):

enhancements/gradle-support/README.md
cpolizzi: I second this suggestion and also believe it will be necessary to do so.

enhancements/gradle-support/README.md
cpolizzi: - HTTPS Auth: https://docs.gradle.org/current/userguide/declaring_repositories.html#sec:authentication_schemes
- Credential handling: https://docs.gradle.org/current/userguide/declaring_repositories.html#sec:handling_credentials

Hope this helps.

enhancements/gradle-support/README.md
jmle: Agreed, we will need to add something like that to the docs.

enhancements/gradle-support/README.md
jmle: Thanks @cpolizzi

PR #179, opened 2024-05-15
:sparkles: kantra support .net provider on windows

djzager:


enhancements/kantra-dotnet-framework/README.md
```diff
@@ -0,0 +1,122 @@
...
+* Propose modifications to Kantra to support running Windows containers to perform the analysis.
+
+### Non-Goals
+
+* Discuss in-cluster (ie. via the hub) analysis of .NET Framework applications
+* Discuss ruleset(s) for .NET Framework applications.
+
+## Proposal
+
+Introduce a new .NET Provider for Kantra. Using
```
eemcmullan, line 78: Right now, we use a 1:1 mapping of language to provider. This sounds like we may need to change the mapping for the .net case, since there is only one .net provider, but two(?) possible configs for it. We also may need to change the naming convention of the "provider map"in kantra for this to not be confusing.

enhancements/kantra-dotnet-framework/README.md
```diff
@@ -0,0 +1,122 @@
...
+
+## Open Questions
+
+For these open questions, we lack the requisite knowledge/experience with
+Windows containers to answer these questions definitively before accepting the
+enhancement. This enhancement should be updated when these are answered
+definitely one way or another, especially if they change the implementation
+details.
+
+1. Currently, Kantra assumes that it can create a single network via `podman`
```
eemcmullan, line 37: I wonder if instead of creating a network beforehand, we could publish a known port as the provider container starts?

enhancements/kantra-dotnet-framework/README.md
```diff
@@ -0,0 +1,122 @@
...
+This story is no different technically than a user wanting to analyze a `go` or
+`java` project today. Kantra would need to start a Linux container for the
+dotnet-external-provider, construct the appropriate provider configuration, and
+perform the analysis. The work required here would simply be recognizing
+projects in this category and constructing the provider settings correctly.
+
+#### Analyze Windows-only .NET Project (.NET Framework 4.5 through .NET Framework 4.8)
+
+This story is **fundamentally** different than how Kantra runs today, in that, it requires Kantra
+to start a Windows container that only `docker` supports. Additionally, communication between
```
eemcmullan, line 99: Currently, kantra defaults to Podman, and if it cannot find it, tries to use Docker instead. Will need to default to Docker for this case.

enhancements/kantra-dotnet-framework/README.md
```diff
@@ -0,0 +1,122 @@
...
+
+Attempting to analyze Windows-only projects using Windows containers alongside
+Linux containers sharing a network and volumes could drastically increase the
+complexity of the `kantra analyze` command if not totally undo some of the
+basic assumptions of the analysis procedure.
+
+## Alternatives
+
+In the case that the proposed changes drastically increase the complexity of the
+`kantra analyze` command OR sharing of either container networks and volumes is not possible
```
eemcmullan, line 119: +1 for additional flag

shawn-hurley (approved):
Looks good to me, and adaquently expresses the desire and high level oh how we will achieve it. 

As we answer the questions, we do need to make sure that we come back and update but everyone is already on board for that.

enhancements/kantra-dotnet-framework/README.md
```diff
@@ -0,0 +1,122 @@
...
+For these open questions, we lack the requisite knowledge/experience with
+Windows containers to answer these questions definitively before accepting the
+enhancement. This enhancement should be updated when these are answered
+definitely one way or another, especially if they change the implementation
+details.
+
+1. Currently, Kantra assumes that it can create a single network via `podman`
+   or `docker` and use that for all containers it starts. Can the network of
+   Linux and Windows containers be shared?
+1. Currently, Kantra creates volume(s) to be shared with each of the providers
```
eemcmullan, line 40: I would be interested to see if we can get around using a shared volume in this case. The shared volume is used only for the source application between the provider and analyzer engine containers. We don't need to change any data in this volume, so I wonder if instead we could mount the source app into both containers separately.

eemcmullan (approved):

eemcmullan:
LGTM. Just some comments on things to think about once we learn more on how things will work with kantra and Windows containers.

enhancements/kantra-dotnet-framework/README.md
djzager: I think the latest update, sidestepping the provider logic if the app, according to alizer is C# and uses the v4.5 Framework, addresses your concern here.

enhancements/kantra-dotnet-framework/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+Introduce a new .NET Provider for Kantra. We will do this by providing
+dotnet-external-provider, analyzer-lsp, and kantra as windows container images.
+Using [alizer](https://github.com/devfile/alizer) to determine the language and
+frameworks used, we can differentiate between .NET projects that can be
+analyzed in a Linux container versus .NET Framework projects that must be
+analyzed in a Windows container. With this knowledge, when we encounter a
+Windows-only .NET Project, we will break away from the normal `analyze`
+subcommand flow and analyze the application using the dotnet-external-provider
+and kantra Windows containers. Once we execute the analysis, we canr etrieve
```
eemcmullan, line 88: NIT: typo "canr etrieve"

PR #182, opened 2024-06-12
:sparkles: Initial draft of generic authentication and authorization

shawn-hurley:


jmontleon:
Does this replace or make use of the openshift auth functionality we have in the operator today?
https://github.com/konveyor/operator/blob/main/roles/tackle/tasks/main.yml#L45

shawn-hurley:
Going to be honest @jmontleon I had no idea that existed, and I am not convinced that this would even work. @aufi @jortel Can someone let me know if this even is a valid option?

jmontleon:
> Going to be honest @jmontleon I had no idea that existed, and I am not convinced that this would even work. @aufi @jortel Can someone let me know if this even is a valid option?

I added this. It worked at the time, but I have not tested it in a long time. It uses the OpenShift oauth-proxy. It only provided authentication, no authorization, so once authenticated you could do anything.

shawn-hurley:
Thank you for reminding me about this! I had forgetten this option. I have a feeling it still works and something that I will need to add:
1. What to do with this option
2. What to do IF this option is in use during upgrade

rromannissen (changes requested):
Please replace mentions to MTA with Konveyor

PR #196, opened 2024-08-12
Containerless Kantra Distribution for Java Analysis

### LINKED ISSUE #188: Containerless distributions of the Kantra CLI - Java
### What is the problem?

The Kantra CLI requires a container runtime to be present at the host it runs as the analyzer engine and the different language providers are spin up in containers.

### Why is this a problem?

Highly regulated organizations from verticals like FSI and public agencies have very strict security policies that severely restrict the software that can be installed in corporate computers. Container runtimes are in most cases not allowed due to the fact that they would allow the user to freely run software outside of the security policies. This leads to users coming from these business vertical not being able to use the Kantra CLI, and by extension, Konveyor as a project to help them with their migration needs. 

On top of that, the current approach of the Kantra CLI makes it very hard to have it integrated on CI/CD pipelines, as the requirement of having direct access to the container runtime to spin up analyzer and provider containers only allows it to run on VMs and bare metal hosts, making it incompatible with Jenkins container agents and Tekton tasks.

### Proposed solution

Create self contained bundles for the CLI that package the analyzer and the provider for a given language, so they can be installed directly in the host without the need of a container runtime. The following should be taken into account when putting these bundles together:
- Requirements for the bundle to run shouldn't be too complex in terms of configuration, as users on those restricted environments normally have limited access to it. Admin permissions and privileged execution requirements must be avoided as well as users are unlikely to have access to them.
- Requirements for each bundle should be clearly documented. For example, OpenJDK will be required for the Java bundle, whereas certain .NET releases might be required for the .NET bundle.
- Windows should be the top priority for the availability of these bundles, as it is the most widespread OS in these restricted corporate environments and the one in which these restrictive policies are being applied most commonly. Linux should follow, with macOS being the least priority.
- Bundles should be put together with containerization in mind to allow their execution in a modern CI/CD tool like Tekton or Jenkins with container agents.

Advanced users with access to container runtimes should be able to continue using Kantra as they do today, allowing them to use multiple providers with a single installation.

eemcmullan:
Closes #188

### DIFF
--- a/enhancements/containerless-kantra/README.md
+++ b/enhancements/containerless-kantra/README.md
@@ -32,9 +32,8 @@ This enhancement will discuss a new implementation of kantra that does not requi
 ## Motivation
 
 Kantra currently requires a container runtime. This presents problems for some users. 
-For example, an organization may have a strict regulations that do not allow installation of container runtimes on company computers. This results in the kantra CLI being unavailable to assist these companies with konveyor analysis.
 
-There is also the issue that the current implementation of kantra is incompatible with some CI/CD tools such as Jenkins container agents and Tekton tasks.
+For example, an organization may have a strict regulations that do not allow installation of container runtimes on company computers. This results in the kantra CLI being unavailable to assist these companies with konveyor analysis.
 
 ### Goals
 
@@ -51,8 +50,6 @@ There is also the issue that the current implementation of kantra is incompatibl
 ### User Stories 
 
 As a user, I want to run a Java application analysis using kantra that does not require installation of a container runtime.
-
-As a developer, I want to set up CI/CD pipelines with kantra using tools such as Jenkins.
 
 ## Implementation Details
 
--- a/enhancements/containerless-kantra/README.md
+++ b/enhancements/containerless-kantra/README.md
@@ -1,5 +1,5 @@
 ---
-title: Containerless Distributions of Kantra
+title: Containerless Distribution of Kantra for Java Analysis
 authors:
   - "@eemcmullan"
 reviewers:
@@ -10,11 +10,18 @@ reviewers:
 approvers:
   - TBD
 creation-date: 2024-08-09
-last-updated: 2024-08-11
+last-updated: 2024-08-15
 status: implementable
 ---
 
-# Containerless Distributions of Kantra
+## Release Signoff Checklist
+
+- [x] Enhancement is `implementable`
+- [x] Design details are appropriately documented from clear requirements
+- [x] Test plan is defined
+- [ ] User-facing documentation is created
+
+# Containerless Distribution of Kantra for Java Analysis
 
 ## Summary
 
@@ -51,7 +58,7 @@ As a developer, I want to set up CI/CD pipelines with kantra using tools such as
 
 ### Requirements
 
-Users will be expected to have or install several tools onto their machines, such as OpenJDK and Maven. We will produce and ship our [rulesets](https://github.com/konveyor/rulesets/), JDTLS, and the [Java Bundle](https://github.com/konveyor/java-analyzer-bundle).
+Users will be expected to have or install several tools onto their machines, such as OpenJDK, Maven, fernflower, and tar. We will produce and ship our [rulesets](https://github.com/konveyor/rulesets/), JDTLS, and the [Java Bundle](https://github.com/konveyor/java-analyzer-bundle).
 
 The user will be instructed to set these packaged binaries and rules in a known location, such as `$HOME/.kantra/`. Alternatively, an install script can set these.
 
@@ -83,22 +90,30 @@ There will be a new package in kantra, `no-container`, that will be added for th
 
 With the analyzer and external Java provider imported, kantra will create and init an internal Java provider, which will be defined as an analyzer `InternalProviderClient`. This Java provider will not require an additional binary nor require running outside of the analyzer engine. It will be started as an internal provider rather than a GRPC provider.
 
-A Java provider config will be created in kantra, which will include the `binaryPath` as the packaged JDTLS binary instead of the Java provider address or binary.
+A Java provider config will be created in kantra, which will include the `binaryPath` as the packaged JDTLS binary instead of the Java provider address or binary. By default, we will also include and run the `builtin` provider.
 
 ```yaml
     {
-        "name": "java",
-        "binaryPath":  "$HOME/.kantra/jdtls/bin/jdtls",
-        "initConfig": [
+        Name: "java",
+        BinaryPath:  "$HOME/.kantra/jdtls/bin/jdtls",
+        InitConfig: [
             {
-                "location": "<input>",
-                "providerSpecificConfig": {
-                    "bundles": "$HOME/.kantra/jdtls/java-analyzer-bundle/java-analyzer-bundle.core/target/java-analyzer-bundle.core-1.0.0-SNAPSHOT.jar",
-                    "lspServerPath": "$HOME/.kantra/jdtls/bin/jdtls"
+                Location: "<input>",
+                ProviderSpecificConfig: {
+                    Bundles: "$HOME/.kantra/jdtls/java-analyzer-bundle/java-analyzer-bundle.core/target/java-analyzer-bundle.core-1.0.0-SNAPSHOT.jar",
+                    LspServerPath: "$HOME/.kantra/jdtls/bin/jdtls"
                 },
-                "analysisMode": "full"
+                AnalysisMode: "full"
             }]
     },
+		{
+			Name: "builtin",
+			InitConfig: [
+				{
+					Location:     <input>,
+					AnalysisMode: "full",
+				}]
+			}
 ```
 
 Kantra will then run the packaged rules and/or custom rules with this provider. Once the analysis is completed, a static report will be generated internally in the output directory.
--- a/enhancements/containerless-kantra/README.md
+++ b/enhancements/containerless-kantra/README.md
@@ -58,15 +58,15 @@ As a developer, I want to set up CI/CD pipelines with kantra using tools such as
 
 ### Requirements
 
-Users will be expected to have or install several tools onto their machines, such as OpenJDK, Maven, fernflower, and tar. We will produce and ship our [rulesets](https://github.com/konveyor/rulesets/), JDTLS, and the [Java Bundle](https://github.com/konveyor/java-analyzer-bundle).
+Users will be expected to have or install several tools onto their machines, such as OpenJDK, Maven, fernflower, and tar. We will produce and ship our [rulesets](https://github.com/konveyor/rulesets/), JDTLS, and the [Java Bundle](https://github.com/konveyor/java-analyzer-bundle) as release artifacts.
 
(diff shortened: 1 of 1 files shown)


enhancements/containerless-kantra/README.md
```diff
@@ -0,0 +1,118 @@
...
+                "providerSpecificConfig": {
+                    "bundles": "$HOME/.kantra/jdtls/java-analyzer-bundle/java-analyzer-bundle.core/target/java-analyzer-bundle.core-1.0.0-SNAPSHOT.jar",
+                    "lspServerPath": "$HOME/.kantra/jdtls/bin/jdtls"
+                },
+                "analysisMode": "full"
+            }]
+    },
+```
+
+Kantra will then run the packaged rules and/or custom rules with this provider. Once the analysis is completed, a static report will be generated internally in the output directory.
```
pranavgaikwad, line 104: Just for my understanding, does this mean we will be re-writing [cmd/analyzer/main.go](https://github.com/konveyor/analyzer-lsp/blob/main/cmd/analyzer/main.go) in Kantra?

pranavgaikwad (commented):
I like the idea overall. I am requesting that we include all pre-requisites for the user to be able to run this on their machines. Also, can we talk about Builtin provider too?

enhancements/containerless-kantra/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+As a user, I want to run a Java application analysis using kantra that does not require installation of a container runtime.
+
+As a developer, I want to set up CI/CD pipelines with kantra using tools such as Jenkins.
+
+## Implementation Details
+
+### Requirements
+
+Users will be expected to have or install several tools onto their machines, such as OpenJDK and Maven. We will produce and ship our [rulesets](https://github.com/konveyor/rulesets/), JDTLS, and the [Java Bundle](https://github.com/konveyor/java-analyzer-bundle).
```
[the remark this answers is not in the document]
shawn-hurley: +1 I forgot about the fernflower peice

enhancements/containerless-kantra/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+The user will be instructed to set these packaged binaries and rules in a known location, such as `$HOME/.kantra/`. Alternatively, an install script can set these.
```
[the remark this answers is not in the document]
shawn-hurley: For JDT.LS and java analyzer bundle, for unistall purposes it would be nice I think if this was all in `.kantra/<folders>` for ease of use, Maven, OpenJDK should be system paths and we use the correct env vars ($M2_<thing> and $JAVA_<thing>) IMO

enhancements/containerless-kantra/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Design Details
+
+### Technical Implmentation
+
+A new analyze subcommand can be introduced for kantra to recognize when to start the containerized version of analysis versus the non-containerized option. This could look like `kantra analyze-java-only <options>`. Another option for starting the analyze subcommand without running containers would be to detect existing packaged requirements at the expected location discussed in the [requirements](#requirements). However, in this case, we would want to provide an option to disable this behavior.
```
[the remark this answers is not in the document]
shawn-hurley: It is "restricte" in that you can't use any other providers (besides builtin which you called out we need to add :) good catch!)

enhancements/containerless-kantra/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+### Non-Goals
+
+- Discuss containerless distributions of kantra outside of Java analysis
+
+## Proposal
+
+### User Stories 
+
+As a user, I want to run a Java application analysis using kantra that does not require installation of a container runtime.
```
djzager, line 46: Does this carry with it the implication that we distribute a binary artifact? As I understand it, users will still need a container runtime to install kantra.

enhancements/containerless-kantra/README.md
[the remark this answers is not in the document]
djzager: Are we going to allow these to be in the system path?

enhancements/containerless-kantra/README.md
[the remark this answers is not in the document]
djzager: I wish I could come up with a decent `analyze-no-container` name that could be more generic.

djzager (commented):
I think this genuinely improves on what I've heard previously discussed. I do somewhat wonder what prevents us from creating a new command altogether (separate go module) and whether that may give us some more flexibility even if in the same repository.

enhancements/containerless-kantra/README.md
eemcmullan: Correct - we will ship with it all necessary binaries

enhancements/containerless-kantra/README.md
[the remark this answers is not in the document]
eemcmullan: Yea, I am still not sure of this subcommand name, so I'm open to any other suggestions.

enhancements/containerless-kantra/README.md
eemcmullan: +1

enhancements/containerless-kantra/README.md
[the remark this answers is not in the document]
eemcmullan: Install script to add `.kantra/<folders>/...` to path? Or users add it manually? And set those as env vars in kantra? And +1 for using expected env vars for Maven and OpenJDK. @djzager Are you instead thinking of using the binaries from the abs path?

eemcmullan:
"I think this genuinely improves on what I've heard previously discussed. I do somewhat wonder what prevents us from creating a new command altogether (separate go module) and whether that may give us some more flexibility even if in the same repository." 
@djzager This is an interesting idea. It would make it more light-weight as well if we do not have to build the entirety of kantra for this distribution. I'd be curious to know what others think @pranavgaikwad @shawn-hurley

enhancements/containerless-kantra/README.md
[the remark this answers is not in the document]
djzager: What I meant was that, if a user installs the binary at `/usr/local/bin` then it should be accessible via Kantra right?

I suppose that what I am suggesting is that we may install stuff at `${XDG_CONFIG_HOME}/.kantra/...` but we should encourage the users to set their `PATH` to include that directory so `kantra` the CLI is free to just execute them (and not go looking for them or expecting environment variables).

enhancements/containerless-kantra/README.md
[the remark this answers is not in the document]
djzager: The best I could come up with `kantra analyze-bin --provider=java`, gives us ability to expand to other providers.

enhancements/containerless-kantra/README.md
```diff
@@ -0,0 +1,133 @@
...
+Kantra is the Konveyor CLI used to analyze and transform applications. As of now, it requires a container runtime due to the analyze and transform commands running in containers. 
+
+This enhancement will discuss a new implementation of kantra that does not require a container runtime for analysis. This enhancement will also only focus on the first phase of this implementation: a containerless distribution that will run a Java analysis. 
+
+## Motivation
+
+Kantra currently requires a container runtime. This presents problems for some users. 
+For example, an organization may have a strict regulations that do not allow installation of container runtimes on company computers. This results in the kantra CLI being unavailable to assist these companies with konveyor analysis.
+
+There is also the issue that the current implementation of kantra is incompatible with some CI/CD tools such as Jenkins container agents and Tekton tasks.
```
shawn-hurley, line 37: Nit: Tekton can run container together, I would have a difficult time understanding that limitiation.

shawn-hurley: I also am having a hard time with jenkins, it should be fine if jenkins is on a dedicated server, unless the organization has locked it down (covered in the line above) and if it is in an k8s environement, this it should be able to run containers to do things on that system. 

I don't know if we should be calling this out. but that might just be my two cents and I could be wrong

enhancements/containerless-kantra/README.md
shawn-hurley: I don't believe it is true that they need a container runtime to install, upstream we should move to publishing the binaries for download, and potentially even tars with everything need in it. 

For downstream I believe we will have tars with everything in it, no `podman cp`. 

If we were in agreement then I am sorry for misunderstanding this thread

enhancements/containerless-kantra/README.md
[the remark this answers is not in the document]
shawn-hurley: I am a +1 on @djzager's flag to add more, or the one in the enhancement

enhancements/containerless-kantra/README.md
```diff
@@ -0,0 +1,133 @@
...
+│   └── no-container
+│       │   ├── no-container.go
+│       │   ├── static-report.go
+│       ├── java
+│       │   ├── internal-java-provider.go
+│       ├── dotnet
+│   ├── .....
+```
+
+With the analyzer and external Java provider imported, kantra will create and init an internal Java provider, which will be defined as an analyzer `InternalProviderClient`. This Java provider will not require an additional binary nor require running outside of the analyzer engine. It will be started as an internal provider rather than a GRPC provider.
```
shawn-hurley, line 91: Why not import the external-provider from the package?

shawn-hurley (approved):
Understanding why a new InternalProviderClient and not just importaing from external is the only open question in my book

Otherwise this LGTM

enhancements/containerless-kantra/README.md
djzager: My fault. I should have been more clear. I do believe that we should document that, as part of the implementation of this enhancement, we will deliver binaries (or tars) as release artifacts for Kantra (and any other necessary component).

djzager (approved):
This looks good to me. The only two (non-blocking) comments I have are:

1. To consider documenting that we will start shipping binaries as the release artifacts for kantra (and any other necessary component) as part of this work.
2. Consider `kantra analyze-bin --provider=java` as a way to satisfy the requirements with room to grow in the future.

enhancements/containerless-kantra/README.md
eemcmullan: @djzager I think that should be included in the `requirements` section.

enhancements/containerless-kantra/README.md
eemcmullan: We could have that as an option, but that'd be another binary to ship, right? The analyzer would either need that or the provider running locally if we do not use it internally.

enhancements/containerless-kantra/README.md
eemcmullan: @shawn-hurley I can leave this section out if we want to have the enhancement only focus on the kantra implementation part.

shawn-hurley: +1 I don't think it is necessary and I think that it is potetially confusing

enhancements/containerless-kantra/README.md
shawn-hurley: It sounds like we are saying the same thing, and I may have misunerstood. Resolving this!

djzager (approved):
ACK

PR #197, opened 2024-08-14
:sparkles: dismissal of incidents

shawn-hurley:
Tried to capture the changes across the projects, keeping the logic for dismissal in the analyzer and adding the UI/UX for how a user could use this feature

### DIFF
--- a/enhancements/dismiss-issues/README.md
+++ b/enhancements/dismiss-issues/README.md
@@ -159,9 +159,9 @@ For the Hub it may make sense that the DIF is actually a new table in the databa
 
 ##### UI
 
-The UI will now have a new button to add to a given incident page. This button will allow the user to dismiss an incident. 
-The UI will also need to add a new page for a given application, that will allow you to see the "dismissed" incidents for that application. 
-The UI will also need to have the ability to download the DIF. This will allow for the user to give to someone running kantra locally.
+The UI will now have a new button to add to a given incident page. This button will allow the user to dismiss an incident. The user will also be able to see with dynamic filters/queries the dimissed issues in this page, and will be able to re-enable them with a similar button.
+The UI will also need to add a new page for a given application, that will allow you to see the DIF entries for that application, this is where you can download or import a DIF file as well.
+
 
 ### Security, Risks, and Mitigations
 
--- a/enhancements/dismiss-issues/README.md
+++ b/enhancements/dismiss-issues/README.md
@@ -1,5 +1,5 @@
 ---
-title: neat-enhancement-idea
+title: dismiss-issues
 authors:
   - "@shawn-hurley"
 reviewers:
@@ -44,8 +44,8 @@ An architect may consider that certain incidents are not on the migration path, 
 ### Goals
 
 1. Allow users to filter incidents that they do not believe are important for their migration.
-2. Persist this filter across runs of analysis in both kantra and the hub.
-3. Allow users to see the dismissed analysis report and via the UI (both tackle-ui and static-report)
+2. Persist this filter across runs of analysis in both Kantra and the hub.
+3. Allow users to see the dismissed analysis report via the UI (both tackle-ui and static-report)
 
 ### Non-Goals
 
@@ -66,7 +66,7 @@ As an architect, I would want the dismissed incidents to persist over subsequent
 
 #### Story 3
 
-As an architect, I would like to be able to export the dismissed rules to be used with a local run of kantra.
+As an architect, I would like to be able to export the dismissed incidents to be used with a local run of Kantra.
 
 #### Story 4
 
@@ -74,7 +74,7 @@ As an architect, I would like to see the dismissed issues for an analysis.
 
 #### Story 5
 
-As an architect, I would like to export a set of dismissed issues from a static report to be used in subsequent analysis runs of kantra.
+As an architect, I would like to export a set of dismissed issues from a static report to be used in subsequent analysis runs of Kantra.
 
 #### Story 6
 
@@ -85,25 +85,24 @@ This will be implemented across the portfolio, and the implementation will need 
 
 #### Analyzer-LSP
 
-In the analyzer, we will implement a new input format, that identifies dismissed issues by a combonation of:
-rule-id, file uri and line number. Today, this combination is already unique for an incident in the analysis run, so it is a good combonation that is already tested to represent a single issue. The command should be able to take more than one dismissed incident files(DIF from here on out).
+In the analyzer, we will implement a new input format, that identifies dismissed issues by a combination of ruleset, rule-id, URI and line number. Today, this combination is already unique for an incident in the analysis run, so it is a good combination that is already tested to represent a single issue. The command should be able to take more than one dismissed incident files(DIF from here on out).
 
-The analysis engine, will create a new output list, `dismissed`, that will contain for each combination found above, the same incident as if it were a violation/incident. The output will be the exact same as if it was in the violations.
+The analysis engine will create a new output list, `dismissed`, that will contain for each combination found above, the same incident as if it were a violation/incident. The output will be the same as if it was in the violations.
 
 #### Kantra & Static Report
 
-For kantra we will need to do two things. The first thing, is enhance the static reports to be able to dismiss incidents as well allow a user to download the DIF to the file system for future use. 
-The second thing we will need to do, is have a consistant way to map applications to the DIF. This file will need to be able to be shared between the hub and kantra.
+For Kantra we will need to do two things. The first thing is to enhance the static reports to be able to dismiss incidents as well as allow a user to download the DIF to the file system for future use. 
+The second thing we will need to do is have a consistent way to map applications to the DIF. This file will need to be able to be shared between the hub and Kantra.
 
 ##### Labeling Applications for Shared Reconigtion
 
-For java applications the most obvious answer is the maven identifier for `<group>.<artifact>.<version>` and potentially ignoring the version in this case. This will not work for other applications or platforms. In the hub we have an identifier for an Application (the Application ID) that would consistant for a single hub instance but that would mean nothing to the kantra user, who only has a filepath to either source code or a binary. Mandating a connection the hub in this case would be disaster as primary use case of kantra is to test out analysis before deploying a hub. 
+For Java applications, the most obvious answer is the maven identifier for `<group>.<artifact>.<version>` and potentially ignoring the version in this case. This will not work for other applications or platforms. In the hub, we have an identifier for an Application (the Application ID) that would be consistent for a single hub instance but that would mean nothing to the Kantra user, who only has a file path to either source code or a binary. Mandating a connection to the hub in this case would be a blocker as the primary use case of Kantra is to test out analysis before deploying a hub. 
 
-Given all of these constraints specified above, I do not believe that we can do anthing automatically, we will need user input.
+Given all of these constraints specified above, I do not believe that we can do anything automatically, we will need user input.
 
-##### Kantra specific application and dismisssed incidents label
+##### Kantra specific application and dismissed incidents label
 
-One goal should be that a user, once they make these dismissals, they will not need to do this on each subsequent run. To achieve this, we should have a new subcommand that will manage the DIF for a given application. 
+One goal should be that a user, once they make these dismissals, will not need to do this on each subsequent run. To achieve this, we should have a new subcommand that will manage the DIF for a given application. 
 
 ```kantra dismissed-incidents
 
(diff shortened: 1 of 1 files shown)


enhancements/dismiss-issues/README.md
```diff
@@ -0,0 +1,179 @@
...
+
+As an architect, I would like to dismiss incidents in the tackle-ui, and they are removed from the main view.
+
+#### Story 2
+
+As an architect, I would want the dismissed incidents to persist over subsequent runs of the analysis.
+
+#### Story 3
+
+As an architect, I would like to be able to export the dismissed rules to be used with a local run of kantra.
```
djzager, line 69: Probably want to be particular about terminology rule vs incident vs violation.

enhancements/dismiss-issues/README.md
```diff
@@ -0,0 +1,179 @@
...
+### Implementation Details/Notes/Constraints [optional]
+
+This will be implemented across the portfolio, and the implementation will need to be consistent across these repositories. Because of this, everything is already using the analysis results as a shared source of truth for a given analysis, so we should implement the underlying dismissal in this repository.
+
+#### Analyzer-LSP
+
+In the analyzer, we will implement a new input format, that identifies dismissed issues by a combonation of:
+rule-id, file uri and line number. Today, this combination is already unique for an incident in the analysis run, so it is a good combonation that is already tested to represent a single issue. The command should be able to take more than one dismissed incident files(DIF from here on out).
+
+The analysis engine, will create a new output list, `dismissed`, that will contain for each combination found above, the same incident as if it were a violation/incident. The output will be the exact same as if it was in the violations.
```
djzager, line 91: So this is for runs of the analyzer after particular issue(s) have been dismissed.

enhancements/dismiss-issues/README.md
```diff
@@ -0,0 +1,179 @@
+---
+title: neat-enhancement-idea
```
djzager, line 2: Might need to change this.

djzager (commented):
A couple of things:
1. I'm genuinely curious how much value we would "lose" by simply allowing users to dismiss at the rule level. This may be a declaration of ignorance but it isn't immediately obvious scenarios where a particular rule generates issues that must be fixed at the same time as ones that don't need to be addressed.
2. I wonder if we could perform the dismissing/filtering as a post-analysis step. My argument for this perspective is that I view this concern as a purely presentation layer concern. That is, at the analyzer level, these are absolutely still incidents; they just happen to be ones we don't want to see in analysis reports in the Konveyor UI or static-report.

There are some implications to these thoughts that I don't want to ramble on if I'm heading in the totally wrong direction.

enhancements/dismiss-issues/README.md
shawn-hurley: +1 you are correct need to be more specific

shawn-hurley:
> I'm genuinely curious how much value we would "lose" by simply allowing users to dismiss at the rule level. This may be a declaration of ignorance but it isn't immediately obvious scenarios where a particular rule generates issues that must be fixed at the same time as ones that don't need to be addressed.

You know, this is a good question that I don't know if I have an answer too. 

One thing that I was thinking of is work at the lowest level to give users the most flexability, and we could add a "bulk" update based on ruleID as a follow up. Now this doesn't mean that it is correct by any stretch of the imagination. 

Something that I was thinking about was a use case like local-storage. They are mandatory issues, that tell you to fix something, but let's say you have two use cases for local storage. In one use case, you are using a local file for configuration, this would most likely be a configmap in k8s and still use local storage. But let's say that you are also using a cache of data over local storage, this probably should be solved. 

So for the tackle test app case, you probably across the entire portfolio want to disable that particular incident for the application config, because you plan to solve that with a configmap across the entire portfolio. 

Just a thought though, wondering if that even makes sense.

shawn-hurley:
> I wonder if we could perform the dismissing/filtering as a post-analysis step. My argument for this perspective is that I view this concern as a purely presentation layer concern. That is, at the analyzer level, these are absolutely still incidents; they just happen to be ones we don't want to see in analysis reports in the Konveyor UI or static-report.

I agree they are still incidents, but because we have two layers (and potentially more with the specific language builds) doing this, I figured two things:

1. That adding it into the analyzer means a single implementation that everyone gets to use. 
2. Making it a first class citizen (like we did with insights for instance) gives the output of the analyzer more structure to get what you really want. Think about the Kai case, where they really just use the output of the analyzer. If in a company they want to ignore certain specific incidents (say similar to the local storage outlined above) then you probably want this to live in the analyzer. 

The analyzer could do this as a post process OR it could just add a field on the incidents instead of making a whole new list. I leaned to the whole new list because it "feels" like something you would want to see at the top level. I could be convinced that it should be at the lower level and just be a field on the incident, but I can't think of a compelling reason at the moment. 

Does that make sense? Should I add any of this to the enhancement?

enhancements/dismiss-issues/README.md
```diff
@@ -0,0 +1,179 @@
...
+2. It will allow NEW incidents to be dissmissed. This will remove them from the main incidents view, and move them to the dimissed view. 
+3. The dismissed view, will allow the user to download the DIF. 
+
+#### Hub and Tackle UI
+
+The Hub and Tackle UI will be enhanced in a couple of ways to achieve a similar UX like the static reports and kantra.
+
+##### Hub
+
+The hub will now allow a DIF to be attached to application. The analyzer-addon will be responsible for pulling this file when running anlysis. This will follow a similar pattern to other application information that is passed to the addon, and the pulling of the file will follow a similar pattern to pulling the rulesets. (@jortel can you verify that this makes sense or how you would do it?).
```
jortel, line 139: Yes. I would think similar to rulesets, the DIF for the application would be fetched/pulled by the addon as needed.

enhancements/dismiss-issues/README.md
```diff
@@ -0,0 +1,179 @@
...
+
+The Hub will also have a to dynamically create and append this DIF as the UI dismisses incidents. To achive this, I propose a new field on incidents, called dismissed that is a boolean. This new incident update API will allow for the UI to tell the hub when an incident has been dismissed. When this happens the hub will need to:
+1. Move the incident to dismissed
+2. Add the incident to the either already existing DIF or create one and add it.
+
+For the Hub it may make sense that the DIF is actually a new table in the database, and it is generated on retrieval. I think this is an open question to be discussed with the Hub team.
```
jortel, line 145: Likely a new 1-* table with _foreign-key_ to application.
How many entries per application expected?  hundreds?  thousands?  more?

enhancements/dismiss-issues/README.md
```diff
@@ -0,0 +1,179 @@
...
+
+As an architect, I would like to import a list of dismissed issues to the hub for subsequent analysis runs.
+### Implementation Details/Notes/Constraints [optional]
+
+This will be implemented across the portfolio, and the implementation will need to be consistent across these repositories. Because of this, everything is already using the analysis results as a shared source of truth for a given analysis, so we should implement the underlying dismissal in this repository.
+
+#### Analyzer-LSP
+
+In the analyzer, we will implement a new input format, that identifies dismissed issues by a combonation of:
+rule-id, file uri and line number. Today, this combination is already unique for an incident in the analysis run, so it is a good combonation that is already tested to represent a single issue. The command should be able to take more than one dismissed incident files(DIF from here on out).
```
jortel, line 89: It is my understanding the ruleid isn't unique across ruleset. Shouldn't DIF entries be qualified by ruleset?
What is the format of DIF files?  Assuming YAML but would help to be explicit.

enhancements/dismiss-issues/README.md
```diff
@@ -0,0 +1,179 @@
...
+
+For java applications the most obvious answer is the maven identifier for `<group>.<artifact>.<version>` and potentially ignoring the version in this case. This will not work for other applications or platforms. In the hub we have an identifier for an Application (the Application ID) that would consistant for a single hub instance but that would mean nothing to the kantra user, who only has a filepath to either source code or a binary. Mandating a connection the hub in this case would be disaster as primary use case of kantra is to test out analysis before deploying a hub. 
+
+Given all of these constraints specified above, I do not believe that we can do anthing automatically, we will need user input.
+
+##### Kantra specific application and dismisssed incidents label
+
+One goal should be that a user, once they make these dismissals, they will not need to do this on each subsequent run. To achieve this, we should have a new subcommand that will manage the DIF for a given application. 
+
+```kantra dismissed-incidents
```
eemcmullan, line 108: Would this make more sense as a subcommand for analyze? It is only relevant to the analyze command, and that way it wouldn't be on the same level as `kantra transform` and `kantra version` etc..

enhancements/dismiss-issues/README.md
shawn-hurley: Correct, it will be part of the standard output (added to the output type).

enhancements/dismiss-issues/README.md
```diff
@@ -0,0 +1,179 @@
...
+
+Application Path        Dismissed File
+abs-path-to-file        <path to DIF>
+abs-path-to-file        <path to DIF>
+abs-path-to-file        <path to DIF>
+```
+
+On the run of an analysis, kantra will look at the abs path to input and determine if there is a DIF(s). We will then  use those files during the analysis of the applications. 
+
+Kantra will also allow a one time use of the DIF(s), using a new flag on analyze:
```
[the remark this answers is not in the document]
shawn-hurley: updating this, but the JS in the browser actually has no way of access files on disk AFAICT. This means the user will actually have to download the file, that is calculated in memory.

enhancements/dismiss-issues/README.md
shawn-hurley: Resolviing, I removed the question

enhancements/dismiss-issues/README.md
shawn-hurley: resolving as it looks like we agree, let me know if you want me to change something else!

enhancements/dismiss-issues/README.md
shawn-hurley: Updated to add ruleset as a part of the identifier.

enhancements/dismiss-issues/README.md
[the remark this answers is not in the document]
shawn-hurley: I tried to make this more clear in the static report flow, if that doesn't work let me know. Resolving this comment for another thread if it is needed.

djzager (approved):
LGTM

enhancements/dismiss-issues/README.md
```diff
@@ -0,0 +1,192 @@
...
+The Hub will also have to dynamically create and append this DIF as the UI dismisses incidents. To achieve this, I propose a new field on incidents, called dismissed that is a boolean. This new incident update API will allow for the UI to tell the hub when an incident has been dismissed. When this happens the hub will need to:
+1. Move the incident to dismissed
+2. Add the incident to the either already existing DIF or create one and add it.
+
+For the Hub it may make sense that the DIF is actually a new table in the database, and it is generated on retrieval. I think this is an open question to be discussed with the Hub team.
+
+##### UI
+
+The UI will now have a new button to add to a given incident page. This button will allow the user to dismiss an incident. 
+The UI will also need to add a new page for a given application, that will allow you to see the "dismissed" incidents for that application.
```
jortel, line 163: Putting my user hat on, I don't know much about the rules but know that there are issues I want to dismiss when viewed for an application.  I would like to right-click on an issue (or a kebab item) and dismiss it.  Along the same lines, I would imaging a button that filters in/out issues (on the existing issues page) using the DIF to see/manage what has been dismissed.  So each issue, would have the option to "Dismiss" or if already dismissed, "Un-Dismiss".  Given, this , I don't think we'd need to manage the DIF separately as it oculd be derived from dismissed issues.

shawn-hurley: We have a page that has entry per DIF entry, that allows users to see the full dismissed incidents filter page that exists for the application. This is where kyou would download the DIF and be able to use locally.

The managing of Issues between states (dismissed/un-dismissed) would all happen in the Issues UI (via dynamic queries to the API, in UX/D it would be like adding filters)

Based on our conversations.

jortel (approved):
LGTM

enhancements/dismiss-issues/README.md
eemcmullan: No strong feelings

eemcmullan (approved):

enhancements/dismiss-issues/README.md
shawn-hurley: Updated the language on the enhancment. going to resolve.

Issue #194, opened 2024-03-19
[Feature] Support for Analyzing Multiple Inputs Applications Together


kthatipally:
Kantra currently can't analyze input multiple applications at once. Adding this feature would let users analyze several projects together, using the same settings and producing combined reports similar to how Windup supports analyzing multiple applications simultaneously.  This change would save time, ensure consistency, and give better insights.

rromannissen:
To be refined to follow @brunoborges' request by 0.6.

PR #198, opened 2024-08-14
:sparkles: Add kantra multiple inputs enhancement

aufi:
Adding multiple inputs support for kantra analyze command.

[Enhancement preview](https://github.com/konveyor/enhancements/blob/27eeaf1e9f2b4bc16c11265780e0aa9b308d6509/enhancements/kantra/multiple-inputs/README.md)

Fixes: https://github.com/konveyor/enhancements/issues/194

### DIFF
--- a/enhancements/kantra/multiple-inputs/README.md
+++ b/enhancements/kantra/multiple-inputs/README.md
@@ -1,0 +1,111 @@
+---
+title: kantra-analyze-multiple-inputs
+authors:
+  - "@aufi" (maufart)
+reviewers:
+  - "@dymurray"
+  - "@brunoborges"
+  - "@pranavgaikwad"
+  - "@rromannissen"
+  - "@shawn-hurley"
+approvers:
+  - TBD
+creation-date: 2024-08-13
+last-updated: 2024-08-14
+status: implementable
+see-also:
+  - "https://github.com/konveyor/enhancements/issues/194"  
+---
+
+# Kantra analyze multiple inputs
+
+This enhancement introduces feature of kantra to analyze multiple input applications with a single kantra command execution with results (issues&incidents) combined into a single static-report.
+
+## Release Signoff Checklist
+
+- [x] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] User-facing documentation is created
+
+## Open Questions
+
+- `--input-bin-dir` support and its relation to multiple `--input` args (or exclusivity of usage)
+
+## Summary
+
+Konveyor consists of the main cloud-native installation driven by _Hub_ that provides advanced features of application analysis execution including analysis task scheduling, parallel tasks executions. And more lighweight command line tool called _kantra_, that can analyze just a single input application on its command execution.
+
+Users community requested add support for multiple inputs to kantra, that would match to previous _mta-cli_ tool multiple inputs feature.
+
+## Motivation
+
+This enhancement should close a feature gap to "original" mta-cli multiple input option and increase usability of kantra tool.
+
+### Goals
+
+Analyze multiple applications with a single kantra command execution.
+
+### Non-Goals
+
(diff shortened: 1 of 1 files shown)


enhancements/kantra/multiple-inputs/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Open Questions
+
+- `--input-bin-dir` support and its relation to multiple `--input` args (or exclusivity of usage)
+
+## Summary
+
+Konveyor consists of the main cloud-native installation driven by _Hub_ that provides advanced features of application analysis execution including analysis task scheduling, parallel tasks executions. And more lighweight command line tool called _kantra_, that can analyze just a single input application on its command execution.
+
+Users community requested add support for multiple inputs to kantra, that would match to previous _mta-cli_ tool multiple inputs feature. In order to accomplish that, kantra will be updated to accept multiple ```--input``` args and handle analysis for all those inputs.
```
shawn-hurley, line 39: nit: wording, I think something more like `We want to to add support...` or something more like that?

enhancements/kantra/multiple-inputs/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Motivation
+
+This enhancement should close a feature gap to "original" mta-cli multiple input option and increase usability of kantra tool.
```
shawn-hurley, line 43: I wonder if we can by pass the feature parity argument and get to the motivation of the need for multiple inputs. Something like 

`Users have a large portfolio of applications, and running kantra on these individually is a time consuming process that leads them to have multiple different static reports. Having a single static report for all the local applications they are working on will greatly increase the visability of known issues across the local portfolio` 

Or something like that?

enhancements/kantra/multiple-inputs/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+### Implementation Details/Notes/Constraints
+
+### Security, Risks, and Mitigations
+
+There is a possibility to overload local machine (CPU or memory) that executes the analysis. That could be caused by size of the application(s) itself or chosen rules. User need to keep in mind, that kantra analyze execution might consume significant resources on their machine. 
+
+## Design Details
+
+It would be nice to generate static-report after each input application analysis completion to have some static-report available before finishing all inputs or for a case that analysis crash or an unexpected termination.
```
shawn-hurley, line 74: are we planning on looping over each input? what happens if the analysis command fails for B while running ABC. I think writing out the error scenarios here would be helpful.

shawn-hurley (changes requested):
I thnk that we need a little more info in the Design section to fully capture what we plan on doing. 

We also need to make sure that any of these changes is added to the this are incorporated in the enhancement from @eemcmullan. Can you two work together to make sure that happens?

enhancements/kantra/multiple-inputs/README.md
aufi: Good point, will update.

eemcmullan, line 74: One thing that may be good to discuss is how the providers will be handled. Will we create new provider containers for every input? Or check to see which are already running? That may get tricky with the shared volume, which is the input dir, and the shared network.

enhancements/kantra/multiple-inputs/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+### Limitations
+
+All analysis options are the same for all input applications and there is no plan to implement different source/targets/... for different input applications.
```
eemcmullan, line 78: Definitely something to document

enhancements/kantra/multiple-inputs/README.md
aufi: I was thinking about adding something like that the performance is not the main focus of this multiple inputs feature, so it might be OK re-create needed containers/networks etc. to not overcomplicate the codebase. But I'm not sure if this is the right approach.

eemcmullan (approved):

shawn-hurley (approved):

PR #199, opened 2024-08-15
:sparkles: Operator managed quota.

jortel:


enhancements/operator-quota/README.md
```diff
@@ -0,0 +1,70 @@
...
+to suggestions here). A percentage is used because it's a value that users and define
+without knowing anything about node configurations. A percentage will automatically
+adjust as nodes are added/removed or reconfigured.
+
+Quotas are routinely use by Development and QE and have been very effective at managing
+cluster resources and maximizing the Task manager scheduling features.
+
+## Design Details
+
+### Test Plan
```
aufi, line 68: Some kind of scale test should be probably defined to prove the quota helps handling higher loads.

enhancements/operator-quota/README.md
```diff
@@ -0,0 +1,70 @@
...
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] User-facing documentation is created
+
+## Summary
+
+Enhance the operator to optionally manage a ResourceQuota in the namespace.
+The user may optionally specify a percentage of cluster resources to be utilized by konveyor.
+This mainly applies to task pods.
```
aufi, line 31: Just curious about the `mainly` word :-) Most of the enhancements looks pretty generic to apply to all resources within the namespace.

Not a blocker, but maybe an example how the quota will stop tasks from being scheduled/overload the cluster, while the Hub happily runs, might be helpful.

aufi (commented):
Looks good, just added few notes.

enhancements/operator-quota/README.md
```diff
@@ -0,0 +1,70 @@
+---
+title: neat-enhancement-idea
```
aufi, line 2: Just a nit, probably `operator-managed-quota` instead.

enhancements/operator-quota/README.md
jmontleon: Also, do we have any upper / lower bounds? 99%, 100% would probably give basically no protection, while 1%, 2% etc. might make it just about impossible to schedule anything.

jortel: Perhaps but hard to say.  2% of a huge cluster may be enough.  On the flip side, 100% is essentially worthless but not broken.

jmontleon:
The logic to get available CPUs on worker nodes is something like this. This should be easy enough to reproduce with ansible.
```
$ tot_cpu=0; for n in $(oc get node --selector='node-role.kubernetes.io/worker=' -o name); do node_cpu=$(oc get $n -o jsonpath='{.status.capacity.cpu}'); ((tot_cpu+=node_cpu)); done; echo $tot_cpu
16
```

And memory (in Ki):
```
tot_mem=0; for n in $(oc get node --selector='node-role.kubernetes.io/worker=' -o name); do node_mem=$(oc get $n -o jsonpath='{.status.capacity.memory}'); ((tot_mem+=$(echo $node_mem | sed 's/Ki//') )); done; echo $tot_mem
65825548
```

enhancements/operator-quota/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+Maximize task manger scheduling features.
+Minimize clutter in the namespace.
+Be a better citizen on the cluster.
+
+### Non-Goals
+
+## Proposal
+
+Enhance the operator to optionally manage a Resource Quota in the namespace.
```
shawn-hurley, line 55: One thing that I would be concerned about, but I think there options around this, is this scenario:
1. Pods are created to MAX of the resource Quota
2. The HUB pod is killed and now is unable to be scheduled
3. This leads to konveyor being taken down for some period of time (as the pods complete it will free up space IIUC.

I think one thing to get around this, is have the hub/ui pods be of a higher priority class, and all the task pods be of another priority class, and use the resouce quota on just the task the pods using the resource quota scope of priority clas. 

Just a thought, and may be a pre-mature optimization but just a concern.

shawn-hurley (approved):
Overall I like the idea, I agree with others, that documentation around best practices and that it should probably default to off. 

I had one other concern, but I think it is managable, and may even be a concern for a different time.

enhancements/operator-quota/README.md
jortel: The quota will apply to the all pods but since the task pods are the most transient, it will k_ind-of_ apply to them the most :)  Unfortunately, quota's don't support label selector etc.  The only way to throttle only task pods would be to put them (and the quota) in a separate namespace.

aufi: Thanks for clarification!

enhancements/operator-quota/README.md
jortel: I agree with this concern.  Unfortunately priority classes are cluster scoped and without a default priority class (which does not exist by default), pods without an assigned class default to Zero(0).  IIUC, this means that tackle pods will run at higher priority than all other pods in the cluster (including controllers).  Also, our operator would need perms to create the cluster-scoped priority classes which I highly doubt could be acceptable.  All of this adds up to - we can't use/rely-on priority classes.

```
$ oc get pc
NAME                      VALUE        GLOBAL-DEFAULT   AGE
system-cluster-critical   2000000000   false            33d
system-node-critical      2000001000   false            33d
```
It is also unfortunate that Quota's don't support labels selectors.

The only option I have considered to address this concern is for task pods (and quota) to be created in another namespace.  Example: _konveyor-task_.  This would have the benefits of:
- Would eliminate clutter in our _main_ in konveyor-tackle.
- Quota would only affect task pods.

But with the following drawbacks:
- mainly that dealing with multiple namespaces would be kind of a pain.  And, perhaps confusing for users.

shawn-hurley: I think that if the other suggestions don't work then as described it makes sense, sorry for the wild goose chase. Just thought there as maybe something that we could do.

With that said the usecase I am worried about is an edge case and baring good options (1-3 I agree are out) and I don't want to re-create a scheduler (option 4) then I think what we have is sufficient

jortel: Well, we already have an intelligent scheduler.  Option 4 only proposed to make it a little smarter.

enhancements/operator-quota/README.md
jortel: > Some kind of scale test should be probably defined to prove the quota helps handling higher loads.

This would be very complicated to automate.

aufi (approved):

PR #204, opened 2024-08-24
:sparkles: dotnet provider treesitter

djzager:


jortel:
This seems very promising.

aufi (commented):
Looks good to me :+1: 

(I don't have good .net analyzer background, so giving just +1)

djzager:
Sticking this in here for my future reference https://chatgpt.com/share/3fd9d6b9-eb81-4fd6-9fa2-427b5f527924

shawn-hurley (commented):
I would like to understand how we plan on also finding issues in dependencies. 

If we currently can't do that, and this is not worse then that is ok, but we should at least think about this.

Is TreeSitter able to generate an AST from a .dll library?

djzager:
> If we currently can't do that, and this is not worse then that is ok, but we should at least think about this.

We can't do this currently...or at least we don't do this currently.

> Is TreeSitter able to generate an AST from a .dll library?

I don't believe so, it looks like we can generate an AST for each file individually. So a dll would need to be unpacked first.

shawn-hurley:
That makes sense, I was just wondering. If this is not worse than before then I don't see a downside. I do want to have a plan, If we do make sure that we can unpack .DLL's and get the AST from them, then I will be completely satisfied that we can achieve what we need to in the future.

Issue #186, opened 2024-07-08
[RFE] Allow the association of target platforms with Archetypes


rromannissen:
### What is the problem?

There is no way for an organization to enforce certain migration paths for applications associated with a certain archetype to land on specific target platforms. This allows migrators to explore migration paths that are not sanctioned by the organization, opening up the door for runtime/platforms and technology stacks that don't meet the corporate policy.

### Why is this a problem?

After interviewing several large organizations, it has been revealed that modernization projects are perceived by them as an opportunity to further standardize the application landscape, narrowing down the number of allowed archetypes, technology stacks and runtime/platforms to a few supported options. For MTA to be useful on that approach at scale, there needs to be a way for organizations to model the supported runtimes/platforms and technology stacks and associate them with archetypes, and from there, streamline and simplify the analysis process for migrators.

### Proposed solution

- The concept of _Target Platform_ will be introduced to model a potential runtime/platform for applications that are associated to a given archetype. 
- For each _Target Platform_ associated to a given archetype, users should be able to associate one or many migration paths.
    - _Target Platforms_ could have different sets of migration paths for different archetypes. For example, the "Kubernetes" _Target Platform_ could include the _OpenJDK 21_, _EAP 8_ and _Containerization_ migration paths for the "EAP 6" archetype, and then include the _OpenJDK 21_ and _Containerization_ migration targets for the "Spring Boot" archetype.

![Screenshot from 2024-07-08 15-52-08](https://github.com/konveyor/enhancements/assets/15966456/c0d8fe5b-c66f-40c0-bc6d-959360c4ce51)

- These relationships could be used to enforce certain migration motions for migrators on the analysis wizard or external systems like the CLI, IDE plugins and potential future integrations with Backstage.

### Open Questions

- Should this relationship be managed in the _Migration_ or the _Administration_ perspective?

sjd78:
Some of my initial thoughts:

  - Migration Paths seem like another way to describe a migration target, which really is a ruleset target label. For example, the "Application server migration to" target card with "JBoss EAP 8" selected ultimately enables the label `konveyor.io/target=eap8` on an analysis.

  - Target Platform looks like a collection of target labels.  Following the example in the diagram, Kubernetes could be a set of labels ` konveyor.io/target=cloud-readiness`, `konveyor.io/target=openjdk21`, and `konveyor.io/target=eap8`.

  - Archetypes would just need to be extended to support a collection of Target Platforms.

[...]

PR #206, opened 2024-08-30
:sparkles: Add target platforms to application archetypes

sjd78:
Resolves: #186

### DIFF
--- a/enhancements/archetype-target-platforms/README.md
+++ b/enhancements/archetype-target-platforms/README.md
@@ -43,26 +43,26 @@ N/A
 
 ## Summary
 
-Application archetypes will be enhanced to be able to store groups of migration
-targets together as a __target platform__.  The __target platforms__ will only
+Application archetypes will be enhanced to be able to store groups of analysis
+target labels together as a __target platform__.  The __target platforms__ will only
 exist on the archetype and are not intended to be first class entities.
 
 Various tools that generate analysis tasks will be able to use the target platforms
-from the application's archetype to inform what analysis targets should be applied
+from the application's archetype to inform what analysis target labels should be applied
 without requiring a user to select them individually.
 
 This enhancement only calls for adding CRUD functionality to the archetype edit screen
-for target platforms as groupings of known analysis targets.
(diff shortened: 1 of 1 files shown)


enhancements/archetype-target-platforms/README.md
```diff
@@ -0,0 +1,215 @@
...
+      "name": "Kubernetes",
+      "targets": [
+        "konveyor.io/target=cloud-readiness",
+        "konveyor.io/target=openjdk21",
+        "konveyor.io/target=eap8"
+      ]
+    },
+    {
+      "name": "Traditional Standalone",
+      "targets": [
```
shawn-hurley, line 148: For future proof, can we think about naming these rulesets or labels?

shawn-hurley (approved):
Just a small naming concern, just thinking about the future of more and more rulesets with more and more labels.

enhancements/archetype-target-platforms/README.md
jortel: We already have a first-class, user facing concept (resource) of a `Target` which is just a named collection of labels.  The goal seems to be to pre-select one or more of these targets when in the analysis wizard using matched archetypes. The PR/issue suggests the selection could be locked as well as a constraint on the user.  Given this, it seems like _targets_ should be a list of references (id) to existing `Target` resources.

jortel: After discussion with Scott whereby users want to build the target platform with specific versions of targets, I think we need to define this with labels.  Given this, I would recommend that "labels" would be better than "targets".

jortel (approved):
LGTM

enhancements/archetype-target-platforms/README.md
sjd78: I changed the sample json prop to "labels".  I also updated the language in the rest of the enhancement to explicitly say "target labels".  This should make it more obvious the intent is to use the specific label within a target in the same way the analysis wizard shows all the targets as cards with a dropdown to select the label, but then only sends the specific labels to start the analysis.

PR #214, opened 2025-01-20
[DRAFT] Create Asset Generation CLI Enhancement

JonahSussman:


enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,205 @@
...
+Organizations migrating workloads from Cloud Foundry to Kubernetes require
+reliable tools to simplify resource discovery and deployment transformation.
+Existing tools like Move2Kube (M2K) provide limited flexibility and lack a
+modular approach, resulting in suboptimal user experiences. This enhancement
+addresses these gaps by extending the Kantra CLI with commands for discovery and
+asset generation while supporting a canonical configuration format for future
+expansion.
+
+We propose modifying the existing Kantra CLI instead of developing an entirely
+new CLI to ensure a smoother adoption process for users. Leveraging Kantra's
```
gciavarrini, line 55: For phase 1, integration into existing MTA/Kantra CLI was set as non-goal. The idea, if I'm not wrong, is to be independent to other projects to deliver faster the 1st phase.
Do you think that having a simple CLI for the phase 1 will overcomplicate a future integration?

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,205 @@
...
+modularity and maintainability, we will create a standalone library of
+functionality to which the CLI will serve as a frontend, allowing for future
+integration with other tools and workflows.
+
+### Goals
+
+- Create a standalone library for integration with other tools and workflows.
+- Extend Kantra to support:
+  - `discover` command: Output a YAML/JSON representation of Cloud Foundry resources.
+  - `generate` command: Produce Helm charts for deployment to Kubernetes.
```
gciavarrini, line 68: I think `generate` can produce `values.yaml` (generated from the canonical form) and an example helm template. That helm template will need to be tuned by the user.
It's too complex to generate decent helm charts.

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,205 @@
...
+        application name, instances, routes), runtime information (e.g., memory,
+        disk), and deployment settings.
+    - **Error Handling:**
+      - Detect and report missing or malformed input files with detailed error messages.
+      - Provide retry options for transient errors during discovery.
+
+  - `generate`: This command takes a canonical representation and a Helm
+    template to produce deployment-ready Helm charts. Details:
+    - **Flags and Options:**
+      - `--input=<file>` (required): Specifies the canonical configuration file
```
gciavarrini, line 124: I'd like to have the option to provide a CF manifest as input. IMO, This will be useful if a user provides the CF manifest directly, for testing purposes and during the development.

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,205 @@
...
+      - Applies the provided Helm template to generate deployment artifacts for
+        Kubernetes.
+      - Validates the generated artifacts for completeness and correctness.
+    - **Error Handling:**
+      - Log errors when applying Helm templates.
+      - Support a dry-run mode to identify potential issues without generating artifacts.
+
+- **Library Architecture:**
+  - Implement the core functionality as a standalone library in Go for reuse in
+    Kantra and other projects.
```
gciavarrini, line 144: Can you clarify this? It seems different from the `Motivation` section

gciavarrini (commented):
I took a look at this draft PR to start getting familiar with your idea for implementing the CLI. Some questions 😀

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,205 @@
...
+      - `--output-dir=<directory>` (optional): Directory to save the generated
+        Helm chart. Defaults to the current directory.
+      - `--type=<type>` (optional): Specifies the type of generator. MVP
+        supports helm.
+    - **Behavior:**
+      - Reads the canonical representation file to extract configuration
+        details.
+      - Applies the provided Helm template to generate deployment artifacts for
+        Kubernetes.
+      - Validates the generated artifacts for completeness and correctness.
```
jordigilh, line 137: Can you specify that the validation is done for non-yaml/json files? Not the ones in the helm template directory?

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,205 @@
...
+    - **Error Handling:**
+      - Log errors when applying Helm templates.
+      - Support a dry-run mode to identify potential issues without generating artifacts.
```
jordigilh, line 140: Do we need to allow server side checks when generating non-yaml/json templates, such as `Dockerfile` or `EAP xml` configuration files?

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,205 @@
...
+      - `--type=<type>` (optional): Specifies the type of generator. MVP
+        supports helm.
+    - **Behavior:**
+      - Reads the canonical representation file to extract configuration
+        details.
+      - Applies the provided Helm template to generate deployment artifacts for
+        Kubernetes.
+      - Validates the generated artifacts for completeness and correctness.
+    - **Error Handling:**
+      - Log errors when applying Helm templates.
```
jordigilh, line 139: What do you mean by `applying Helm templates`? Do you think the CLI should render the helm templates? or just render the non-yaml/json templates?

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,205 @@
...
+2. Modularity increases the maintenance burden.
+   - **Mitigation:** Maintain clear documentation and well-defined APIs for each module.
+
+## Design Details
+
+### Test Plan
+
+1. Unit tests for the discover and generate commands.
+2. Integration tests for end-to-end workflows (Cloud Foundry -> canonical -> Helm charts).
+3. Manual validation with sample Cloud Foundry applications.
```
jordigilh, line 167: nitpick: I'd suggest to replace this with CI/CD automated E2E tests. Manual validation is never a good idea long term.

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,205 @@
...
+  - Release as part of upcoming Kantra/Konveyor in tech preview.
+
+### Security, Risks, and Mitigations
+
+- Conduct code reviews with security specialists to mitigate risks.
+- Validate CLI outputs to prevent injection vulnerabilities in generated
+  artifacts.
+
+#### **Risks**
+1. Generated artifacts may not match user expectations.
```
jordigilh, line 156: Note: I expect the CLI to generate the `values.yaml` based on the canonical form. The integration between the contents of the `values.yaml` and the templates should fall into the user's responsibility. Ours should be documenting how the canonical form manifest from the discovery is translated into the `values.yaml` and how to run the `generate` command that outputs the helm chart.

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,205 @@
...
+- **Library Architecture:**
+  - Implement the core functionality as a standalone library in Go for reuse in
+    Kantra and other projects.
+  - Modularize the functionality to support future enhancements. 
+- **Integration Path:**
+  - Release as part of upcoming Kantra/Konveyor in tech preview.
+
+### Security, Risks, and Mitigations
+
+- Conduct code reviews with security specialists to mitigate risks.
```
jordigilh, line 151: Should this section include access to CF platform credentials and RBAC access rights?

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,205 @@
...
+making it easy to deploy on Kubernetes with minimal manual intervention.
+
+### Implementation Details
+
+- **Command Design:**
+  - `discover`: This command is responsible for analyzing an application from a
+    specified platform and generating a canonical representation in YAML/JSON
+    format. Details:
+    - **Flags and Options:**
+      - `--platform=<platform>` (required): Specifies the platform to discover.
```
jordigilh, line 105: Do we need to also include fields for credentials to access the CF platform? as well as the URL for the REST API server?

jordigilh: We're currently working under the assumption that the CF manifest files are accessible on the local file system. I think that option might be best for MVP at this point, later on being enhanced to support other options, such as REST API. WDYT?

savitharaghunathan: @jordigilh can you elaborate on what do you mean by - CF manifest files are accessible on the local file system. we still retrieve that from the CF platform?

enhancements/assets-generation/cli/README.md
JonahSussman: The approach I'm envisioning is to integrate with the existing Kantra from the beginning to ensure we can deliver value to customers more quickly, as they already will have Kantra vetting in their security processes. 

The core functionality will be implemented as a standalone Go library, which will serve as the foundation for this enhancement. This ensures that the Kantra integration remains a thin frontend layer while keeping the library modular and reusable for future projects.

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,205 @@
...
+        Supported value for MVP: cf.
+      - `--output=<file>` (optional): Writes the output to a specified file.
+        Defaults to standard output.
+      - `--verbose` (optional): Provides additional details about the discovery
+        process in the logs.
+    - **Behavior:**
+      - Scans the provided application input (e.g., Cloud Foundry manifest
```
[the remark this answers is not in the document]
JonahSussman: For the MVP, we plan to support both fetching Cloud Foundry application manifests from a file system location (e.g., a cloned git repo) and fetching manifests directly from a CF platform via the REST API. That said, if supporting both approaches in the initial release adds too much complexity, we can prioritize one and revisit the other in a subsequent phase

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,237 @@
...
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] User-facing documentation is created
+
+
+## Open Questions
+1. Do we need an interactive mode where users can preview and modify
+   configurations during discovery or generation?
+2. Should the CLI provide validation against Kubernetes best practices
```
eemcmullan, line 31: I am +1 for this maybe in a later phase?

eemcmullan, line 31: We currently have the yaml provider in the analyzer that has this capability. Maybe that could somehow be utilized to achieve this validation?

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,237 @@
...
+overall workflow for existing users already familiar with the tool. To achieve
+modularity and maintainability, we will create a standalone library of
+functionality to which the CLI will serve as a frontend, allowing for future
+integration with other tools and workflows.
+
+### Goals
+
+- Create a standalone library for integration with other tools and workflows.
+- Extend Kantra to support:
+  - `discover` command: Output a YAML/JSON representation of Cloud Foundry resources.
```
eemcmullan, line 67: Should discovery be used for gathering platform info rather than output? It mentions here output of CF resources, but maybe that could be another step after discovery? (e.g. discovery -> runOutPut(?) -> generate). wdyt?

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,237 @@
...
+### Implementation Details
+
+- **Command Design:**
+  - `discover`: This command is responsible for analyzing an application from a
+    specified platform and generating a canonical representation in YAML/JSON
+    format. Details:
+    - **Flags and Options:**
+      - `--platform=<platform>` (required): Specifies the platform to discover.
+        Supported value for MVP: cf.
+      - `--application=<name>` (optional): Specifies the application to be analyzed.
```
eemcmullan, line 109: To keep consistency with the analyzer/kantra, we could call this flag `--input`.

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,237 @@
...
+making it easy to deploy on Kubernetes with minimal manual intervention.
+
+### Implementation Details
+
+- **Command Design:**
+  - `discover`: This command is responsible for analyzing an application from a
+    specified platform and generating a canonical representation in YAML/JSON
+    format. Details:
+    - **Flags and Options:**
+      - `--platform=<platform>` (required): Specifies the platform to discover.
```
eemcmullan, line 107: It may be helpful to have another optional flag such as `--list-platforms` that lists ones we support. Maybe this is for the future, when we move away from only CF.

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,237 @@
...
+        Supported value for MVP: cf.
+      - `--application=<name>` (optional): Specifies the application to be analyzed.
+      - `--use-agent=<agent_name>` (optional): Enables agent-based discovery for extracting data.
+      - `--use-live-connection` (optional): Uses live platform connections for real-time discovery.
+      - `--output=<file>` (optional): Writes the output to a specified file.
+        Defaults to standard output.
+      - `--verbose` (optional): Provides additional details about the discovery
```
eemcmullan, line 114: NIT: like above, if integrating with kantra, could call this `--log-level`.

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,237 @@
...
+      - `--type=<type>` (optional): Specifies the type of generator. MVP
+        supports helm.
+    - **Behavior:**
+      - Reads the canonical representation file to extract configuration
+        details.
+      - Applies the provided Helm template to generate deployment artifacts for
+        Kubernetes.
+      - Validates the generated artifacts for completeness and correctness.
+    - **Error Handling:**
+      - Log errors when applying Helm templates.
```
eemcmullan, line 171: It may be nice send helm-specific logs/errors to a log, and keep the cli errors separate.

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,237 @@
...
+   configurations during discovery or generation?
+2. Should the CLI provide validation against Kubernetes best practices
+   (e.g., resource limits, liveness probes) during artifact generation?
+
+## Summary
+
+This proposal introduces enhancements to the existing Kantra tool to support the
+discovery and transformation of resources from Cloud Foundry to
+Kubernetes. The goal is to create a modular CLI that outputs a canonical
+representation of resources and generates deployment artifacts (e.g., Helm
```
jortel, line 39: s/artifacts/assets

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,237 @@
...
+- Utilize the [canonical configuration enhancement](link: TODO) to serve as a consistent
+  intermediary for platform transformations.
+- Enhance logging and error reporting for better user experience and debugging.
+- Provide detailed documentation and examples for both commands to facilitate adoption.
+
+### Non-Goals
+
+- Handling of sensitive resource data, such as secrets, is explicitly out of
+  scope for this CLI. Instead, we will rely on Konveyor's secure encryption
+  stores, which users already use with Konveyor.
```
jortel, line 78: Encryption should be outside the scope of the CLI.  However, the CF provider in the CLI is the only thing that understands CF and can identify which values are sensitive.  Thinking the canonical manifest (schema) should have a `secrets` section.  When the CLI identifies a sensitive value, it would add an entry to the `secrets` section containing the sensitive data.  The sensitive value would be replace with a reference to the secret. 

 For example (something like):
CF manifest
```
password: ABC
```
canonical manifest:
```
application:
  password: $(password)
secrets:
 - name: password
 - value: ABC
```
When the manifest is stored in konveyor, the secret values could be encrypted.

Naming is for example.

enhancements/assets-generation/cli/README.md
jordigilh: @savitharaghunathan we're currently working under the assumption (@gciavarrini and me) that for the first release, the MVP for CF is to assume the CF manifests are located on the file system to simplify the scope of work, since querying CF server requires an additional layer of complexity, e.g.: setting up a CF platform instance or equivalent, implement the REST API interface and filter out the output, since CF augments the manifest with fields that are not found in the CF manifest spec but are included as part of the REST API, from what we've seen. 

As @JonahSussman suggested, REST API retrieval is still planned but in terms of priority it will expedite things if we don't have to deal with CF REST API for the first iteration. WDYT?

enhancements/assets-generation/cli/README.md
jordigilh: If we provide structured logs, that should be an easy task to filter.

enhancements/assets-generation/cli/README.md
savitharaghunathan: is there a reason we would need an intermediate step? Right now the discovery command is defaulting to standard out and provides an option to write to a file.

enhancements/assets-generation/cli/README.md
savitharaghunathan: I like this idea. +1

enhancements/assets-generation/cli/README.md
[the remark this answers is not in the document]
savitharaghunathan: My understanding is that we would be implementing the second option. I can capture this info in the enhancement

enhancements/assets-generation/cli/README.md
savitharaghunathan: added the suggestion

enhancements/assets-generation/cli/README.md
JonahSussman: Upon further evaluation, I'm thinking that validation of generated artifacts might be out of scope for the CLI. The burden of making sure the artifacts work will be on the the generator. WDYT?

enhancements/assets-generation/cli/README.md
JonahSussman: I don't think I understand. Could you phrase it differently?

enhancements/assets-generation/cli/README.md
JonahSussman: The idea is that the Kantra CLI will use the library, but other projects can use the library as well if they want to

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,243 @@
...
+## Summary
+
+This proposal introduces enhancements to the existing Kantra tool to support the
+discovery and transformation of resources from Cloud Foundry to
+Kubernetes. The goal is to create a modular CLI that outputs a canonical
+representation of resources and generates deployment assets (e.g., Helm
+charts). The enhancement will provide a foundational capability for platform
+migrations, enabling a tech-preview release in an upcoming version of Kantra and
+integration into downstream workflows.
+
```
istein1, line 45: @JonahSussman ,
How would that scale in case we have many applications with same/overlapping run time requirements?
Also. would we check on Kubernetes side whether the required resources are there, and in this case no asset generation will be required?

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,243 @@
...
+functionality to which the CLI will serve as a frontend, allowing for future
+integration with other tools and workflows.
+
+### Goals
+
+- Create a standalone library for integration with other tools and workflows.
+- Extend Kantra to support:
+  - `discover` command: Output a YAML/JSON representation of Cloud Foundry resources.
+  - `generate` command: Produce Helm charts for deployment to Kubernetes. For the mvp,
+    the we will generate a values.yaml file based on the CF discovery manifest,
```
istein1, line 71: remove 'the'

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,243 @@
...
+   - **Mitigation:** Maintain clear documentation and well-defined APIs for each module.
+
+## Design Details
+
+### Test Plan
+
+1. Unit tests for the discover and generate commands.
+2. Integration tests for end-to-end workflows (Cloud Foundry -> canonical -> Helm charts).
+3. Automated validation with sample Cloud Foundry applications.
+
```
istein1, line 206: 4. Test each of the components separately: Discovery, Agent and Generate.

enhancements/assets-generation/cli/README.md
savitharaghunathan: @istein1 We will be mapping the canonical values to the helm template provided by the user. Any validation and verification of generated assets/manifests are out of scope for this MVP.

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,243 @@
...
+### Non-Goals
+
+- Handling of sensitive resource data, such as secrets, is explicitly out of
+  scope for this CLI. Instead, we will rely on Konveyor's secure encryption
+  stores, which users already use with Konveyor.
+- Automatic detection of application archetypes.
+- Full compatibility with the Konveyor add-on process.
+- Immediate support for platforms beyond Cloud Foundry and Kubernetes.
+- Actual agent implementation. we will have stubs for the `agent` 
+  subcommand for initial mvp.
```
nachandr, line 87: This looks like an oversight since `agent` is a command and not a sub command .

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,243 @@
...
+expansion.
+
+We propose modifying the existing Kantra CLI instead of developing an entirely
+new CLI to ensure a smoother adoption process for users. Leveraging Kantra's
+established user base allows for quicker buy-in and ensures users have an
+enhanced tool in their hands sooner. Additionally, this approach simplifies the
+overall workflow for existing users already familiar with the tool. To achieve
+modularity and maintainability, we will create a standalone library of
+functionality to which the CLI will serve as a frontend, allowing for future
+integration with other tools and workflows.
```
nachandr, line 63: Hi @JonahSussman , @savitharaghunathan , What CLI would the new standalone library for asset generation and platform awareness work with ? Containerized CLI or Containerless CLI ?

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,243 @@
...
+1. Modularity increases the maintenance burden.
+   - **Mitigation:** Maintain clear documentation and well-defined APIs for each module.
+
+## Design Details
+
+### Test Plan
+
+1. Unit tests for the discover and generate commands.
+2. Integration tests for end-to-end workflows (Cloud Foundry -> canonical -> Helm charts).
+3. Automated validation with sample Cloud Foundry applications.
```
nachandr, line 205: Hi @JonahSussman . @savitharaghunathan , Could you please clarify what exactly would be validated ?
Would it  be the generated artifacts for the CF applications for completeness and correctness ?

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,243 @@
...
+
+## Summary
+
+This proposal introduces enhancements to the existing Kantra tool to support the
+discovery and transformation of resources from Cloud Foundry to
+Kubernetes. The goal is to create a modular CLI that outputs a canonical
+representation of resources and generates deployment assets (e.g., Helm
+charts). The enhancement will provide a foundational capability for platform
+migrations, enabling a tech-preview release in an upcoming version of Kantra and
+integration into downstream workflows.
```
dymurray, line 44: Remove mention about downstream

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,243 @@
...
+
+## Motivation
+
+Organizations migrating workloads from Cloud Foundry to Kubernetes require
```
dymurray, line 48: Would prefer this to mention platform awareness as a concept and not specifically this migration path; but this migration path can be used as an example

enhancements/assets-generation/cli/README.md
dymurray: This does not use analysis; so it's not a matter of containerless vs container CLI in this context. These are new commands.

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,243 @@
...
+overall workflow for existing users already familiar with the tool. To achieve
+modularity and maintainability, we will create a standalone library of
+functionality to which the CLI will serve as a frontend, allowing for future
+integration with other tools and workflows.
+
+### Goals
+
+- Create a standalone library for integration with other tools and workflows.
+- Extend Kantra to support:
+  - `discover` command: Output a YAML/JSON representation of Cloud Foundry resources.
```
dymurray, line 69: Again, please ensure we are delineating the command away from the provider. So it would be `discover --provider pcf` or whatever.

dymurray, line 69: Just an addition that this should default to YAML; since we are targeting helm/ansible as the target generator

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,243 @@
...
+  - `generate` command: Produce Helm charts for deployment to Kubernetes. For the mvp,
```
dymurray, line 70: Based on my new understanding of how this works; I wonder if we should instead call this `render`? Under the hood we are essentially passing in a location to a helm chart and rendering the templates. I don't know; I originally liked `generate`

enhancements/assets-generation/cli/README.md
dymurray: and please generalize away from CF

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,243 @@
...
+    the we will generate a values.yaml file based on the CF discovery manifest,
+     and combine it with the user's templates to generate a helm chart
+- Utilize the [canonical configuration enhancement](link: TODO) to serve as a consistent
```
dymurray, line 73: The more I understand about this; there is no "canonical config". `discover` spits out a `values.yaml` and the user puts that into their helm chart; then calls `generate` with a path to a helm chart where the templates get rendered.

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,243 @@
...
+        and deployment settings.
+      - **Error Handling:**
+        - Detect and report missing or malformed input files with detailed error messages.
+        - Provide retry options for transient errors during discovery.
+      - **Discovery Methods:**
+        - Via live connections using APIs or other supported interfaces.
+        - Through filesystem access to the platform's installation paths (e.g., for Application Servers
+          or Servlet Containers), implemented as an agent on the host.
+          
+  - `agent`: This command provides management capabilities for platform agents.
```
dymurray, line 136: Where is this coming from? What is an agent?

enhancements/assets-generation/cli/README.md
savitharaghunathan: This section has been removed and will be added back when we implement agents

enhancements/assets-generation/cli/README.md
savitharaghunathan: updated. PTAL

enhancements/assets-generation/cli/README.md
savitharaghunathan: updated. PTAL

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,243 @@
...
+   configurations during discovery or generation?
+2. Should the CLI provide validation against Kubernetes best practices
+   (e.g., resource limits, liveness probes) during artifact generation?
+   Note: yaml provider in the analyzer that has this capability.
+   We could leverage that if needed.
+
+## Summary
+
+This proposal introduces enhancements to the existing Kantra tool to support the
+discovery and transformation of resources from Cloud Foundry to
```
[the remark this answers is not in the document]
savitharaghunathan: updated. PTAL

enhancements/assets-generation/cli/README.md
savitharaghunathan: Validation of generated assets is out of scope and I have removed it from the test plan

enhancements/assets-generation/cli/README.md
savitharaghunathan: updated

enhancements/assets-generation/cli/README.md
savitharaghunathan: updated. PTAL

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,221 @@
...
+        - Detect and report missing or malformed input files with detailed error messages.
+        - Provide retry options for transient errors during discovery.
+      - **Discovery Methods:**
+        - Via live connections using APIs or other supported interfaces.
+        - Through filesystem access to the platform's installation paths (e.g., for Application Servers
+          or Servlet Containers), implemented as an agent on the host.
+          
+  - `generate`: This command takes a canonical representation and a Helm
+    template to produce deployment-ready Helm charts. Details:
+    - **Flags and Options:**
```
rromannissen, line 137: I would add some additional options here. I just want to convey the functionality, I'm open to changes in names and format:

- `--set key=value`: Sets a value for the given in the command line. If the key already exists in the values file that was passed, it will override its value.
- `--non-k8s-only`: Only renders the non-k8s templates available in the files/konveyor path from the target Helm Chart.

enhancements/assets-generation/cli/README.md
rromannissen: @jordigilh that would mean writing some custom validators. XML should be quite straightforward as we have XSD schemas, but other files such a Dockerfile would be a bit trickier. Sounds like a nice to have for further iterations, but I'd move forward without it for the moment.

rromannissen (changes requested):

enhancements/assets-generation/cli/README.md
savitharaghunathan: > --set key=value: Sets a value for the given in the command line. If the key already exists in the values file that was passed, it will override its value.

@rromannissen - this will override the value for the key in discovery manifest, which in turn will override the values.yaml file in the given helm template. is that correct?

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,225 @@
...
+
+- **Command Design:**
+  - `discover`: This command is responsible for analyzing an application from a
+    specified platform and generating a canonical representation in YAML/JSON
+    format. Details:
+    - **Flags and Options:**
+      - `--platform=<platform>` (required): Specifies the platform to discover.
+        Supported value for MVP: cf.
+      - `--input=<name>` (optional): Specifies the application to be analyzed.
+      - `--use-agent=<agent_name>` (optional): Enables agent-based discovery for extracting data.
```
gciavarrini, line 113: @savitharaghunathan since we (temporary) dropped the agent part, is it necessary to remove also this option?

enhancements/assets-generation/cli/README.md
gciavarrini: you're right, i misread the option description: let's ignore my comment :)

enhancements/assets-generation/cli/README.md
rromannissen: Not really. The discovery manifest and the values.yaml file would remain the same, but when rendering the value of the given key would be overridden by the given value. This is similar to the --set flag from the `helm template` command.

savitharaghunathan: I updated the flags. Thanks @rromannissen for the clarification

enhancements/assets-generation/cli/README.md
savitharaghunathan: did we decide on the naming? @dymurray

enhancements/assets-generation/cli/README.md
savitharaghunathan: updated

enhancements/assets-generation/cli/README.md
rromannissen: @jordigilh @gciavarrini is this something we can add to #215 ? Have we identified anything sensitive in the CF spec that should be stored encrypted in the hub in subsequent iterations?

enhancements/assets-generation/cli/README.md
rromannissen: There shouldn't be a need for the user to _put_ the values file in the chart. That should be an input parameter for the CLI, and it will make sure it is passed correctly to the templating engine. Users shouldn't have to move files around for this to work.

rromannissen (approved):
LGTM after the changes, I think we are ready to start implementation.

enhancements/assets-generation/cli/README.md
jordigilh: CF app manifest exposes the docker username as a field in the spec. I would consider that to be sensitive data. Right now we're capturing it as a field in the canonical form, but we can follow on what @jortel suggest and store it in a separate section, although it would be stored in a clear format. Let me know if that works for you and I'll update both the logic in the discovery and the enhancement.

Throughout our research on CF, we've seen across a few examples where credentials (username + password) were injected as a [environment variables](https://docs.cloudfoundry.org/devguide/deploy-apps/environment-variable.html#view-env) in the parameter field of the service in free text format (k/v) for registering services. The value can be structured (json) but it's dependant on the target service and we would not be able to parse and extract the information in all cases. I would suggest to let the users clean up the values after the canonical form is generated to suit their deployment needs for this case, since the service registration in CF does not follow the same pattern as in K8S services.

enhancements/assets-generation/cli/README.md
dymurray: No, no containers should be executed. Sorry for the confusion

enhancements/assets-generation/cli/README.md
dymurray: `discovery manifest` I think is what we agreed upon

savitharaghunathan: @dymurray ack. For the command, are we good with `generate` ?

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,270 @@
...
+        to use as input.
+      - `--template=<path>` (required): Path to the Helm template to use for
+        chart generation.
+      - `--output-dir=<directory>` (optional): Directory to save the generated
+        Helm chart. Defaults to stdio.
+      - `--type=<type>` (optional): Specifies the type of generator. MVP
+        supports helm. Defaults to `Helm`
+      - `--set key=value` (optional): Sets a value for the given in the command line.
+        If the key already exists in the values file that was passed, it will override its value.
+      - `--non-k8s-only` (optional): Only renders the non-k8s templates available in the files/konveyor
```
jordigilh, line 147: How does this flag work when the type generator is not helm? We know that for helm we will have to render the files from a directory inside the helm chart, but is that also applicable for other template engines? To me this flag is helm specific and it's being exposed at the same level as the template engine flag. 
Maybe the type value should be a command instead of a flag?
```
kantra generate helm --template ./ --output-dir /tmp --non-k8s-only
```

and then if we were to use ansible:
```
kantra generate ansible --playbook ./ --output-dir /tmp 
```

It looks cleaner since the flags specific for helm (`template`) are not applicable to ansible (`playbook` for instance).
WDYT?

enhancements/assets-generation/cli/README.md
jordigilh: Agreed

enhancements/assets-generation/cli/README.md
jordigilh: Sounds good.

enhancements/assets-generation/cli/README.md
jordigilh: Sounds good.

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,270 @@
...
+making it easy to deploy on Kubernetes with minimal manual intervention.
+
+### Implementation Details
+
+- **Command Design:**
+  - `discover`: This command is responsible for analyzing an application from a
+    specified platform and generating a canonical representation in YAML
+    format. Details:
+    - **Flags and Options:**
+      - `--platform=<platform>` (required): Specifies the platform to discover.
```
jordigilh, line 110: Same thing here: PCF applications might require to pass specific arguments to enable discovery (e.g. performing discovery over live platform). I suggest to turn this flag as a subcommand to allow specifying flags per platform:

```bash
$> kantra discover cloud-foundry --use-live-connection --username <username> --password <password> --server-url https://localhost:8443
```

Common flags can remain as part of the discovery command, but specific ones will fall into each sub-command/platform.

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,270 @@
...
+  ...
+  ...
+
+Run 'kantra COMMAND --help' for more information on a command.
+
+Discover Command:
+  Usage: kantra discover [OPTIONS]
+
+  Options:
+    --platform <platform>      Specify the platform to discover (e.g., cloudfoundry).
```
nachandr, line 188: @dymurray @rromannissen @savitharaghunathan @JonahSussman , What are your thoughts on calling this option 'provider' instead of 'platform' ?

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,270 @@
...
+    --input <name>             Specify the application to analyze.
+    --use-live-connection      Use live platform connections for discovery.
+    --output <file>            Output file (default: standard output).
+    --list-platforms           List available discovery providers.
```
nachandr, line 192: Same comment as before 

Could we call this option 'list-providers' instead of 'list-platforms'  ?

enhancements/assets-generation/cli/README.md
dymurray: Yes!

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,270 @@
...
+          
+  - `generate`: This command takes a canonical representation and a Helm
+    template to produce deployment-ready Helm charts. Details:
+    - **Flags and Options:**
+      - `--input=<file>` (required): Specifies the canonical configuration file
+        to use as input.
+      - `--template=<path>` (required): Path to the Helm template to use for
+        chart generation.
+      - `--output-dir=<directory>` (optional): Directory to save the generated
+        Helm chart. Defaults to stdio.
```
jordigilh, line 142: s/stdio/stdout

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,270 @@
...
+          or Servlet Containers), implemented as an agent on the host.
+          
+  - `generate`: This command takes a canonical representation and a Helm
+    template to produce deployment-ready Helm charts. Details:
+    - **Flags and Options:**
+      - `--input=<file>` (required): Specifies the canonical configuration file
+        to use as input.
+      - `--template=<path>` (required): Path to the Helm template to use for
+        chart generation.
+      - `--output-dir=<directory>` (optional): Directory to save the generated
```
jordigilh, line 141: I think we're no longer saving a helm chart, but generating (rendering) the files as a result of executing the helm template engine against the template files and the discovery manifest, right?

jordigilh:
@JonahSussman can you help to replace all references to `canonical form` to `discovery manifest` to align with the asset generation for PCF enhancement? Unless there is a reason to keep it as canonical form in this enhancement, of course 😄

eemcmullan:
> @JonahSussman can you help to replace all references to `canonical form` to `discovery manifest` to align with the asset generation for PCF enhancement? Unless there is a reason to keep it as canonical form in this enhancement, of course 😄

@JonahSussman @savitharaghunathan +1 to this comment, and then LGTM. Thanks!

enhancements/assets-generation/cli/README.md
```diff
@@ -0,0 +1,270 @@
...
+        - Via live connections using APIs or other supported interfaces.
+        - Through filesystem access to the platform's installation paths (e.g., for Application Servers
+          or Servlet Containers), implemented as an agent on the host.
+          
+  - `generate`: This command takes a canonical representation and a Helm
+    template to produce deployment-ready Helm charts. Details:
+    - **Flags and Options:**
+      - `--input=<file>` (required): Specifies the canonical configuration file
+        to use as input.
+      - `--template=<path>` (required): Path to the Helm template to use for
```
jordigilh, line 139: What do you think of renaming this field to `chart-dir`? Since the command will be generating manifest from both the `templates` and `files` directories inside the chart?

savitharaghunathan: updated

enhancements/assets-generation/cli/README.md
savitharaghunathan: updated. ptal

enhancements/assets-generation/cli/README.md
savitharaghunathan: updated to use discovery-providers

eemcmullan (approved):

PR #215, opened 2025-01-23
Asset generation for Cloud Foundry Applications

jordigilh:
@savitharaghunathan @rromannissen @jortel @dymurray @eemcmullan @JonahSussman please review.

@pkliczewski FYI

### DIFF
--- a/enhancements/assets-generation/cloud-foundry-application-discovery/README.md
+++ b/enhancements/assets-generation/cloud-foundry-application-discovery/README.md
@@ -283,8 +283,8 @@ See [metadata specification](#metadata-specification). |
 | **services** | Services | See [service specification](#service-specification). |
 | **sidecars** | Sidecars | See [sidecar specification](#sidecar-specification). |
 | **metadata** | Metadata | See [metadata specification](#metadata-specification). |
-| **timeout** | StartupTimeout | Maximum time allowed for an application to respond to readiness or health checks during startup.If the application does not respond within this time, the platform will mark the deployment as failed. |
-| **instances** | Replicas | Number of CF application instances |
+| **timeout** | Timeout | Maximum time allowed for an application to respond to readiness or health checks during startup.If the application does not respond within this time, the platform will mark the deployment as failed. |
+| **instances** | Instances | Number of CF application instances |
 | **stack** | Stack | Stack is derived from the `stack` field in the application manifest. The value is captured for information purposes because it has no relevance in Kubernetes. |
 
 ```go
@@ -315,6 +315,8 @@ type Application struct {
   BuildPacks []string `json:"buildPacks,omitempty"`
   // Docker captures the Docker specification in the CF application manifest.
   Docker *Docker `json:"docker,omitempty"`
+  // Instances captures the number of instances to run concurrently for this application
+  Instances *uint `json:"instances,omitempty"`
 }
 ```
 
--- a/enhancements/assets-generation/cloud-foundry-application-discovery/README.md
+++ b/enhancements/assets-generation/cloud-foundry-application-discovery/README.md
@@ -350,7 +350,7 @@ type Sidecar struct {
   // accumulate more than one type.
   ProcessTypes []ProcessType `json:"processType" validate:"required,oneof=worker web"`
   // Command captures the command to run the sidecar
-  Command []string `json:"command" validate="required"`
+  Command []string `json:"command" validate:"required"`
   // Memory represents the amount of memory to allocate to the sidecar.
   // It's an optional field.
   Memory string `json:"memory,omitempty"`
--- a/enhancements/assets-generation/cloud-foundry-application-discovery/README.md
+++ b/enhancements/assets-generation/cloud-foundry-application-discovery/README.md
@@ -227,7 +227,7 @@ Example:
 | Name | Type | Description |
 | ----- | ----- | ----- |
 | **route** | string | **(Required)** The route URI |
-| **protocol** | string | (Optional) Protocol to use for this route. Valid protocols are `http1`, `http2`, and `tcp`. |
+| **protocol** | string | (Optional) Protocol to use for this route. Valid protocols are `http`, `http2`, and `tcp`. |
 
 #### [Service-level configuration](https://docs.cloudfoundry.org/devguide/deploy-apps/manifest-attributes.html#services-block)
 
@@ -290,7 +290,7 @@ See [metadata specification](#metadata-specification). |
 ```go
 type Application struct {
   // Metadata captures the name, labels and annotations in the application.
-  Metadata Metadata `json:",inline"`
+  Metadata Metadata `json:",inline" validate:"required"`
   // Env captures the `env` field values in the CF application manifest.
   Env map[string]string `json:"env,omitempty"`
   // Routes represent the routes that are made available by the application.
@@ -309,8 +309,8 @@ type Application struct {
   // respond to readiness or health checks during startup.
   // If the application does not respond within this time, the platform will mark
   // the deployment as failed. The default value is 60 seconds.
-  // https://github.com/cloudfoundry/docs-dev-guide/blob/96f19d9d67f52ac7418c147d5ddaa79c957eec34/deploy-apps/  large-app-deploy.html.md.erb#L35
-  StartupTimeout uint `json:"startupTimeout,omitempty"`
+  // https://github.com/cloudfoundry/docs-dev-guide/blob/96f19d9d67f52ac7418c147d5ddaa79c957eec34/deploy-apps/large-app-deploy.html.md.erb#L35
+  StartupTimeout *uint `json:"startupTimeout,omitempty"`
   // BuildPacks capture the buildpacks defined in the CF application manifest.
   BuildPacks []string `json:"buildPacks,omitempty"`
 }
@@ -328,13 +328,13 @@ type Application struct {
 ```go
 type Sidecar struct {
   // Name represents the name of the Sidecar
-  Name string `json:"name"`
+  Name string `json:"name" validate:"required"`
   // ProcessTypes captures the different process types defined for the sidecar.
   // Compared to a Process, which has only one type, sidecar processes can
   // accumulate more than one type.
-  ProcessTypes []ProcessType `json:"processType"`
+  ProcessTypes []ProcessType `json:"processType" validate:"required,oneof=worker web"`
   // Command captures the command to run the sidecar
-  Command []string `json:"command"`
+  Command []string `json:"command" validate="required"`
   // Memory represents the amount of memory to allocate to the sidecar.
   // It's an optional field.
   Memory string `json:"memory,omitempty"`
@@ -358,7 +358,7 @@ type Service struct {
   // application. This field represents the runtime name of the service, captured
   // from the 3 different cases where the service name can be listed.
   // For more information check https://docs.cloudfoundry.org/devguide/deploy-apps/manifest-attributes.html#services-block
-  Name string `json:"name"`
+  Name string `json:"name" validate:"required"`
   // Parameters contain the k/v relationship for the aplication to bind to the service
   Parameters map[string]interface{} `json:"parameters,omitempty"`
   // BindingName captures the name of the service to bind to.
@@ -380,7 +380,7 @@ type Service struct {
 ```go
 type Metadata struct {
   // Name capture the `name` field int CF application manifest
-  Name string `json:"name"`
+  Name string `json:"name" validate:"required"`
   // Space captures the `space` where the CF application is deployed at runtime. The field is empty if the
   // application is discovered directly from the CF manifest. It is equivalent to a Namespace in Kubernetes.
   Space string `json:"space,omitempty"`
@@ -429,11 +429,11 @@ type Process struct {
   // Memory represents the amount of memory requested by the process.
   Memory string `json:"memory,omitempty"`
   // HealthCheck captures the health check information
-  HealthCheck Probe `json:"healthCheck"`
+  HealthCheck *Probe `json:"healthCheck,omitempty"`
   // ReadinessCheck captures the readiness check information.
-  ReadinessCheck Probe `json:"readinessCheck"`
+  ReadinessCheck *Probe `json:"readinessCheck,omitempty"`
   // Replicas represents the number of instances for this process to run.
-  Replicas uint `json:"replicas"`
+  Replicas uint `json:"replicas" validate:"required"`
   // LogRateLimit represents the maximum amount of logs to be captured per second.
   LogRateLimit string `json:"logRateLimit,omitempty"`
 }
@@ -464,17 +464,19 @@ const (
 | **health-check-http-endpoint** | Endpoint | HTTP endpoint to be used for health checks, specifying the path to be monitored. |
 | **health-check-invocation-timeout** | Timeout | Maximum time allowed for each health check invocation to complete. |
 | **health-check-interval** | Interval | Interval at which health checks are performed to monitor the application’s status. |
-| **health-check-type** | N |  | Specifies the type of health check to perform (`none`, `http`, `tcp`, or `process`). Note: `none` is deprecated and an alias for process. |
+| **health-check-type** | Type  | Specifies the type of health check to perform (`none`, `http`, `tcp`, or `process`). Note: `none` is deprecated and an alias for process. |
 
 ```go
 type Probe struct {
   // Endpoint represents the URL location where to perform the probe check.
-  Endpoint string `json:"endpoint"`
+  Endpoint string `json:"endpoint" validate:"required"`
   // Timeout represents the number of seconds in which the probe check can be considered as timedout.
   // https://docs.cloudfoundry.org/devguide/deploy-apps/manifest-attributes.html#timeout
-  Timeout uint `json:"timeout"`
+  Timeout uint `json:"timeout" validate:"required"`
   // Interval represents the number of seconds between probe checks.
-  Interval uint `json:"interval"`
+  Interval uint `json:"interval" validate:"required"`
+  // Type specifies the type of health check to perform
+  Type string `json:"type" validate:"required,oneof=http tcp process"`
 }
 ```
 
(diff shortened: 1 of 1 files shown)


enhancements/assets-generation/cloud-foundry-application-discovery/README.md
```diff
@@ -0,0 +1,668 @@
...
+   accepted as ready to execute. Anything marked as `implementable` should
+   clearly communicate how an enhancement is coded up and delivered. If an
+   enhancement describes a new deployment topology or platform, include a
+   logical description for the deployment, and how it handles the unique aspects
+   of the platform. Aim for single topic PRs to keep discussions focused. If you
+   disagree with what is already in a document, open a new PR with suggested
+   changes.
+
+The `Metadata` section above is intended to support the creation of tooling
+around the enhancement process.
```
savitharaghunathan, line 59: nit: This template data should be removed

enhancements/assets-generation/cloud-foundry-application-discovery/README.md
```diff
@@ -0,0 +1,668 @@
...
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] User-facing documentation is created
+
+## Open Questions [optional]
+
+This is where to call out areas of the design that require closure before deciding
+to implement the design.  For instance,
+ > 1. This requires exposing previously private resources which contain sensitive
+  information.  Can we do this?
```
savitharaghunathan, line 73: nit: same as above

enhancements/assets-generation/cloud-foundry-application-discovery/README.md
```diff
@@ -0,0 +1,668 @@
...
+of Move2Kube is to migrate Cloud Foundry applications to Kubernetes, but users
+have been struggling with the template language used to capture the Kubernetes
+resources, as it is not well known and requires an additional effort to master,
+ compared to other templating languages.
+
+## Motivation
+
+The challenge brought by the templating language severely impacts the usability
+and acceptance of the tool. Thus the existence of this enhancement to provide a
+similar tool to MTA, extensible, and that improves on the templating engine, so
```
savitharaghunathan, line 95: nit: MTA -> Konveyor

enhancements/assets-generation/cloud-foundry-application-discovery/README.md
jordigilh: Done

enhancements/assets-generation/cloud-foundry-application-discovery/README.md
```diff
@@ -0,0 +1,641 @@
...
+### Proposal Specification
+
+#### Space specification
+
+| Name | Mapped (Y/N) | Canonical Form | Description |
+| ----- | :---: | ----- | ----- |
+| **applications** | Y | Application | Direct mapping to a slice of canonical form manifests, each one representing the discovery results of a CF application. See [app-level specification](#application-specification) |
+| **space** | Y | Metadata.Space | See [metadata specification](#metadata-specification). This field is only populated at runtime. |
+| **version** | Y | Metadata.Version | The manifest schema version; currently the only valid version is 1, defaults to 1 if not provided. This field is only populated at runtime. |
+
```
rromannissen, line 269: As agreed, we should be mapping all fields we discover in CF, so the Mapped column in these specification tables should be removed

enhancements/assets-generation/cloud-foundry-application-discovery/README.md
```diff
@@ -0,0 +1,641 @@
...
+  // StartupTimeout specifies the maximum time allowed for an application to
+  // respond to readiness or health checks during startup.
+  // If the application does not respond within this time, the platform will mark
+  // the deployment as failed. The default value is 60 seconds.
+  // https://github.com/cloudfoundry/docs-dev-guide/blob/96f19d9d67f52ac7418c147d5ddaa79c957eec34/deploy-apps/  large-app-deploy.html.md.erb#L35
+  StartupTimeout uint `json:"startupTimeout,omitempty"`
+  // BuildPacks capture the buildpacks defined in the CF application manifest.
+  BuildPacks []string `json:"buildPacks,omitempty"`
+}
+```
```
rromannissen, line 317: I'm not sure about the cardinality here. I'm not a Go guy myself, but for example I'm not reading that an application can be associated to multiple services in here as it is right now. Again, might be my ignorance on Go, but I would like to have a clear view on the cardinality between the different structs

rromannissen (changes requested):
Really looks good to me, very thorough documentation of the proposed canonical configuration manifest spec. Only some minor changes to do and we should be good to go IMHO.

jordigilh (commented):
Updated with suggestions

enhancements/assets-generation/cloud-foundry-application-discovery/README.md
```diff
@@ -0,0 +1,659 @@
...
+| Name | Type | Description |
+| ----- | ----- | ----- |
+| **name** | string | **(Required)** The identifier for the sidecars to be configured |
+| **command** | string | The command used to start the sidecar |
+| **process\_types** | array of strings | List of processes to associate sidecar with |
+| **memory** | integer | Memory in MB that the sidecar will be allocated |
+
+### Proposal Specification
+
+#### Space specification
```
jortel, line 262: _Space_ seems CF oriented and wondering if _Namespace_ would be more generic and better aligned with konveyor vernacular?

jordigilh: We had a similar thought on several situations. We kept the source name for:
* `Process` -> `Container`
* `Sidecar` -> `Container`
But we decided to change for:
* `Instances -> `Replicas`. At the time we thought to align with Kubernetes domain name, but it was a long debate and still ongoing, hopefully this comment can help us decide 😄 
* `Timeout` -> `StartupTimeout` (This one we renamed because `timeout` didn't convey it's intent, compared to the other `timeout` field instances found in the application health and readiness checks)

So, in a nutshell, what is the normal expectation in the discovery? use a nomenclature that is aligned with the source or with konveyor so that all discovery outputs share a common naming convention?

rromannissen: I'd say we should use a nomenclature aligned with the source platform terminology. For me, it doesn't make sense to try to change the source names, since that would introduce some bias from our side on how we see the conversion happening, when that should be totally up to the templates author.

jordigilh:
@rromannissen is it ok if we don't interpret the default values in the CF manifests when not provided in the source manifest? For instance, the number of `instances` defaults to 1 when no value is provided in the CF manifest.

dymurray:
I'd suggest that we rename this enhancement to specifically cover asset generation and the definition of the canonical configuration. The example of using cloud foundry as the source should be a specific example called out in the enhancement, but not the title of the enhancement

rromannissen:
> @rromannissen is it ok if we don't interpret the default values in the CF manifests when not provided in the source manifest? For instance, the number of `instances` defaults to 1 when no value is provided in the CF manifest.

I think we need to take that into account, because default values might be different in the target platform. In other words, if default for instances is 1, we need to explicitly populate the canonical manifest with that number, even if it's missing in the source deployment manifest.

enhancements/assets-generation/cloud-foundry-application-discovery/README.md
jortel: It's my understanding that the proposal is for the _canonical_ form which is intended to be source agnostic.  An abstraction.  So "use a nomenclature aligned with the source platform terminology. " does not make sense to me.  If anything, it seems better to be biased toward kubernetes nomenclature which is well-known to the konveyor community.
Am I missing something?

enhancements/assets-generation/cloud-foundry-application-discovery/README.md
```diff
@@ -0,0 +1,661 @@
...
+| **Memory** | Memory | (Optional) The amount of memory to allocate to the sidecar. |
+
+```go
+type Sidecar struct {
+  // Name represents the name of the Sidecar
+  Name string `yaml:"name" validate:"required"`
+  // ProcessTypes captures the different process types defined for the sidecar.
+  // Compared to a Process, which has only one type, sidecar processes can
+  // accumulate more than one type.
+  ProcessTypes []ProcessType `yaml:"processType" validate:"required,oneof=worker web"`
```
jortel, line 355: Enum: (worker|web) feel CF specific.

eemcmullan (commented):
The enhancement looks good and detailed so far. Thank you for putting it together. I am wondering about @dymurray and @jortel's comments on this enhancement: aiming to be more standardized towards any platform, while using CF as an example scenario. In which case, some of these configs could be implemented as general platform types, with optional fields for platform-specific manifest fields. I'm interested to know of any thoughts on this.

enhancements/assets-generation/cloud-foundry-application-discovery/README.md
jordigilh: I'm not sure I'm following you: Is your comment related to the fact that we're following the names for CF process types?

jordigilh:
> The enhancement looks good and detailed so far. Thank you for putting it together. I am wondering about @dymurray and @jortel's comments on this enhancement: aiming to be more standardized towards any platform, while using CF as an example scenario. In which case, some of these configs could be implemented as general platform types, with optional fields for platform-specific manifest fields. I'm interested to know of any thoughts on this.

My impression was that we would have a canonical form for each source discovery platform. The benefit of this is that each source platform is contained within its own canonical form, and since the discovery captures the source fields verbatim, it makes it an easy task to map each field back to the source. For instance, mandatory fields in CF might not apply to other platforms, which makes the validation of the canonical form difficult because all fields will have to be optional by default in the schema, and the validation is then implemented in the code.
 
New source platforms can be discovered to their own canonical form. The process of interpreting the data then falls into the transformation. With Helm, this is done in the templates. I also find it easier to maintain a specific canonical form per platform in terms of code, testing and documentation.

The downside from using a generic canonical form for all platforms is that it becomes an effort in the discovery to map the values, because each source field needs to be now interpreted and mapped to the canonical form, if possible. And if it's not clearly possible, it can spawn a new field in the generic canonical form to encapsulate this meaning. Additionally, the canonical form is no longer clearly defined, since it is extendable and each new platform added can extend the canonical form by adding new fields that are not relevant to other platforms.

rromannissen:
I think @jordigilh nailed it with its explanation. If we were to find a generic canonical form to store data, we would be implicitly forcing a certain transformation path, and that should be entirely up to the templates author.

dymurray:
@jordigilh I much better understand the content of this enhancement now. So IIUC, the high level summary is that there will exist a canonical configuration format which has a subsection `cloudFoundry` (for example). That subsection dictionary is what is defined in this enhancement.

I was slightly confused at first thinking this was the definition of the root level canonical configuration being heavily guided by CF data. I now understand that you intend for this to be a platform specific section in the config. We will define the canonical configuration spec in a separate enhancement.

enhancements/assets-generation/cloud-foundry-application-discovery/README.md
```diff
@@ -0,0 +1,661 @@
...
+creation-date: 2025-01-17
+last-updated: 2025-01-17
+status: provisional
+see-also:
+  - https://github.com/konveyor/enhancements/pull/210
+replaces:
+superseded-by:
+---
+
+# Cloud Foundry Application Discovery to canonical form
```
dymurray, line 29: I know this is pedantic; but the more I understand about this enhancement I don't like the term "canonical configuration". In general; the term I like for this is "manifest". We are talking about converting a CF application runtime configuration into a yaml file that can be used for app deployment on a new platform. "Canonical configuration" doesn't really define this

enhancements/assets-generation/cloud-foundry-application-discovery/README.md
```diff
@@ -0,0 +1,661 @@
...
+
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] User-facing documentation is created
+
+## Summary
```
dymurray, line 39: Please add a blurb here about how the goal of the enhancement is to define the manifest output format for the CF discovery process to be consumed by the generation command for the assets

enhancements/assets-generation/cloud-foundry-application-discovery/README.md
```diff
@@ -0,0 +1,661 @@
...
+required for an application.
+
+Move2Kube is an IBM research project with the goal to provide tools to migrate
+applications to other platforms, particularly Kubernetes. One of the use cases
+of Move2Kube is to migrate Cloud Foundry applications to Kubernetes, but users
+have been struggling with the template language used to capture the Kubernetes
+resources, as it is not well known and requires an additional effort to master,
+ compared to other templating languages.
+
+## Motivation
```
dymurray, line 55: I'd like to add something here about how Konveyor as a whole supports analysis of an application with the goal of finding issues with an application's compatibility with a target platform. Asset generation as a feature enables Konveyor to give users the ability to auto generate deployment manifests for that target platform

enhancements/assets-generation/cloud-foundry-application-discovery/README.md
```diff
@@ -0,0 +1,661 @@
...
+
+The challenge brought by the templating language severely impacts the usability
+and acceptance of the tool. Thus the existence of this enhancement to provide a
+similar tool to Konveyor, extensible, and that improves on the templating engine, so
+that it offers a pluggable design that can be used to implement well known
+templating engines.
+
+### Goals
```
dymurray, line 63: Let's add a goal about defining how the manifest gets used by our generator to render templates?

Unsure if that's the right wording; but we need the story captured about how the YAML file is used to render templates

dymurray: Well... thinking about it I suppose that might be out of scope; maybe more specifically defining and documenting the YAML format of the manifest

enhancements/assets-generation/cloud-foundry-application-discovery/README.md
dymurray: After more discussion it sounds like "canonical configuration" has confused most of us. We now agree this is the specific platform configuration for PCF; and the naming should change. It is *not* source agnostic; and will be well documented for PCF

enhancements/assets-generation/cloud-foundry-application-discovery/README.md
```diff
@@ -0,0 +1,661 @@
...
+*Mitigation*
+
+Users will have to do due diligence before deploying on K8S, including the
+creation of resources prior to the deployment of the K8S assets, such as
+`namespaces`, `services`, `secrets` and container images for `buildpacks`,
+for instance. \
+Developers may need to refactor applications or implement new solutions that
+align with Kubernetes practices.
+
+## Design Details
```
dymurray, line 613: Please include an example of the YAML that gets generated from this struct

dymurray (changes requested):
Just some mild changes requested; but overall this looks really good. I followed the example PoC and it made a lot sense. After these comments are addressed I will feel good about it

enhancements/assets-generation/cloud-foundry-application-discovery/README.md
jordigilh: Done

enhancements/assets-generation/cloud-foundry-application-discovery/README.md
jordigilh: Done

enhancements/assets-generation/cloud-foundry-application-discovery/README.md
jordigilh: Let me know what name the team decides to use and we'll make the changes in the document.

rromannissen: Discovery manifest works for me!

enhancements/assets-generation/cloud-foundry-application-discovery/README.md
rromannissen: Discovery manifest should do.

rromannissen (commented):
Looks great to me after the changes, I think this is ready to merge!

enhancements/assets-generation/cloud-foundry-application-discovery/README.md
jordigilh: @dymurray please review and close the comment if you're ok with the changes I made.

enhancements/assets-generation/cloud-foundry-application-discovery/README.md
jordigilh: @dymurray please review and close the comment if you're ok with the changes I made.

enhancements/assets-generation/cloud-foundry-application-discovery/README.md
jordigilh: Do you still want to add any reference here? perhaps that the goal to the canonical form manifest is used to render the templates is out of scope in this document?

enhancements/assets-generation/cloud-foundry-application-discovery/README.md
jordigilh: if @jortel and @dymurray are fine, I'll update the document to refer to the canonical form as discovery manifest.

enhancements/assets-generation/cloud-foundry-application-discovery/README.md
jordigilh: @dymurray please review and close the comment if you're ok with the changes I made.

dymurray: Great thanks!

enhancements/assets-generation/cloud-foundry-application-discovery/README.md
dymurray: Nah after my first review I think this is fine. Sorry for the confusion, resolving

enhancements/assets-generation/cloud-foundry-application-discovery/README.md
dymurray: I like discovery manifest, please!

enhancements/assets-generation/cloud-foundry-application-discovery/README.md
dymurray: Good by me!

jordigilh: Replaced to `discovery manifest`

jordigilh:
@dymurray @rromannissen `canonical form` renamed to `discovery manifest`

eemcmullan (approved):

Issue #232, opened 2025-08-20
[RFE] Use local mechanism to get dependency information


jmle:
Instead of relying on the Maven Search service online, we need to find a way of providing this information without the need to connect to the internet. We need to first make some research about the possibilities and then have a chat about the best one to proceed.

Some ideas are:
- Packaging the info directly with the CLI/UI
- Giving the user the option to download it if needed
- Putting up an alternative online service (won't work without internet connection)

shawn-hurley:
I do wonder if we separate out the solutions for the different things. 

### Hub
1. Having a service that is shipped with the hub, that contains some subset of the index, and is accessible to the addon/task pod probably makes sense. 
2. What I am hoping is that this can be its own image.

### Kantra
1. Having the ability to expose the same service over RPC may be necessary. (Option 1)
2. Making it so that the java-provider can read this information and use it locally, probably by reading the file in the `.kantra` directory

For both options, I think it would be really interesting to add the ability to optionally query the Maven proxy that a company uses to get this information. I think to start, supporting [Nexus](https://help.sonatype.com/en/searching-for-components.html) and [Artifactory](https://jfrog.com/help/r/jfrog-rest-apis/checksum-search). We can then add an option to use the service provided by the hub in `1` above for Kantra as well, or you could run the service locally? 

I'm not sure if that makes sense, but I wanted to add to the issue as an idea to consider during the enhancement.

PR #234, opened 2025-09-18
:sparkles: Add enhancement for local Maven index search

jmle:
Addresses https://github.com/konveyor/enhancements/issues/232

## Summary by CodeRabbit

* **Documentation**
  * Added a design document for a Local Maven Search using a pre-indexed artifact list for fast, low-memory lookups.
  * Describes data layout (ordered artifact list + offset index), access patterns, and performance characteristics.
  * Covers data generation, bundling options, upgrade/downgrade strategy, test plan, risks, and trade-offs.
  * Compares alternatives and records implementation considerations for future work.

### DIFF
--- a/enhancements/local-maven-search/README.md
+++ b/enhancements/local-maven-search/README.md
@@ -81,7 +81,7 @@ and saving hundreds of MB of RAM. Search is done via binary search, which runs i
 #### Story 1
 As a user,
 I want to be able to get a full list of dependencies
-When I analyze a Java binary with embedded libraries
+When I analyze a Java binary with embedded libraries without internet connection
 
 ### Implementation Details/Notes/Constraints
 
@@ -97,26 +97,25 @@ Ideally, we would want to modernize this project, since it's very old and gobble
 
 The process of regenerating the index could either be included in the process of generating the Java provider image,
 or could be completely independent. The Java provider image build would then simply download the pre-generated data
-from a Github releases link.
+from a GitHub releases link.
 
-#### 2. Data bundling
-Data could be pre-generated and potentially **compressed** for a smaller disk footprint:
-```
--rw-r--r--. 1 jmle jmle 877M Sep 16 10:57 central.archive-metadata.sorted.txt
--rw-r--r--. 1 jmle jmle 504M Sep 16 11:49 index.bin
-```
+- Given the urgency of this enhancement, a first phase will get the previously-generated data from a [pre-made release in GitHub](https://github.com/konveyor/maven-search-index/releases/tag/v0.0.1).
 
-would become
-```
--rw-r--r--. 1 jmle jmle 388M Sep 18 16:31 data.tar.gz
--rw-r--r--. 1 jmle jmle 241M Sep 18 16:30 index.tar.gz
-```
+#### 2. Data contents
+The data would consist of two files:
+1. **The artifacts list**: an ordered *.txt file with a map-like structure, that will contain entries in the form `<sha> <groupId>:<artifactId>:<packaging>::<version>`. For instance: `00000002832188c16354e5dbb141f3b8e7462976 org.geneweaver:gweaver-bulk-import:jar::1.0.11`
+2. **The index for the artifacts list**: a binary file containing, for each sha, the offset in the artifacts list (1) where the information for that particular SHA is located. This file will be searched on using binary search and give the location of the entry in the artifacts file. Its contents will be in the form `<sha> <offset>`.
 
-This data could be directly included in the Java provider and the CLI. If compressed, the data would have to be decompressed
-before the first analysis
+#### 3. Data bundling
+The easiest and most straightforward way of including this data is to bundle it within the images themselves and kantra.
+The disk footprint of the artifacts will increase, but this will allow us to avoid making the client download the data at any time, which, in any case, wouldn't be possible in disconnected environments.
 
-#### 3. Data access
-Accessing the data would be done through a few simple functions to query the files.
+Having the data compressed would enable us to decrease the size of the images in a few hundred MBs, but we would incur in lessened performance
+due to having to decompress the data before each run (for instance, when running the Java provider in the hub, decompression would be needed before
+analysis each time a task is created). Disk size shouldn't be a problem nowadays, so the data will be included as-is.
+
+#### 4. Data access
+Accessing the data would be done through a few simple functions to query the files. The flow diagram is explained in the Proposal section.
 
 
 ### Security, Risks, and Mitigations
@@ -131,12 +130,10 @@ Testing would be done by comparing a successful analysis of an older version of 
 with an analysis done by the latest version with local index data.
 
 ### Upgrade / Downgrade Strategy
-
 Upgrading of the data would be done at first by using the Windup project mentioned earlier. Ideally, we would upgrade
 and modify it so that it is more modern and suits our needs.
 
 ## Implementation History
-
 TBD
 
 ## Drawbacks
@@ -151,5 +148,4 @@ A possible alternative would be to load all the information into memory, or usin
 approaches also come with their own drawbacks.
 
 ## Infrastructure Needed
-
-No special infrastructure will be needed.
+TBD - we might need additional infrastructure to run the data generation process


enhancements/local-maven-search/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+There are several sides to this proposal:
+
+#### 1. Data generation
+The number of artifacts available in Maven Central is constantly changing (increasing), therefore we need to be able to
+regenerate the list of artifacts at will. There exists an [old Windup project](https://github.com/windup/nexus-repository-indexer)
+which does exactly this, and then creates a new, simplified index for consumption in Windup. The part we are interested in is the
+*.txt file(s) created as a byproduct of the Windup index generation (we are not interested in the new Lucene index created after this).
+
+Ideally, we would want to modernize this project, since it's very old and gobbles up a lot of RAM during execution.
```
eemcmullan, line 96: Is this something to be completed before analyzer implementation?

enhancements/local-maven-search/README.md
```diff
@@ -0,0 +1,155 @@
...
+and modify it so that it is more modern and suits our needs.
+
+## Implementation History
+
+TBD
+
+## Drawbacks
+
+The main drawback of this approach is the increase in the size of the image (in the case of the Konveyor UI) and the
+CLI release (in the case of Kantra). My argument is that ideally we want to avoid using even more RAM memory, as
```
eemcmullan, line 145: Do you know about how much memory this would consume?

enhancements/local-maven-search/README.md
```diff
@@ -0,0 +1,155 @@
...
+to know whether a JAR found within a Java binary is a publicly available library or not.
+
+This is not convenient and causes issues in several scenarios:
+- When the user does not have access to the internet.
+- When the MCI search service is down.
+
+Even though we have several fallback routines, like relying on a local list of open-source Group IDs or
+directly parsing the POM, they are not perfect and their results leave room for improvement.
+
+Here we will present an alternative, consisting on locally bundling the information and querying it.
```
eemcmullan, line 48: Just to clarify, we prioritize the local search, but can still use the maven search -> fallbacks if the necessary index file is not found?

enhancements/local-maven-search/README.md
```diff
@@ -0,0 +1,155 @@
...
+- To be able to query the information quickly from the Java provider, be it in the UI or the CLI.
+- To be able to consume as few resources as possible.
+
+### Non-Goals
+
+N/A
+
+## Proposal
+
+The main part of the proposal consists of a new binary with the ability to query a file with the information (Ordered list of artifacts). This file will have been previously indexed, creating an additional file (offsets file) which contains the offsets in disk of each artifact's hash in the ordered list of artifacts.
```
eemcmullan, line 63: When is the offset file created?

enhancements/local-maven-search/README.md
```diff
@@ -0,0 +1,155 @@
...
+-rw-r--r--. 1 jmle jmle 504M Sep 16 11:49 index.bin
+```
+
+would become
+```
+-rw-r--r--. 1 jmle jmle 388M Sep 18 16:31 data.tar.gz
+-rw-r--r--. 1 jmle jmle 241M Sep 18 16:30 index.tar.gz
+```
+
+This data could be directly included in the Java provider and the CLI. If compressed, the data would have to be decompressed
```
eemcmullan, line 115: For the CLI & UI, apart from updating the Java provider and release artifacts, will there be further work? I can't imagine much more in the CLI, but am unsure of additional UI steps.

enhancements/local-maven-search/README.md
jmle: Having the local search, I'm not sure we would need the HTTP calls or the group ID list anymore.

aufi (commented):
Looks great, just left few comments to clarify things.

enhancements/local-maven-search/README.md
aufi: Discussed directly with Juanma, sample line format of the txt file was provided:

`<sha> <groupId>.<artifactId>.<version>.<packaging>`

This looks good to me, just the way of searching this file (with or without text-based index file) is something that could be resolved during implementation process when we could see what actually better.

aufi (approved):

enhancements/local-maven-search/README.md
```diff
@@ -0,0 +1,155 @@
...
+TBD
+
+## Summary
+
+This enhancements presents a possible solution for locally bundling and querying
+the Maven Index list of publicly available artifacts instead of relying on external services.
+
+## Motivation
+
+Up until now, Konveyor has relied mainly on HTTP queries to the [MCI search service](https://central.sonatype.com/search)
```
shawn-hurley, line 38: I think that it might be better to use the [rest-api](https://central.sonatype.org/search/rest-api-guide/#rest-api) link, which tells you to still use the deprecated `search.maven.org` for the rest api.

enhancements/local-maven-search/README.md
```diff
@@ -0,0 +1,155 @@
...
+to know whether a JAR found within a Java binary is a publicly available library or not.
+
+This is not convenient and causes issues in several scenarios:
+- When the user does not have access to the internet.
+- When the MCI search service is down.
```
shawn-hurley, line 43: Worth adding a line about how the rest api is deprecated, and has been increasingly unreliable.

enhancements/local-maven-search/README.md
shawn-hurley: I would like to add this and the offset file's structure's to the enhancement.

enhancements/local-maven-search/README.md
```diff
@@ -0,0 +1,155 @@
...
+Ideally, we would want to modernize this project, since it's very old and gobbles up a lot of RAM during execution.
+
+The process of regenerating the index could either be included in the process of generating the Java provider image,
+or could be completely independent. The Java provider image build would then simply download the pre-generated data
+from a Github releases link.
+
+#### 2. Data bundling
+Data could be pre-generated and potentially **compressed** for a smaller disk footprint:
+```
+-rw-r--r--. 1 jmle jmle 877M Sep 16 10:57 central.archive-metadata.sorted.txt
```
shawn-hurley, line 105: What is included in this file and what is it's structure

enhancements/local-maven-search/README.md
```diff
@@ -0,0 +1,155 @@
...
+-rw-r--r--. 1 jmle jmle 504M Sep 16 11:49 index.bin
```
shawn-hurley, line 106: What is in this file and what is it's structure

enhancements/local-maven-search/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+### Security, Risks, and Mitigations
+
+No security risks are identified.
+
+## Design Details
+
+### Test Plan
+
+Testing would be done by comparing a successful analysis of an older version of Konveyor with enabled Maven Search
```
shawn-hurley, line 130: Instead of comparing to older versions, I think we should have a test plan that verifies that the actual data is correct.

For this, what I am mean Is create a jar, that has known specific versions of dependencies in it, and verify that the dependencies are found, with the correct versions, and labeled correctly.  I am wondering if this could be a  go-konveyor-test? 

For a local test in analyzer-lsp could we make sure that the demo dep output file for the a java file, will be able to verify this is working?

shawn-hurley (changes requested):
I think that we need to gain consensus on:

1. What image contains this new set of data.
2. When is it generated.
3. Is the generated data stored in github plaintext as the maven.default.index is? (This is because to bring it downstream for offline builds we can't generate the file in that system).

It would be good to compare what is being done already to generate the maven.default.index, and what is being proposed here and see if there is overlap. There probably is, and we probably just want to have a single process generate the data if there are both working off the same thing.

enhancements/local-maven-search/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+### Non-Goals
+
+N/A
+
+## Proposal
+
+The main part of the proposal consists of a new binary with the ability to query a file with the information (Ordered list of artifacts). This file will have been previously indexed, creating an additional file (offsets file) which contains the offsets in disk of each artifact's hash in the ordered list of artifacts.
+
+```
```
pranavgaikwad, line 65: How is the offset calculated? Or is this an internal function of the Lucene indexer we will re-use and out-of-scope for us?

pranavgaikwad (changes requested):
Are you proposing to discontinue the existing maven index text file with this?

enhancements/local-maven-search/README.md
```diff
@@ -0,0 +1,151 @@
...
+The process of regenerating the index could either be included in the process of generating the Java provider image,
+or could be completely independent. The Java provider image build would then simply download the pre-generated data
+from a GitHub releases link.
+
+- Given the urgency of this enhancement, a first phase will get the previously-generated data from a [pre-made release in GitHub](https://github.com/konveyor/maven-search-index/releases/tag/v0.0.1).
+
+#### 2. Data contents
+The data would consist of two files:
+1. **The artifacts list**: an ordered *.txt file with a map-like structure, that will contain entries in the form `<sha> <groupId>:<artifactId>:<packaging>::<version>`. For instance: `00000002832188c16354e5dbb141f3b8e7462976 org.geneweaver:gweaver-bulk-import:jar::1.0.11`
+2. **The index for the artifacts list**: a binary file containing, for each sha, the offset in the artifacts list (1) where the information for that particular SHA is located. This file will be searched on using binary search and give the location of the entry in the artifacts file. Its contents will be in the form `<sha> <offset>`.
```
shawn-hurley: How much memory would be used to bring this in as a map in go?

PR #247, opened 2025-11-07
Kantra Refactor - Subcommands & Analysis Config

eemcmullan:
## Summary by CodeRabbit

* **New Features**
  * Added new CLI subcommands: analyze, rules, config, provider, transform, discover, generate
  * Enabled config-driven analysis via an Analysis Config File with clear precedence where command-line flags override config
  * Backward-compatibility mapping for legacy options with deprecation notices

* **Documentation**
  * Added comprehensive CLI refactor guide: usage examples, config schema, test plan, goals, signoff checklist, and known drawbacks

### DIFF
--- a/enhancements/kantra/subcommands/README.md
+++ b/enhancements/kantra/subcommands/README.md
@@ -1,0 +1,213 @@
+---
+title: Kantra Subcommands and Analysis Config
+authors:
+  - "@eemcmullan"
+reviewers:
+  - TBD
+approvers:
+  - TBD
+creation-date: 2025-11-03
+last-updated: 2025-11-06
+status: implementable
+---
+
+# Kantra CLI Refactor
+
+This enhancement introduces a refactor of the Kantra CLI to include a new set of 
+subcommands and a configuration file for analysis options.
+
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] User-facing documentation is created
+
+
+## Summary
+
+In the first iteration of the Kantra CLI, the goal was feature parity with 
+Windup https://github.com/windup/windup with two major subcommands: `analyze` and 
+`transform` for Java applications. 
+
+Currently, the scope of Kantra has expanded to analyze multiple language applications,
+gather insights and modify applicaitons running on platforms, and communicate 
+with the Konveyor Hub for synchronized appliction analysis configurations.
+
+We can improve the user experience of the Kantra CLI by updating the subcommands 
+that make sense today. We can also set all analysis options from a config file, 
+similar to the [centralized config](https://github.com/konveyor/enhancements/pull/241) 
+anaysis profile. This file will include analyzer `providerConfig` options, as 
+well as CLI analysis options.
+
+
+### Goals
+
+- Refactor the CLI subcommands and flags so that each subcommand is
+  self-contained, improving discoverable actions for users.
+- Include the ability to configure analysis options from a file.
+- Maintain backwards compatibility.
+
+### Non-Goals
+
+- Discuss moving non-anlysis related subcommands.
+
+
+## Proposal
+
+The Kantra CLI will include the following subcommands:
+
+- `kantra analyze`
+
+- `kantra rules`
+
+- `kantra config`
+
+- `kantra provider`
+
+- `kantra transform`
+
+- `kantra discover`
+
+- `kantra generate`
+
+The Kantra CLI will also have the ability to read provider and analysis options
+from a config file.
+
+
+## Implementation Details
+
+### Subcommands:
+
+#### analyze
+
+`kantra analyze` will only be responsible for options involving running analysis.
+Such as:
+
+- `kantra analyze --config-file <path/to/config> --output <output-dir>`
+
+*There will also be an undetermined default path to this file.*
+
+- `kantra analyze --input <path/to/app> --output <output-dir> --target <target>`
+
+#### rules
+
+`kantra rules` will provide options for interacting with rules and rulesets for 
+analysis configuration. This includes both default rulesets and custom rulesets. 
+
+The current `test` subcommand will fall under this new subcommand.
+
+- `kantra rules --list-targets`
+
+- `kantra rules test <path-to-test-rules>`
+
+#### config
+
+`kantra config` will be responsible for options to configure analysis config 
+files and profiles. This will include:
+
+- `kantra config login`
+
+- `kantra config --sync <app>`
+
+- `kantra config --list-profiles`
+
+#### provider
+
+`kantra provider` will handle functionality around using providers, including 
+installing and removing providers. This will be included once Kantra supports 
+[socket communication](https://github.com/konveyor/enhancements/pull/231).
+
+- `kantra provider --list-providers`
+
+- `kantra provider --install <provider>`
+
+- `kantra provider --remove <provider`
+
+#### transform
+
+`kantra transform` will remain as is.
+
+#### discover
+
+`kantra discover` will remain as is.
+
+#### generate
+
+`kantra generate` will remain as is.
+
+
+### Analysis Config File
+
+Below is an example config file that can be used to read and set all
+analysis options from the provider, as well as any CLI analysis options.
+
+```yaml
+providerConfig:
+  name: "java"
+  binaryPath: "/path/to/java/binary"
+  useSocket: true
+  initConfig:
+    location: "/path/to/application/source/code"
+    analysisMode: "full"
+    providerSpecificConfig:
+      name: "java" 
+      bundles: "/path/to/bundle"
+      lspServerPath: "/path/to/language/server/binary"
+      depOpenSourceLabelsFile: "/path/to/maven.default.index"
+      dependencyProviderPath: "/path/to/dependency/provider/binary"
+analysisConfig:
+  rules:
+    targets:
+    - "cloud-readiness"
+    - "quarkus"
+    sources:
+    - ""
+    labelSelectors:
+    - ""
+    rulesets:
+    - "path/to/custom-rules"
+    - "path/to/other/custom-rules"
+    useDefaultRules: false
+    analyzeKnownLibraries: true
+  scope:
+    depAanlysis: true
+    withKnownLibs: false
+    incidentSelector: "!package=com.example.apps"
+  output:
+    location: "path/to/output-dir"
+    skipStaticReport: false
+    overwrite: true
+    jsonOutput: false
+
+```
+
+*A profile config will take priority over the CLI analysis config.*.  
+
+*Analysis input flags will take priority over the CLI analysis config.*
+
+
+### Backwards Compatibility 
+
+To ensure backwards compatibility with the current subcommands, the Kantra CLI will 
+accept current input values that will then be set to the refactored subcommand/option. 
+For example:
+
+`kantra analyze --list-targets`
+
+The refactored CLI will read this input to the new `rules` subcommand. 
+
+
+## Test Plan
+
+- Verify analysis results from passing in config for analysis options.
+
+- Verify results from new subcommands, such as `list-providers` and `list-targets`.
+
+- Verify backwards compatibility with previous subcommands.
+
+
+## Drawbacks
+
+Some current CLI options will be moved to a new subcommand.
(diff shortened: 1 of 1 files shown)


enhancements/kantra/subcommands/README.md
```diff
@@ -0,0 +1,213 @@
...
+- `kantra analyze --input <path/to/app> --output <output-dir> --target <target>`
+
+#### rules
+
+`kantra rules` will provide options for interacting with rules and rulesets for 
+analysis configuration. This includes both default rulesets and custom rulesets. 
+
+The current `test` subcommand will fall under this new subcommand.
+
+- `kantra rules --list-targets`
```
shawn-hurley: The list of available rules will probably need to be in a config file, you may want to mention that as well?

enhancements/kantra/subcommands/README.md
```diff
@@ -0,0 +1,213 @@
...
+
+- `kantra config --list-profiles`
+
+#### provider
+
+`kantra provider` will handle functionality around using providers, including 
+installing and removing providers. This will be included once Kantra supports 
+[socket communication](https://github.com/konveyor/enhancements/pull/231).
+
+- `kantra provider --list-providers`
```
shawn-hurley: Can these be subcommands instead of flags that do different command actions?

enhancements/kantra/subcommands/README.md
shawn-hurley: Is there a world where you would discover new rulesets? maybe it makes sense to have this be a subcommand rather than a flag?

shawn-hurley (commented):
Overall this is a good start, I think it would make sense to add a concrete type or example for the configuration's (provider specific config, kantra analysis config, provier/ruleset location for discovery??) and where they would live, and the directory structure. 

I think it is also worth figuring out the provider install bits, and where it goes, what is the install mechanism, and any other details.

enhancements/kantra/subcommands/README.md
eemcmullan: Why is that? We currently parse target labels from the downloaded default rulesets. Do you see this changing?

eemcmullan: Do you mean an option to discover any rulesets you may not have downloaded?

enhancements/kantra/subcommands/README.md
```diff
@@ -0,0 +1,213 @@
...
+`kantra generate` will remain as is.
+
+
+### Analysis Config File
+
+Below is an example config file that can be used to read and set all
+analysis options from the provider, as well as any CLI analysis options.
+
+```yaml
+providerConfig:
```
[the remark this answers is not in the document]
eemcmullan: They could live somewhere such as KANTRA_DIR/provider/java/javaProviderConfig.yaml for example.
wdyt?

enhancements/kantra/subcommands/README.md
shawn-hurley: I actually have a couple of thoughts:
1. I would like to treat non-default rulesets the same as default rulesets; the default ones just get dropped on disk by default. They shouldn't be treated any differently from another ruleset, IMO.
2. Can we remove the concept of source and target, and show the labels? My thinking is that it doesn't matter what the source is; if I am running a ruleset, I want to find all the issues the ruleset flags. Similarly, the target doesn't make sense. If I had a ruleset for, say, best practices, what would the target be? A good indication that these are not obvious or generically useful is looking at the ask for the metadata fields and the subsequent ruleset changes those folks made. 

Let's focus on making the labels easier to use, and make an easy way to create a label selector or something?

shawn-hurley: or a common place like a repo that has links to other rulesets or something. A defined mechanism for providing rulesets was where my head was at, but we can come back to this concept later.

enhancements/kantra/subcommands/README.md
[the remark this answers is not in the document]
shawn-hurley: yeah, that works but if it is in the `provider/java` directory, it can probably just be config.yaml.

That way, for a given provider X, we look in `provider/x/config.yaml` for its config, and we don't have to hard-code anything.

enhancements/kantra/subcommands/README.md
```diff
@@ -0,0 +1,213 @@
...
+*Analysis input flags will take priority over the CLI analysis config.*
+
+
+### Backwards Compatibility 
+
+To ensure backwards compatibility with the current subcommands, the Kantra CLI will 
+accept current input values that will then be set to the refactored subcommand/option. 
+For example:
+
+`kantra analyze --list-targets`
```
shawn-hurley, line 179: Isn't there a way to hide subcommands and flags in cobra? I wonder if we should hide these options. 

You can still use them, but we won't show them in the help text.

enhancements/kantra/subcommands/README.md
```diff
@@ -0,0 +1,213 @@
...
+#### analyze
+
+`kantra analyze` will only be responsible for options involving running analysis.
+Such as:
+
+- `kantra analyze --config-file <path/to/config> --output <output-dir>`
+
+*There will also be an undetermined default path to this file.*
+
+- `kantra analyze --input <path/to/app> --output <output-dir> --target <target>`
```
jmle, line 92: So we basically will keep the current way of executing kantra but also have the option of relying on an external config file for analysis?

enhancements/kantra/subcommands/README.md
```diff
@@ -0,0 +1,213 @@
...
+
+#### rules
+
+`kantra rules` will provide options for interacting with rules and rulesets for 
+analysis configuration. This includes both default rulesets and custom rulesets. 
+
+The current `test` subcommand will fall under this new subcommand.
```
jmle: +1

enhancements/kantra/subcommands/README.md
jmle: Nothing super important, but I think it would be really cool to have our own completion script, like `kubectl` has

enhancements/kantra/subcommands/README.md
jmle: If we keep the current way too, it would be cool to be able to generate that config file automatically when running an analysis for the first time, and later be able to change it if needed.

I also think that having only the `--config-file` way of running the analysis would be a bit arid for the user.

jmle (commented):
I'm in for the refactoring and creating subcommands, just a bit concerned that relying on an external (and fairly complex and low-level at times) config file will scare users away. If we keep the current way of doing analysis, which I think is more user-friendly, and add the option of passing the config file, I think it would be better.

Please tell me if I'm missing something.

enhancements/kantra/subcommands/README.md
eemcmullan: Correct. I think it will work in a similar way the the --profile option works today. 
Also +1 with suggestion of saving the analysisConfig file is one does not exist. Or maybe this is an option to set. Would like thoughts on this.

enhancements/kantra/subcommands/README.md
eemcmullan: Worth opening as a separate RFE in kantra outside of this enhancement?

enhancements/kantra/subcommands/README.md
shawn-hurley: I think there are two things here:

1. The configuration files need to be defaulted and will live in a configuration directory. The tool no longer hard-codes values for these. 
2. When you install or set up a provider, the config file should be automatically created by that process.
3. The user never sees the config file unless they have to. 
4. We need to simplify the config file, because many things should be defaultable that are not (when running in a container image). 

Brings up a good point: that we would have to consider that the config for hybrid mode and the config for local/socket mode would be different. We would need to add some way of getting the right values for the right mode of running.

enhancements/kantra/subcommands/README.md
```diff
@@ -0,0 +1,220 @@
...
+
+## Implementation Details
+
+### Subcommands:
+
+#### analyze
+
+`kantra analyze` will only be responsible for options involving running analysis.
+
+*There will also be an undetermined default path to this file which will contain
```
shawn-hurley: note: I think we need to have this default path figured out. 

I think it should live in ~/.kantra or ~/.config/kantra IMO

enhancements/kantra/subcommands/README.md
```diff
@@ -0,0 +1,220 @@
...
+  initConfig:
+    location: "/path/to/application/source/code"
+    analysisMode: "full"
+    providerSpecificConfig:
+      name: "java" 
+      bundles: "/path/to/bundle"
+      lspServerPath: "/path/to/language/server/binary"
+      depOpenSourceLabelsFile: "/path/to/maven.default.index"
+      dependencyProviderPath: "/path/to/dependency/provider/binary"
+analysisConfig:
```
shawn-hurley: I wonder if the analysis config should be its own top-level config, and the provider Kantra config should have:

1. The Provider config
2. The location of the default rules for the provider)
3. Potentially the OpenAPI schema for the conditions


It seems to me that the analysis config here looks a lot like a profile?

enhancements/kantra/subcommands/README.md
```diff
@@ -0,0 +1,220 @@
...
+    location: "path/to/output-dir"
+    skipStaticReport: false
+    overwrite: true
+    jsonOutput: false
+
+```
+
+*A profile config will take priority over the CLI analysis config.*.  
+
+*Analysis input flags will take priority over the CLI analysis config.*
```
shawn-hurley, line 170: I kind of think that if someone adds a flag, we don't use any of the configs. It will be really confusing to use the CLI and expect the default option to be applied, but the config changes it without any real insight.

shawn-hurley (changes requested):
It might be helpful to fully see the flags for each subcommand and what they will do.

The high-level descriptions make a ton of sense, and I think it's good to go; I just want to dive into the details a little to verify the UX. 

What are your thoughts on the provider config not having anything to do with the analysis config, is there a usecase that you are thinking of that I am unaware of?

enhancements/kantra/subcommands/README.md
eemcmullan: Makes sense

enhancements/kantra/subcommands/README.md
eemcmullan: That makes sense. I think when I added this, I imagined the analysisConfig would contain any analysis options, which now is what a profile does. Does it make sense to keep these 2 things separate? So the kantraDir would look like: 

```
.kantra
├── provider
│   ├── java
│   │   ├── config.yaml
│   ├── profiles
│   │   ├── profile-example
│   │   |   ├── profile.yaml
```

shawn-hurley (approved):
This looks good, and I think it should be ready to merge!

enhancements/kantra/subcommands/README.md
```diff
@@ -0,0 +1,196 @@
+---
+title: Kantra Subcommands and Analysis Config
+authors:
+  - "@eemcmullan"
+reviewers:
+  - TBD
+approvers:
+  - TBD
+creation-date: 2025-11-03
+last-updated: 2025-11-06
```
jmle: Let's update the dates

enhancements/kantra/subcommands/README.md
```diff
@@ -0,0 +1,196 @@
...
+## Implementation Details
+
+### Subcommands:
+
+#### analyze
+
+`kantra analyze` will only be responsible for options involving running analysis.
+
+*There will also be a default path in `~/.kantra` to this file which will contain
+the analysis configuration files from provider install.*
```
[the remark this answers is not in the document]
jmle: Maybe we can state that the precedence is the same as the $KANTRA_DIR precedence (I guess that's it)?

enhancements/kantra/subcommands/README.md
```diff
@@ -0,0 +1,196 @@
...
+### Analysis Config File
+
+Below is an example config file that can be used to read and set analysis options 
+from the provider, as well as any CLI analysis options. This will be created during 
+a provider install. The proposed location for these config files 
+is `KANTRA_DIR/provider/<provider-name>/config.yaml`
+
+Because analysis can be run in containerless mode or hybrid mode (providers in container),
+there will need to be separate config files for each. We can differentiate by config
+file name, or add a value to the config file, to indicate which mode will be used.
```
jmle, line 150: Maybe we should decide which one of these we're going to use?

enhancements/kantra/subcommands/README.md
```diff
@@ -0,0 +1,196 @@
+---
+title: Kantra Subcommands and Analysis Config
+authors:
+  - "@eemcmullan"
+reviewers:
+  - TBD
+approvers:
+  - TBD
```
jmle, line 9: I guess let's add these too

jmle (approved):
Looks good in general, just a few things

PR #259, opened 2026-01-23
Enhancement: Analyzer MCP Server

JonahSussman:
## Summary

Enhancement proposal for a standalone MCP server that wraps kai-analyzer-rpc to expose Konveyor's analysis capabilities to AI agents and automation tools.

## Implementation PRs

- [konveyor/kai#926](https://github.com/konveyor/kai/pull/926) — Export Analyzer accessors (merged)
- [konveyor-ecosystem/analyzer-mcp#1](https://github.com/konveyor-ecosystem/analyzer-mcp/pull/1) — MCP server implementation (9 tools, 59 tests)
- [konveyor/koncur#63](https://github.com/konveyor/koncur/pull/63) — MCP and RPC target integration for e2e testing
- [konveyor/ci#207](https://github.com/konveyor/ci/pull/207) — CI action and image build matrix for e2e tests

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive design documentation for the Analyzer MCP Server, defining 9 MCP tools for analysis execution, cached result retrieval, incident querying, and provider management. Covers supported transports (stdio and HTTP), bearer-token authentication, security handling, resource-exhaustion mitigations, testing strategy, upgrade/downgrade procedures, and infrastructure deployment requirements.

### DIFF
--- a/enhancements/analyzer-mcp/README.md
+++ b/enhancements/analyzer-mcp/README.md
@@ -23,6 +23,12 @@ superseded-by:
 
 Add a standalone Model Context Protocol (MCP) server that wraps the kai-analyzer-rpc engine to expose Konveyor's analysis capabilities to AI agents, IDEs, and automation tools.
 
+**Implementation PRs:**
+- [konveyor/kai#926](https://github.com/konveyor/kai/pull/926) - Export Analyzer accessors (merged)
+- [konveyor-ecosystem/analyzer-mcp#1](https://github.com/konveyor-ecosystem/analyzer-mcp/pull/1) - MCP server implementation
+- [konveyor/koncur#63](https://github.com/konveyor/koncur/pull/63) - MCP and RPC target integration for e2e testing
+- [konveyor/ci#207](https://github.com/konveyor/ci/pull/207) - CI action and image build matrix
+
 ## Release Signoff Checklist
 
 - [x] Enhancement is `implementable`
@@ -32,8 +38,7 @@ Add a standalone Model Context Protocol (MCP) server that wraps the kai-analyzer
 
 ## Open Questions [optional]
 
-1. Should we support additional transports beyond stdio and HTTP (e.g., WebSocket)?
-2. What is the long-term strategy for versioning the MCP tools as the analyzer API evolves?
+1. What is the long-term strategy for versioning the MCP tools as the analyzer API evolves?
 
 ## Summary
 
@@ -42,15 +47,15 @@ This enhancement adds a standalone MCP server (`analyzer-mcp`) in a dedicated re
 The server exposes 9 tools through MCP, supports stdio and HTTP transports with bearer token authentication, and integrates with [koncur](https://github.com/konveyor/koncur) for end-to-end testing.
 
 Key capabilities exposed through MCP:
-- **analyze**: Execute full analysis with configurable rules, label selectors, and scoping
-- **get_analysis_results**: Retrieve cached results from the last analysis run
-- **analyze_incidents**: Query and filter incidents by file, rule, or category
-- **list_rules**: Enumerate available rules with metadata and labels
-- **validate_rules**: Validate rule syntax and structure
-- **list_providers**: List available analysis providers and their capabilities
-- **get_dependencies**: Extract dependency trees from analyzed projects
-- **get_migration_context**: Infer migration sources and targets from rules
-- **notify_file_changes**: Notify providers of file changes for incremental analysis
+- `analyze`: Execute full analysis with configurable rules, label selectors, and scoping
+- `get_analysis_results`: Retrieve cached results from the last analysis run
+- `analyze_incidents`: Query and filter incidents by file, rule, or category
+- `list_rules`: Enumerate available rules with metadata and labels
+- `validate_rules`: Validate rule syntax and structure
+- `list_providers`: List available analysis providers and their capabilities
+- `get_dependencies`: Extract dependency trees from analyzed projects
+- `get_migration_context`: Infer migration sources and targets from rules
+- `notify_file_changes`: Notify providers of file changes for incremental analysis
 
 ## Motivation
 
@@ -222,7 +227,7 @@ analyzer-mcp/
 - 2026-04-07: Team meeting decided to move to separate repo wrapping kai-analyzer-rpc
 - 2026-04-07: Closed analyzer-lsp PR, created [konveyor-ecosystem/analyzer-mcp](https://github.com/konveyor-ecosystem/analyzer-mcp) repo
 - 2026-04-16: Upstream kai-analyzer exports merged ([konveyor/kai#926](https://github.com/konveyor/kai/pull/926))
-- 2026-04-16: Initial implementation PR ([konveyor-ecosystem/analyzer-mcp#1](https://github.com/konveyor-ecosystem/analyzer-mcp/pull/1)) — 9 tools, 59 tests, stdio + HTTP transport
+- 2026-04-16: Initial implementation PR ([konveyor-ecosystem/analyzer-mcp#1](https://github.com/konveyor-ecosystem/analyzer-mcp/pull/1)) - 9 tools, 59 tests, stdio + HTTP transport
 - 2026-04-16: Koncur MCP/RPC target integration ([konveyor/koncur#63](https://github.com/konveyor/koncur/pull/63))
 - 2026-04-16: CI e2e infrastructure ([konveyor/ci#207](https://github.com/konveyor/ci/pull/207))
 
@@ -245,7 +250,7 @@ Only implement stdio transport, skip HTTP server.
 **Pros:** Simpler, no authentication concerns, smaller security surface.
 **Cons:** No remote access for CI/CD, can't deploy as a service.
 
-## Infrastructure Needed [optional]
+## Infrastructure Needed
 
 - [x] New repository: [konveyor-ecosystem/analyzer-mcp](https://github.com/konveyor-ecosystem/analyzer-mcp)
 - [ ] Container image: `quay.io/konveyor/analyzer-mcp`
--- a/enhancements/analyzer-mcp/README.md
+++ b/enhancements/analyzer-mcp/README.md
@@ -1,0 +1,314 @@
+---
+title: mcp-server-integration
+authors:
+  - "@JonahSussman"
+reviewers:
+  - TBD
+approvers:
+  - TBD
+creation-date: 2026-01-23
+last-updated: 2026-01-23
+status: implementable
+see-also:
+  - "https://modelcontextprotocol.io/"
+  - "https://github.com/modelcontextprotocol/specification"
+replaces:
+  - N/A
+superseded-by:
+  - N/A
+---
+
+# MCP Server Integration
+
+Add a Model Context Protocol (MCP) server to analyzer-lsp to expose analysis capabilities to AI assistants and other MCP clients, enabling programmatic access to rule analysis, dependency extraction, and code inspection features.
+
+## Release Signoff Checklist
+
+- [x] Enhancement is `implementable`
+- [x] Design details are appropriately documented from clear requirements
+- [x] Test plan is defined
+- [ ] User-facing documentation is created
+
+## Open Questions [optional]
+
+1. Should we support additional transports beyond stdio and HTTP (e.g., WebSocket)?
+2. What is the long-term strategy for versioning the MCP tools as the analyzer-lsp API evolves?
+3. Should we implement caching for frequently-run analyses to improve performance?
+
+## Summary
+
+This enhancement adds a Model Context Protocol (MCP) server implementation to analyzer-lsp, exposing six core analysis tools through a standardized protocol. The MCP server enables AI assistants (like Claude Desktop), IDEs, and other automation tools to programmatically analyze codebases, validate rules, extract dependencies, and query analysis results.
+
+The implementation supports both stdio transport (for local CLI usage and IDE integration) and HTTP transport (for remote access and CI/CD integration), with OAuth 2.1 authentication for the HTTP endpoint. This unlocks new use cases including AI-assisted code migration, automated rule development, and integrated analysis workflows.
+
+Key capabilities exposed through MCP:
+- **analyze_run**: Execute full analysis on codebases with configurable rules and filters
+- **rules_list**: Enumerate available rules with metadata
+- **rules_validate**: Validate rule syntax and structure
+- **dependencies_get**: Extract dependency trees from projects
+- **providers_list**: List available analysis providers and their capabilities
+- **analyze_incidents**: Query and filter incidents from analysis results
+
+## Motivation
+
+Current analyzer-lsp usage requires direct CLI invocation or LSP server integration, limiting automation and AI-assisted workflows. Users who want to integrate analysis into custom tooling, CI/CD pipelines, or AI agents must parse command-line output or interact with the LSP protocol, which is designed for IDEs rather than programmatic access.
+
+The Model Context Protocol provides a standardized way for AI assistants and automation tools to interact with external services. By implementing an MCP server, we enable:
+
+1. **AI-Assisted Migration**: AI assistants can run analyses, interpret results, and suggest code fixes
+2. **Custom Workflows**: Developers can build automation tools that leverage analyzer-lsp capabilities
+3. **CI/CD Integration**: Remote HTTP access enables analysis in containerized and distributed environments
+4. **Interactive Exploration**: Users can query rules, validate syntax, and explore dependencies conversationally
+5. **Cross-Platform Access**: Standard protocol works across different tools and platforms
+
+### Goals
+
+- Expose core analyzer-lsp functionality through MCP protocol
+- Support both local (stdio) and remote (HTTP) access patterns
+- Maintain security through OAuth 2.1 authentication for HTTP
+- Provide clear documentation for integration with Claude Desktop and other MCP clients
+- Maintain API compatibility with existing analyzer-lsp commands
+- Enable seamless integration with AI assistants and automation tools
+
+### Non-Goals
+
+- Replace the existing CLI interface (MCP is complementary)
+- Replace the LSP server for IDE integration
+- Implement real-time streaming of analysis progress
+- Support WebSocket or other transports in initial implementation
+- Provide a web UI for the MCP server
+- Cache or persist analysis results (stateless operation)
+- Modify core analyzer-lsp analysis engine behavior
+
+## Proposal
+
+Implement an MCP server as a new command (`cmd/analyzer/mcp/`) that wraps existing analyzer-lsp functionality and exposes it through the Model Context Protocol. The server will support two transports: stdio for local usage and HTTP for remote access.
+
+### User Stories [optional]
+
+#### Story 1: AI-Assisted Code Migration
+
+As a developer migrating a Java EE application to Quarkus, I want to use Claude Desktop to analyze my codebase, explain migration issues, and suggest fixes.
+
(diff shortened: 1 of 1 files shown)


enhancements/analyzer-mcp/README.md
```diff
@@ -0,0 +1,314 @@
...
+- Replace the LSP server for IDE integration
+- Implement real-time streaming of analysis progress
+- Support WebSocket or other transports in initial implementation
+- Provide a web UI for the MCP server
+- Cache or persist analysis results (stateless operation)
+- Modify core analyzer-lsp analysis engine behavior
+
+## Proposal
+
+Implement an MCP server as a new command (`cmd/analyzer/mcp/`) that wraps existing analyzer-lsp functionality and exposes it through the Model Context Protocol. The server will support two transports: stdio for local usage and HTTP for remote access.
```
shawn-hurley: Would it make sense to create this in a new repo, as another integrator of analyzer-lsp?

Is there a need for this to be in the tree?

enhancements/analyzer-mcp/README.md
```diff
@@ -0,0 +1,314 @@
...
+├── dependencies.go      # dependencies_get implementation
+├── providers.go         # providers_list implementation
+├── incidents.go         # analyze_incidents implementation
+└── transport_http.go    # HTTP server with OAuth 2.1
+```
+
+**Key Design Decisions:**
+
+1. **Reuse Existing Code**: All tools delegate to existing analyzer-lsp libraries (`parser`, `engine`, `provider`, etc.) rather than reimplementing logic
+2. **Stateless Operation**: Each tool invocation is independent; no session state is maintained
```
shawn-hurley: How is this done when running providers?

enhancements/analyzer-mcp/README.md
```diff
@@ -0,0 +1,314 @@
...
+**Cons:**
+- No remote access for CI/CD
+- Can't deploy as a service
+- Limited to local usage only
+
+## Infrastructure Needed [optional]
+
+**Development:**
+- No new infrastructure required
+- Existing analyzer-lsp repository and CI pipeline sufficient
```
shawn-hurley: I don't believe that this is the case

shawn-hurley (changes requested):

enhancements/analyzer-mcp/README.md
tsanders-rh: @shawn-hurley Good question about stateless operation with providers. I don't think true statelessness is viable here.

**The Performance Problem**:
- Provider `Init()` + `Prepare()` phase takes 10s-5min depending on codebase size (symbol cache population)
- If each MCP tool invocation is truly independent, every query would re-run this = 5min wait for AI assistants = terrible UX
- The `Prepare()` phase is specifically designed to pre-populate caches for subsequent queries

**What I Think We Need**:
```
Stateful session + Stateless tools
│
├─ Session: Long-running analyzer-lsp process (stays alive between queries)
│   - Init() once on first query
│   - Prepare() once, cache persists
│   - File watching + incremental updates via NotifyFileChanges()
│
└─ Tools: Each MCP call is independent (get_violations, get_dependencies, etc.)
    - Query against cached state
    - No side effects between tool calls
    - Idempotent operations
```

**Proposal**:
- Clarify in the enhancement doc: "MCP server maintains a stateful session per workspace, but individual tool invocations are stateless operations against that session"
- Add session lifecycle management (start on first query, configurable timeout, explicit cleanup)
- Design for eventual multi-workspace support (one session per workspace root)

**Questions**:
1. Does this stateful-session model align with your concerns?
2. Should we add explicit session management tools (`start_session`, `get_session_status`, `end_session`)?
3. How do we handle session cleanup? (Timeout after N minutes idle? Explicit user command?)

Related to your infrastructure question - this definitely requires process management infrastructure.

enhancements/analyzer-mcp/README.md
tsanders-rh: Thinking through the performance/UX requirements, here's what we'd actually need:

**Phase 1 (MVP) - Minimal Infrastructure**:
- ✅ Long-running process management (not batch execution like current CLI)
- ✅ GRPC client libraries in MCP server (TypeScript/Python)
- ✅ MCP protocol implementation (stdio/HTTP transports)
- ✅ Basic session state management

**Phase 2 (Production) - Significant Infrastructure**:
- ✅ Caching layer for violation results (in-memory or Redis for multi-process)
- ✅ File watching system (chokidar, watchdog, or similar)
- ✅ Cache invalidation logic when files change
- ✅ Background re-analysis orchestration
- ✅ Indexes for fast querying (by file, rule, category, severity)

**Phase 3 (Advanced) - Optional**:
- ⚠️ Possible embedding/vector store for semantic search
- ⚠️ Multi-workspace coordination if shared server

**Why This Infrastructure Matters**:

Without caching (Phase 2), every MCP query would be:
```
AI Assistant: "Show me violations in Main.java"
MCP Server: "Running full analysis... (5 minutes later)... here are results"
AI Assistant: "Now show me violations in Controller.java"
MCP Server: "Running full analysis again... (5 minutes later)... here are results"
```

This is unusable. We need:
```
AI Assistant: "Show me violations in Main.java"
MCP Server: "Checking cache... found 12 violations (200ms)"
AI Assistant: "Now show me violations in Controller.java"
MCP Server: "Checking cache... found 8 violations (150ms)"
```

**Recommendation**: Update the proposal to:
- Phase 1: "Minimal infrastructure - long-running process, basic GRPC wrapper"
- Phase 2: "Production infrastructure - caching, file watching, incremental updates"
- Acknowledge this is more than a thin wrapper, but infrastructure is scoped and manageable

**Questions**:
1. Does this align with your infrastructure concerns?
2. Should Phase 1 include basic caching, or can we defer to Phase 2?
3. What's the expected deployment model? (One process per workspace? Shared service? Embedded in IDE?)

tsanders-rh:
## Strategic Opportunity: MCP as Platform Play

I want to raise a strategic framing that's missing from this proposal. Currently it's positioned as "add MCP support to analyzer," but there's a much bigger opportunity here.

### The Platform Vision

**Current Framing**: "Analyzer gets another output format (MCP in addition to GRPC)"

**Alternative Framing**: "Konveyor becomes the migration context platform that powers AI development tools"

**What This Means**:

Think of it like:
- **Stripe** doesn't compete on checkout UI - they provide payment infrastructure
- **Auth0** doesn't compete on login forms - they provide auth infrastructure
- **Konveyor** shouldn't compete on IDE UX - we provide migration context infrastructure

```
┌─────────────────────────────────────┐
│  Konveyor Migration Context Platform│
│  (analyzer-lsp + rules + analysis)  │
└────────────────┬────────────────────┘
                 │ MCP + GRPC APIs
                 ▼
┌─────────┬──────────┬─────────┬──────────┐
│ Cursor  │ Continue │ Copilot │ Custom   │
│ (owns   │ (owns    │ (owns   │ Tools    │
│  UX)    │  UX)     │  UX)    │          │
└─────────┴──────────┴─────────┴──────────┘

We provide: Migration context
They provide: Developer experience
```

### Why This Matters

**Network Effects**:
- VSCode extension = Linear growth (each install = 1 user)
- MCP platform = Exponential growth (each tool integration = entire user base)
  - Cursor integration → 100k+ users
  - Copilot integration → 1M+ users
  - Continue integration → 50k+ users

**Sustainable Advantage**:
- Hard to replicate: 2,400 community rules, enterprise patterns, multi-language analysis
- Easy to replicate: VSCode extension UI, AI chat interface
- **Focus on the hard part, expose via MCP, let tools handle the easy part**

### What This Changes About the Proposal

**Success Metrics Should Include**:
- ✅ MCP server works (current proposal)
- ✅ **At least 1 AI tool integrates** (new - Cursor, Continue, or Claude Desktop)
- ✅ **Users query via AI tools** (validation of platform model)
- ✅ **AI tool partners find value** (testimonials, co-marketing potential)


**Positioning Should Emphasize**:
- "Konveyor powers migration intelligence in your AI assistant"
- Not just "Konveyor has an MCP server"

### Questions for the Team

1. **How does this affect VSCode extension strategy?** If MCP is the platform play:
   - Extension becomes one consumer among many
   - Focus on power-user features (visualizations, batch operations)
   - Don't compete with Cursor/Continue on UX

I think there's real strategic value here beyond just "another integration." Happy to discuss further.

tsanders-rh:
## Technical Gap: Source/Target Migration Context

One technical issue I noticed - the proposal doesn't address how source/target migration context works with MCP.

**The Problem**:

Currently, source/target is configured via `--label-selector` at analyzer CLI startup:
```bash
analyzer-lsp --label-selector "konveyor.io/source=eap6" \
             --label-selector "konveyor.io/target=eap8"
```

From the codebase:
- ❌ No GRPC API to query current label selectors at runtime
- ❌ No GRPC API to change label selectors after Init()
- ⚠️ Can infer from violation labels, but not directly queryable

**Why This Matters for MCP**:

AI assistants will want to know:
- "What migration am I analyzing?" (EAP6→EAP8? Spring Boot 2→3?)
- "Which rules are active for this migration path?"
- "Should I suggest fixes for this violation?" (depends on target)

**Example Scenario**:
```
AI: "What migration context are we in?"
MCP: ??? (no API to query this)

AI: "Show me javax.servlet violations"
MCP: Returns violations (but doesn't explain why they matter)
AI: Should I tell the user to fix this?
MCP: ??? (depends on target, which we can't query directly)
```

**Potential Solutions**:

**Option 1**: Store in MCP server config
- Read source/target from `.konveyor.yaml` or similar
- Pass to analyzer on Init() via CLI wrapper
- Cache in MCP server memory
- Return via optional `get_migration_context` tool

**Option 2**: Infer from violations
- Parse labels from first analysis output
- Extract `konveyor.io/source` and `konveyor.io/target` labels
- Cache and return to clients

**Option 3**: Add GRPC API upstream (longer-term)
- Enhance analyzer-lsp to expose label selector state
- Add `GetConfiguration()` or similar GRPC method
- MCP can query directly

**Option 4**: Document limitation
- "MCP server must be configured with source/target upfront"
- Clients provide via config, not runtime query

enhancements/analyzer-mcp/README.md
shawn-hurley: I think it starts to look very much like `kai_analyzer_rpc`. The interaction between VSCode and analyzer-lsp is pretty "stateless," aside from the "hey, this updated, you should re-look" message.

That is an important bit, though. We need to know when the files were updated so we can re-gen the cache for those files.

So I don't know if the session is the right layer to understand the state at, if that makes sense? 

We have some cache invalidation in `kai_analyzer_rpc` that invalidates the cache when you request a specific file. I don't know if this would be desirable from a performance standpoint for this use case. 

I think that this idea is valuable, and obviously necessary for where I think we should be in terms of agentic workflows and such. I just think there are many more design questions here that we should at least capture. 

I guess we could do an MVP of an option and see where we get, but we do have some prior art here, and I think we should probably take the learnings from that.

shawn-hurley:
> ✅ Users query via AI tools (validation of platform model)

I would like to consider a world where the user never queries this information directly as a primary use case. I think we should focus on enabling agents to reliably use this information to generate better code. I personally would like to have some way to see that in an automated test early. I feel like that should be a big lesson learned from the initial kai work: we didn't have a good way to know the impact of any given change to the query. I could easily see a similar issue arising here. 

> Technical Gap: Source/Target Migration Context

Of note on this section, these labels are purely from the rulesets, and are not stored anywhere else (well, I guess in the hub database, but I don't think that is relevant here). I think any solution to this problem requires a way to manage rulesets for the MCP server. 

I think that we should probably break that down into two categories:

1. Default rulesets that should come out of the box.
2. Custom rulesets or rulesets downloaded/generated or whatever that you would also like to include. 

It is worth noting that you *DO NOT* need a source/target or label-selector to run rules from rulesets. If you do not specify one of these options in the analysis-option, we run all rules that have accessible providers for the capabilities in the rules. 

I would argue that we should start any solution to this problem with a user experience centered on the ruleset, and allow the label-selector to filter a large set of rules.

> Alternative Framing: "Konveyor becomes the migration context platform that powers AI development tools."

I agree with this framing and think it is an excellent goal. As I alluded to in another comment, this is exactly the way we need to go IMO.

Should we consider alternatives to an MCP server, though?

I think that we should consider:

1. https://www.utcp.io 
2. https://agentskills.io/home 

I am sure there are others that folks have heard of or seen that I am not aware of, might be nice to at least write down why an MCP server vs other options in the enhancement.

I personally had a hard time seeing how we could make an MCP truly useful in an agentic system when I tried to build one awhile ago. The protocol itself doesn't seem to have changed drastically, so I think it may still be the case. You just don't have enough information for the system to understand what it should do with the data. 

It also seems that there is a lot of talk, rightly or wrongly, about MCP servers having a real scale problem. I don't know how real that is, but it's something we should at least consider. 

Sorry for the wall of text. I do wonder if this is a good topic for the next community meeting?

enhancements/analyzer-mcp/README.md
pranavgaikwad: @shawn-hurley I agree with Shawn here. There are many design details to talk out here. But here's something I have been thinking about "sessions". The only distinction between a MCP server and an analysis skill that I see is being able to be stateful. I 100% agree that this MCP server is kai_analyzer_rpc more than it is analyzer_rpc and as a result, I see it being implemented there. However, It wouldn't be ideal to compare IDE <-> kai_analyzer_rpc because there's a strict 1:1 relationship between kai_analyzer_rpc instance and an IDE instance.  That is not true for an MCP server. Imagine an MCP server having potentially more than one workspaces initiated. The core of it still looks like kai_analyzer_rpc, while the session layer differentiates the workspaces.

enhancements/analyzer-mcp/README.md
pranavgaikwad: @tsanders-rh Most of the points here such as caching, process management, grpc are already implemented in kai_analyzer_rpc. The only thing we need is a file watcher (which we currently rely on vscode for) and session management for multiple workspace management. This makes my point in the previous comment even more relevant I think. This MCP server belongs more in kai_analyzer_rpc rather than analyzer-lsp

enhancements/analyzer-mcp/README.md
```diff
@@ -0,0 +1,314 @@
...
+As a developer migrating a Java EE application to Quarkus, I want to use Claude Desktop to analyze my codebase, explain migration issues, and suggest fixes.
+
+1. Configure Claude Desktop with the analyzer-lsp MCP server
+2. Ask Claude: "Analyze my project for Java EE to Quarkus migration issues"
+3. Claude uses `rules_list` to find relevant rules, then `analyze_run` to execute analysis
+4. Claude explains the incidents found and suggests code changes
+5. I can ask follow-up questions like "Show me all logging framework issues"
+6. Claude uses `analyze_incidents` to filter and explain specific issues
+
+#### Story 2: Custom Rule Development Workflow
```
savitharaghunathan: should we keep this story as a part of rule gen mcp? or move rule generation tools to this mcp server and keep them tightly coupled? this is one of the open questions in the PR I have - https://github.com/konveyor/enhancements/pull/275/changes#diff-9c91dda24a5c871d4c538761bf924038c89d202f378209480fdc5323812c0827R40

enhancements/analyzer-mcp/README.md
```diff
@@ -0,0 +1,314 @@
...
+2. **Security Surface**: HTTP transport introduces authentication and network security concerns
+   - Counterpoint: OAuth 2.1 is industry standard, well-understood
+   - Mitigation: Security review before HTTP deployment, TLS required in production
+
+3. **Performance Overhead**: MCP protocol adds JSON serialization/deserialization
+   - Counterpoint: Overhead is minimal (<5%) and acceptable for automation use cases
+   - Mitigation: Direct CLI/LSP remains available for performance-critical usage
+
+4. **Dependency on External Protocol**: Tied to MCP specification evolution
+   - Counterpoint: MCP is backed by Anthropic and growing ecosystem
```
savitharaghunathan: Note to add or FYI: MCP is a part of Agentic AI foundation (AAIF), so its also vendor neutral 
link: https://aaif.io/

enhancements/analyzer-mcp/README.md
```diff
@@ -0,0 +1,314 @@
...
+
+**Initial Release:**
+- MCP server is a new optional component, no upgrade concerns
+- Can be deployed alongside existing analyzer-lsp installations
+- No changes to existing CLI or LSP functionality
+
+**Future Version Compatibility:**
+- Tool input schemas will be versioned in descriptions
+- Breaking changes will result in new tool names (e.g., `analyze_run_v2`)
+- Deprecated tools will remain available for 2+ releases
```
savitharaghunathan: do we have a Konveyor wide policy regarding the number of releases we support? if so, we should follow that. If not, then should discuss about it

enhancements/analyzer-mcp/README.md
savitharaghunathan: The more i think about it, it makes sense to have rules gen tools with analyzer_lsp mcp server.. 

generate rules -> validate -> analyze app -> fix violations 

wdyt?

enhancements/analyzer-mcp/README.md
```diff
@@ -0,0 +1,257 @@
...
+
+## Summary
+
+This enhancement adds a standalone MCP server (`analyzer-mcp`) in a dedicated repository ([konveyor-ecosystem/analyzer-mcp](https://github.com/konveyor-ecosystem/analyzer-mcp)) that wraps the [kai-analyzer-rpc](https://github.com/konveyor/kai/tree/main/kai_analyzer_rpc) engine. By living in a separate repo, the server can import both `analyzer-lsp` and `kai-analyzer` without circular dependency issues, and it gets kai-analyzer's caching, incremental analysis, and provider lifecycle management for free.
+
+The server exposes 9 tools through MCP, supports stdio and HTTP transports with bearer token authentication, and integrates with [koncur](https://github.com/konveyor/koncur) for end-to-end testing.
+
+Key capabilities exposed through MCP:
+- `analyze`: Execute full analysis with configurable rules, label selectors, and scoping
+- `get_analysis_results`: Retrieve cached results from the last analysis run
```
djzager, line 51: https://github.com/konveyor/enhancements/pull/267#discussion_r3080063090 is relevant here. I'm not sure if @pranavgaikwad was saying there should only be one endpoint or if the skills should only use one.

djzager (commented):
I'm excited about this future, the only thing that really caught my eye was the analyze vs get_analysis_results.

enhancements/analyzer-mcp/README.md
```diff
@@ -0,0 +1,257 @@
...
+- `analyze_incidents`: Query and filter incidents by file, rule, or category
+- `list_rules`: Enumerate available rules with metadata and labels
+- `validate_rules`: Validate rule syntax and structure
+- `list_providers`: List available analysis providers and their capabilities
+- `get_dependencies`: Extract dependency trees from analyzed projects
+- `get_migration_context`: Infer migration sources and targets from rules
+- `notify_file_changes`: Notify providers of file changes for incremental analysis
```
pranavgaikwad, line 58: this looks too brittle to be exposed to an LLM. I think this should be handled elsewhere.

enhancements/analyzer-mcp/README.md
pranavgaikwad: Yes, right now, there is no use case for letting an agent run analysis, send file change events, etc. Its best to keep these things deterministic and outside of MCP server. If I understand correctly, right now, our primary use cases for the MCP server are - IDE and a standalone runner in the container. I think...in both these environments we only ever need the `get_analysis_results` function. Am I missing something?

PR #262, opened 2026-03-13
:book: Refactor generic provider into specific providers

jmle:
- Proposal for refactoring the generic provider into their own specific providers (python, nodejs, ...)
- Also committed additional information regarding the current status of the generic provider, from which the enhancement was generated.
- Some open questions appear that I have left there:
  - YAML provider strategy: I guess there are some traces of yaml provider being generic, but we already have an independent yaml provider, if I'm not mistaken.
  - Image topology: we'll be generating individual images
  - Backwards compatibility is a good point, although I guess most people using this are doing so through konveyor/kantra, so probably not a big concern

Addresses https://github.com/konveyor/analyzer-lsp/issues/1013

## Summary by CodeRabbit

* **Documentation**
  * Added detailed enhancement docs describing a planned refactor from a single generic provider to language-specific providers (Go, Python, Node), YAML handling updates, test/CI and packaging implications, transition and upgrade strategies, risks, and implementation notes.

### DIFF
--- a/enhancements/generic-to-specific-providers/generic-provider-to-specific-providers.md
+++ b/enhancements/generic-to-specific-providers/generic-provider-to-specific-providers.md
@@ -8,7 +8,7 @@ reviewers:
 approvers:
   - TBD
 creation-date: 2026-03-13
-last-updated: 2026-03-13
+last-updated: 2026-03-28
 status: provisional
 see-also:
   - "https://github.com/konveyor/analyzer-lsp/issues/1013"
@@ -17,7 +17,7 @@ see-also:
 
 # Refactor Generic Service Client to Language-Specific Providers
 
-This enhancement proposes splitting the current "generic external provider" in analyzer-lsp into separate, language-specific providers (e.g. Go, Python, Node.js), each with its own binary and initialization path. The goal is to align with the reality that LSP server initialization and behavior differ too much across languages to be reliably abstracted behind a single generic client.
+This enhancement proposes splitting the current "generic external provider" in analyzer-lsp into separate, language-specific providers (Go, Python, Node.js), each with its own binary and initialization path. The goal is to align with the reality that LSP server initialization and behavior differ too much across languages to be reliably abstracted behind a single generic client. Maintainer review ([konveyor/enhancements#262](https://github.com/konveyor/enhancements/pull/262)) settled the following: **fully remove** the generic multi-language binary and dispatcher; ship **one container image per language** (smaller images, simpler debugging, fewer unrelated providers in a given deployment); treat YAML analysis as the responsibility of the existing **yq** provider and **deprecate/remove** the generic provider’s `yaml_language_server` path rather than spinning up a parallel YAML LSP provider; accept a **breaking change** for direct consumers of provider binaries (hub/kai/kantra and related tooling are the main integrators—users who need the old layout can pin an older image); and name the Go backend explicitly (e.g. **golang** / `go-external-provider`) so nothing called “generic” implies support for arbitrary language servers.
 
 The YAML `title` is lowercased with spaces replaced by `-`.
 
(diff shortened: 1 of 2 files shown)


enhancements/generic-to-specific-providers/generic-provider-to-specific-providers.md
```diff
@@ -0,0 +1,138 @@
...
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] User-facing documentation is created
+
+## Open Questions
+
+1. **YAML strategy:** Default container settings use `yq-external-provider` for YAML, not the generic provider's `yaml_language_server` config. Should we (a) introduce a dedicated yaml-external-provider that uses the current yaml_language_server code, or (b) leave YAML to yq only and deprecate/remove the yaml_language_server backend from the generic provider?
+
+2. **Image topology:** Should we ship one container image per language (e.g. go-provider, python-provider, nodejs-provider) or a single image that includes multiple provider binaries and an entrypoint that selects by argument/env? The former simplifies per-language ownership and sizing; the latter may reduce CI and deployment surface.
```
eemcmullan: I think the former. Especially with more multi-language support, I think it makes sense to have separate containers to reduce image size and possible irrelevant provider issues (non-running providers).

enhancements/generic-to-specific-providers/generic-provider-to-specific-providers.md
```diff
@@ -0,0 +1,138 @@
...
+
+3. **Backward compatibility:** Should we retain a "generic" binary that supports only the current "generic" (gopls) client so that existing configs using `binaryPath: ".../generic-external-provider"` with `lspServerName: "generic"` continue to work without change, or is a breaking change to provider config acceptable?
```
eemcmullan: If this is the only generic provider, would it make sense to rename this to a `golang` provider? It may be confusing to have a generic provider that is only one supported provider. If we track the configs that would need to be updated, I don't think we'd need backwards compatibility?

enhancements/generic-to-specific-providers/generic-provider-to-specific-providers.md
```diff
@@ -0,0 +1,138 @@
...
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] User-facing documentation is created
+
+## Open Questions
+
+1. **YAML strategy:** Default container settings use `yq-external-provider` for YAML, not the generic provider's `yaml_language_server` config. Should we (a) introduce a dedicated yaml-external-provider that uses the current yaml_language_server code, or (b) leave YAML to yq only and deprecate/remove the yaml_language_server backend from the generic provider?
```
shawn-hurley: I think we should move to yq personally

enhancements/generic-to-specific-providers/generic-provider-to-specific-providers.md
shawn-hurley: I would personally prefer multiple images, it makes debugging a lot eaiser to figure out what is going on

enhancements/generic-to-specific-providers/generic-provider-to-specific-providers.md
```diff
@@ -0,0 +1,138 @@
...
+- **Clarity:** One binary per language (or one clear entrypoint per language) so that configuration, debugging, and ownership are straightforward.
+- **Consistency:** Align Go, Python, and Node.js providers with the existing pattern used by the Java provider (dedicated binary and package).
+- **Maintainability:** Remove the generic provider's runtime switch on `lspServerName` and the `SupportedLanguages` map so that adding or changing a language does not touch a central dispatcher.
+
+Success will be evident when: (1) Go, Python, and Node.js (and optionally YAML) are each served by a dedicated provider binary/package; (2) no single process dispatches to multiple LSP types via `lspServerName`; (3) provider settings and docs reference distinct binaries/images per language; (4) existing tests and demos pass with the new layout.
+
+### Non-Goals
+
+- Changing the analyzer-lsp provider gRPC API or the `BaseClient`/`ServiceClient` interfaces.
+- Removing or rewriting the shared `lsp/base_service_client` layer; it remains the foundation for LSP-based providers.
```
shawn-hurley: What if we did want to make this a goal? 

It seems to me that the current library code is a little hard to follow, and I wonder if thereis an opportunity to make things clearer?

enhancements/generic-to-specific-providers/generic-provider-to-specific-providers.md
```diff
@@ -0,0 +1,138 @@
...
+
+- **Target architecture:** Each of Go, Python, and Node.js (and optionally YAML) gets its own provider package and binary. Each binary implements `BaseClient` (e.g. `Capabilities()`, `Init()`) and constructs exactly one service client type (no map, no switch on `lspServerName`). The analyzer continues to spawn one process per provider config; configs will reference different `binaryPath` values per language. Shared code in `lsp/base_service_client` (and shared capability helpers where applicable) is reused by each provider.
```
shawn-hurley: Once we move them out, can we compare just using the Provider.Server directly vs using the lsp/base_service_client

enhancements/generic-to-specific-providers/generic-provider-to-specific-providers.md
```diff
@@ -0,0 +1,138 @@
...
+  - **Generic provider (remove or shrink):** `external-providers/generic-external-provider/main.go`, `pkg/generic_external_provider/provider.go`, `pkg/server_configurations/constants.go`; move `server_configurations/generic`, `pylsp`, `nodejs`, (and optionally `yaml_language_server`) into new provider modules.
+  - **New providers:** New directories, e.g. `go-external-provider`, `python-external-provider`, `nodejs-external-provider`, each with its own `main.go`, provider type, and one server configuration (no dispatcher).
+  - **Config and build:** Update `provider_container_settings.json`, Makefile, root Dockerfile, generic-provider Dockerfile (or replace with per-language Dockerfiles), `.github/workflows/image-build.yaml`, `.github/workflows/demo-testing.yml`, and testdata/docs as needed.
+  - **Tests:** Move or duplicate `server_configurations/*/service_client_test.go` into the new provider packages; adjust `provider/provider_test.go` if it asserts on generic-provider-specific behavior.
```
shawn-hurley, line 93: e2e tests should also be broken up across the individual repos. It should already be broken up.

enhancements/generic-to-specific-providers/generic-provider-to-specific-providers.md
jmle: Yeah, I left these questions here but I think they have straightforward answers, see PR description

enhancements/generic-to-specific-providers/refactoring-generic-provider-to-specific-providers.md
```diff
@@ -0,0 +1,295 @@
...
+
+- **Config loading:** `provider.GetConfig(settingsFile)` reads provider configs (e.g. `provider_container_settings.json` or YAML). Each entry has `name`, `binaryPath` or `address`, and `initConfig[]`.
+- **Client creation:** For every config with `name != "builtin"`, `provider/lib.GetProviderClient(config, log)` is used. That function delegates to `grpc.NewGRPCClient(config, log)` for all non-builtin providers.
+- **Binary startup (generic only):** In `provider/grpc/provider.go`, `start()` is called when `config.BinaryPath != ""`. It:
+  1. Reserves a free port.
+  2. Reads `lspServerName` from `config.InitConfig[0].ProviderSpecificConfig["lspServerName"]` (default `"generic"`).
+  3. Runs: `exec.CommandContext(ctx, config.BinaryPath, "--port", port, "--name", name)`.
+  4. Dials `localhost:{port}` and returns the gRPC connection and stdout pipe.
+
+So for a settings file that defines three “providers” (e.g. go, python, nodejs) all using the same `binaryPath: "/usr/local/bin/generic-external-provider"`, the analyzer spawns **three separate processes** of the same binary, each with a different `--name` (e.g. `generic`, `pylsp`, `nodejs`).
```
eemcmullan, line 31: My question is, is the plan to completely remove the generic provider, so each provider would have their own binary? cc @shawn-hurley

enhancements/generic-to-specific-providers/generic-provider-to-specific-providers.md
jmle: I guess my question is, would there even be any backwards compatibility issues at all? I'd go for removing it, but maybe I'm missing something

enhancements/generic-to-specific-providers/refactoring-generic-provider-to-specific-providers.md
jmle: That was my understanding from the RFE at least

enhancements/generic-to-specific-providers/generic-provider-to-specific-providers.md
shawn-hurley: I am a plus-one on removing it. At this point, it is an implementation detail of the hub/kai and kantra. 

If someone still needs the image with that particular name, they can use the old one IMO.

enhancements/generic-to-specific-providers/refactoring-generic-provider-to-specific-providers.md
shawn-hurley: Yes, that is what we should be doing IMO.

Right now, the generic provider ASFAICT can not work with a generic language provider, so the name is misleading as it is. Let's handle this if it comes up from a user. If we need to handle the generic case, I think we would need a different solution.

enhancements/generic-to-specific-providers/generic-provider-to-specific-providers.md
```diff
@@ -0,0 +1,138 @@
...
+
+### Security, Risks, and Mitigations
+
+- **Risk: More binaries/images.** More artifacts to build, sign, and distribute. Mitigation: Document the new layout clearly; consider a single image with multiple binaries and an entrypoint if the community prefers to limit the number of images.
+- **Risk: Breaking existing configs.** Users who currently use one `binaryPath` and multiple `initConfig`/`lspServerName` entries will need to point each language to its own binary. Mitigation: Provide a migration path (e.g. retain a "generic" binary for gopls-only) and update default settings and docs.
+- **Risk: Regressions in behavior.** Each language’s logic is moved into a new package. Mitigation: Preserve existing tests (move or re-run), run existing e2e/demos (golang-e2e, python-e2e, nodejs-e2e) against the new providers, and add any missing integration tests.
+- **Security:** No new network exposure or auth model; providers remain out-of-process gRPC services. Binary execution is unchanged (analyzer still spawns one process per provider config). Review by maintainers familiar with the provider and analyzer-lsp security model is sufficient.
```
shawn-hurley, line 103: Note that if we use the server implementation, we should get token auth and ssl over grpc for free

enhancements/generic-to-specific-providers/generic-provider-to-specific-providers.md
```diff
@@ -0,0 +1,138 @@
...
+  - **Tests:** Move or duplicate `server_configurations/*/service_client_test.go` into the new provider packages; adjust `provider/provider_test.go` if it asserts on generic-provider-specific behavior.
+
+- **Constraint:** The analyzer calls `Capabilities()` before `Init()`, so each process must advertise the correct capabilities at startup. With one binary per language, this is natural; no runtime switch is required.
+
+### Security, Risks, and Mitigations
+
+- **Risk: More binaries/images.** More artifacts to build, sign, and distribute. Mitigation: Document the new layout clearly; consider a single image with multiple binaries and an entrypoint if the community prefers to limit the number of images.
+- **Risk: Breaking existing configs.** Users who currently use one `binaryPath` and multiple `initConfig`/`lspServerName` entries will need to point each language to its own binary. Mitigation: Provide a migration path (e.g. retain a "generic" binary for gopls-only) and update default settings and docs.
```
shawn-hurley: Personally, I am comfortable with this as a breaking change, this is an implementation detail of the hub/kai/kantra that doesn't really get exposed.

jortel:
This will require an update in the operator to the python and nodejs extensions to reference the new images.

shawn-hurley (approved):
I am comfortable with the general outline of what is being proposed. 

A question for you and Emily is about keeping the LSP server base and that code, or just using the Server directly and handling each provider's internals more directly.

enhancements/generic-to-specific-providers/generic-provider-to-specific-providers.md
jmle: Okay, I'm including this in the enhancement

enhancements/generic-to-specific-providers/generic-provider-to-specific-providers.md
jmle: Added explicit mention to this

enhancements/generic-to-specific-providers/generic-provider-to-specific-providers.md
jmle: Cool, removing this risk

jmle:
Updated enhancement with PR comments

eemcmullan (approved):

djzager (commented):
Nothing in this enhancement stands out as something I should bulk at. The primary way this impacts the editor-extensions work is in:
1) asset collection. We will need to update what binaries/etc we look for and collect to package the extension.
2) language specific extensions. Work is required in each of the go/javascript extensions to make sure they use the new assets with the correct arguments.

For my later reference (and/or if you want to include it in this doc or follow-up issues), konveyor/editor-extensions needs to:
- Update scripts/collect-assets.js to download new per-language binaries by name
- Update scripts/copy-dist.js to package new binaries into extension dist folders
- Update Go extension (vscode/go/) — new binary path, remove -name "generic" spawn arg, update providerSpecificConfig
- Update JavaScript extension (vscode/javascript/) — new binary path, remove -name "nodejs" spawn arg, update providerSpecificConfig
- Update package.json includedAssetPaths in both extensions

shawn-hurley:
To davids and Jeffs comments, I think they should be sub issues that are tracked on the enhancement tracking issue we should create once this is merged.

@jmle I think that we should be good to merge, if you want to merge it. If there is something you want to add or change though I will leave it up to you :) 

Thanks!

PR #265, opened 2026-03-27
Hub OIDC Provider

jortel:
Proposal to eliminate dependence on keycloak.

[README](https://github.com/jortel/enhancements/blob/hub-oidc-provider/hub-oidc-provider/README.md) 

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive enhancement and design docs for a built-in OIDC authentication system: describes operational modes (local, delegated/broker, LDAP/AD), standard OIDC endpoints and flows, user/role/permission and PAT models, token validation and revocation semantics (including legacy compatibility), credential lifecycle and security/storage expectations, operator configuration and migration guidance from Keycloak, plus testing and upgrade/downgrade plans.

### DIFF
--- a/enhancements/hub-oidc-provider/README.md
+++ b/enhancements/hub-oidc-provider/README.md
@@ -83,12 +83,18 @@ Access tokens may be revoked (effective next refresh).
 Add support for API-Keys (Reference [RFE-266](https://github.com/konveyor/enhancements/issues/266)).
 
 #### Generation
-POST /auth/apikey returns a 256-bit base64-encoded generated key which has been stored in the DB along with associated
-permissions (scopes).
+
+POST /auth/apikey returns a 256-bit base64URL-encoded generated key which has been stored in the DB along 
+with associated permissions (scopes).
+Body:
+```yaml
+expiration: # OPTIONAL
+```
 Returned:
 ```yaml
 id: uint
-key: base64-encoded key. (This is the key presented to the API).
+expiration: # OPTIONAL
+key: cvP1sjff7_X2dCEIzUPf8f0IzKSbwiSDf1dZChZuRxY # base64URL-encoded key. (This is the key presented to the API).
 ```
 
 #### Revocation
--- a/enhancements/hub-oidc-provider/README.md
+++ b/enhancements/hub-oidc-provider/README.md
@@ -40,14 +40,11 @@ Eliminate dependence on Keycloak.
 ### Goals
 
 - **To discontinue dependence on keycloak**.
-- To provide AuthZ _out-of-the-box_.
-- To (optionally) delegate authentication to an external OIDC provider
-- To (optionally) delegate authorization to an external OIDC provider.
 - To discontinue seeding the realm in keycloak.
-- To support user management in the tackle UI.
-  - User CRUD.
-  - Role CRUD with association scopes.
-  - Grant roles to users.
+- To provide OIDC-based Auth _out-of-the-box_.
+- To delegate AuthN, AuthZ to an _external_ OIDC provider. (option)
+- To delegate AuthN, AuthZ to an _external_ LDAP, Active Directory with group mapping. (option)
+- To provide RBAC (users, roles, permissions) management in the inventory.
 
 ### Non-Goals
 
--- a/enhancements/hub-oidc-provider/README.md
+++ b/enhancements/hub-oidc-provider/README.md
@@ -169,7 +169,7 @@ erDiagram
     API_KEY {
         uint id PK
         uint user_id FK
-        string key "unique"
+        string key "unique (encrypted)"
         datetime expiration
     }
 
--- a/enhancements/hub-oidc-provider/README.md
+++ b/enhancements/hub-oidc-provider/README.md
@@ -58,6 +58,16 @@ to delegate authentication and/or authorization to an external provider. The hub
 is augmented to include Users and Roles. Users may be associated to roles and roles may
 be associated to permissions (scopes).
 
+The UI will be updated to use OIDC (instead of keycloak) and be configured to use the
+hub OIDC provider.  The UI will have pages to manage user, roles and permissions.
+
+### Security, Risks, and Mitigations
+
+The [go-oidc](https://github.com/luikyv/go-oidc) package is **OpenID certified** and is actively maintained. It has no reported CVEs.  AI code analysis
+reports no vulnerabilities or backdoors.
+
+## Design Details
+
 High Level Model:
 
 ```mermaid
@@ -287,16 +297,6 @@ sequenceDiagram
 
     deactivate ProtectedAPI
 ```
-
-The UI will be updated to use OIDC (instead of keycloak) and be configured to use the
-hub OIDC provider.  The UI will have pages to manage user, roles and permissions.
-
-### Security, Risks, and Mitigations
-
-The [go-oidc](https://github.com/luikyv/go-oidc) package is **OpenID certified** and is actively maintained. It has no reported CVEs.  AI code analysis
-reports no vulnerabilities or backdoors.
-
-## Design Details
 
 ### Test Plan
 
--- a/enhancements/hub-oidc-provider/README.md
+++ b/enhancements/hub-oidc-provider/README.md
@@ -147,12 +147,13 @@ erDiagram
     USER }o--o{ ROLE : "granted"
     ROLE }o--o{ PERMISSION : "has"
     USER ||--o{ API_KEY : "owns"
+    TASK ||--0{ API_KEY : "owns"
     IDP_IDENTITY ||--|| USER : "EXTERNAL identity"
     IDP_IDENTITY ||--|| TOKEN : "delegated authentication"
 
     USER {
         uint id PK
-        string username "unique"
+        string userid "unique"
         string password "encrypted"
         string email "unique"
     }
@@ -171,7 +172,8 @@ erDiagram
     API_KEY {
         uint id PK
         uint user_id FK
-        string digest "unique"
+        uint task_id FK
+        string digest "unique, hashed-secret"
         datetime expiration
     }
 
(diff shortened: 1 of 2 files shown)


enhancements/hub-oidc-provider/README.md
```diff
@@ -0,0 +1,299 @@
...
+by the operator, enabling branding customization without code changes.
+
+All standard OIDC discovery endpoints will be implemented
+(`.well-known/openid-configuration`, authorization, token, JWKS, userinfo,
+introspection, and revocation), allowing any OIDC-compliant client to integrate
+with the Hub.
+
+#### Default Role Definitions
+
+The system will include predefined roles tailored to common migration project
```
shawn-hurley, line 123: When a new user authenticates with an external authentication system, what default roles/access do they get?

shawn-hurley (approved):
Something to consider is that you can "disable" the default, immutable roles so that they are not accidentally used if I have my full custom setup. 

Other than that, I have one question about the default access a user would get when we have Authn set up as delegated but Authz as the hub.

enhancements/hub-oidc-provider/README.md
jortel: The hub will support two common OIDC federation models.
1. Delegate **authentication-only**: The user identity is verified by the external provider but access-token will likely be ignored (just need the id-token and refresh-token).  The user would then be mapped by claim (subject,email,etc) to a user defined in tackle which would have roles assigned.  These claims/scopes would added (as usual) to the token issued by the hub.
2.  Delegate **both**:  The identity is verified and claims/scopes in the access token issued by the external provider are propagated (or perhaps mapped) into the token issued by the hub.

enhancements/hub-oidc-provider/README.md
```diff
@@ -0,0 +1,299 @@
...
+- Token refresh and revocation
+- UI authentication flows using the Hub OIDC provider
+
+**End-to-End Testing:**
+- Complete user workflows: login, access protected resources, logout
+- Role and permission management through the UI
+- API key creation and usage for programmatic access
+- External provider integration scenarios
+
+### Upgrade / Downgrade Strategy
```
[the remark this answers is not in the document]
fabianvf: I was also going to flag this one - I feel like it's probably something we should figure out - what do people using the existing keycloak implementation do? How do they move over without losing data?

enhancements/hub-oidc-provider/README.md
```diff
@@ -0,0 +1,299 @@
...
+The task manager and addon API will be refactored to use API keys instead of
+custom HMAC-signed JWTs. For backwards compatibility, existing HMAC tokens will
+continue to be honored for in-flight tasks.
+
+API keys support optional expiration dates and can be explicitly revoked through
+the Hub API. The keys themselves are not stored in the database—only
+cryptographic digests—ensuring that compromise of the database does not directly
+expose API keys.
+
+#### Tools Integration
```
fabianvf, line 183: One note, this will break the llm-proxy the way it is currently integrated (jwks validation). We'll need to make sure we have proper api-key orchestration through the operator and update the llm-proxy config so it can continue to authenticate

jortel: Seems like there are 4 use cases for api-key.
1. As an IDE user, I need to obtain an api-key (on behalf of a user) to give to an AI Agent needed to access the LLM Proxy.
2. As an IDE user, I need to obtain an api-key (on behalf of a user) needed to fetch analysis profile bundles.
3. As a Kantra user, I need to obtain an api-key (on behalf of a user) needed to fetch analysis profile bundles.
4. As the LLM Proxy (service), I need to be configured with an api-key needed to access the Hub API. This is orchestrated by the operator.

Perhaps 1&2 can be combined?

Is this correct?

fabianvf: This is more regarding the flow for the LLM proxy, right now it's

get token from hub -> send token to llm-proxy -> llm-proxy validates token with jwks against keycloak 

so I just wanted to note that this will break. I'm not sure exactly what the flow will look like after this - I doubt the llm-proxy will be authing to the hub at any point, since it just talks to the LLMs, so presumably there will be an auth layer sittiing in front of it that handles validating the API keys and passes the traffic through

jortel: Understood.
FWIW, the hub (oidc provider) will have an /oidc/jwks endpoint per RFC 8414. This assumes the token is a JWT (not PAT/api-key.  The api-key approach would require the LLM Proxy to be behind the hub /services endpoint like we do for other micro-services like the KAI Solution Server and disable auth in the LLAMA stack.

I'll update the enhancement.

enhancements/hub-oidc-provider/README.md
```diff
@@ -0,0 +1,416 @@
...
+
+The Hub inventory will be extended with new entities for users, roles,
+permissions, tokens, PATs, and external identity mappings. Users are
+assigned to roles, and roles are associated with permissions (represented as
+OAuth scopes).
+
+The UI will be refactored to use OIDC for authentication (replacing the current
+Keycloak client integration) and will include new pages for managing users,
+roles, and permissions. The login page will be served from a ConfigMap managed
+by the operator, enabling branding customization without code changes.
```
sjd78:
```suggestion
The web UI will be refactored to use OIDC with RFE [Adopt a modern OIDC flow on the UI](https://github.com/konveyor/enhancements/issues/268).  This will handle the login, token refresh, and expired/revoked token handling.

To support the full Hub OIDC provider UX, base user management pages will be implemented and deployed to the operator.  These pages will include the login flow, user self management (i.e. change password), user management, role management, and permission management as needed.  In a federated configuration, this UX will be delegated to the federated OIDC provider.
```

sjd78: Let's call out the web UI enhancement and explicitly call out the OIDC provider UX as being another entity to manage.

enhancements/hub-oidc-provider/README.md
```diff
@@ -0,0 +1,416 @@
...
+The UI deployment will be configured to authenticate against the Hub's OIDC provider instead of
+Keycloak:
+
+- **OIDC client registration**: The operator will configure the UI with OIDC client credentials
+  via ConfigMap or environment variables:
+  - `OIDC_ISSUER`: Points to the Hub's OIDC endpoint (e.g., `https://hub.konveyor.io`)
+  - `OIDC_CLIENT_ID`: Unique identifier for the UI client (e.g., `konveyor-ui`)
+  - `OIDC_CLIENT_SECRET`: Client secret for confidential client authentication (stored in a Secret)
+  - `OIDC_REDIRECT_URI`: Callback URL for the authorization code flow (e.g., `https://ui.konveyor.io/auth/callback`)
+  - `OIDC_SCOPES`: Requested scopes (e.g., `openid profile email`)
```
sjd78: I don't think this is the correct set of information needed by the UI when following the "Authorization Code Flow with Proof Key for Code Exchange (PKCE)" flow that essentially the UI is currently using.  I believe these settings require a backend channel.  We could go this direction and add some oidc helpers to the web ui's express server, but that would be a larger change.

enhancements/hub-oidc-provider/README.md
```diff
@@ -0,0 +1,416 @@
...
+
+For existing deployments currently using Keycloak, the operator will automatically migrate the 
+authentication architecture to use the Hub's built-in OIDC provider with Keycloak configured as an 
+external (federated) IdP.
+
+##### Current Keycloak Configuration (Pre-Migration)
+
+Currently, Keycloak integration is configured via environment variables in the Hub deployment:
+- `AUTH_REQUIRED` - Enable/disable authentication (e.g., `true`)
+- `KEYCLOAK_HOST` - Keycloak server URL (e.g., `https://keycloak.example.com`)
```
sjd78:
```suggestion
- `KEYCLOAK_SERVERL_URL` - Keycloak server URL (e.g., `https://keycloak.example.com`)
```

enhancements/hub-oidc-provider/README.md
```diff
@@ -0,0 +1,416 @@
...
+- `KEYCLOAK_REALM` - Realm name (e.g., `konveyor`)
+- `KEYCLOAK_CLIENT_ID` - Client identifier for the Hub (e.g., `tackle-hub`)
+- `KEYCLOAK_CLIENT_SECRET` - Client secret (stored in Secret)
+- Additional realm-specific settings
```
sjd78, line 353: These 2 do not currently exist.

sjd78 (changes requested):
Reviewing the README -- it will be helpful to separate the web UI updates from the required new OIDC provider pages.  The env vars needed, and how they currently exist need some adjustment as well.

Otherwise, the rest is a good coverage of all the touch points (+OIDC, -Keycloak, +PAT)

enhancements/hub-oidc-provider/README.md
jortel: You are referring specifically to the client-secret. Correct?  We can leave this as PKCE but I think given the web-ui can be a trusted client, sharing the client secret (confidential-client) is faster/simpler than PKCE because it skips the challenge/verifier handshake. PKCE  (public-client) is typically for less trusted clients and will be used for IDE and CLI tools.
That said, I don't feel strongly about this.

enhancements/hub-oidc-provider/README.md
jortel: understood.
This was meant to be more figurative than literal.  We can be more precise on the operator tasks.

enhancements/hub-oidc-provider/README.md
sjd78: There is no way to keep the client secret a secret with a web ui only flow.  That's what PKCE is for.  I can go the route of a normal structure with a trusted client, but that requires the express server to get involved.  All the express server does now is proxy a few endpoints and (mostly) statically serve the webapp resources.

Also, if we do set the client secret and go that way, that value would not get pushed to the web client.

I'm ok leaving this as is and finalizing in implementation.

sjd78 (approved):
LGTM regarding the web UI touch points.

OIDC enabling pages (login) and hub user/role/scope permission pages are indicated and will need tacking issues.

djzager (approved):

PR #267, opened 2026-03-31
:sparkles: agent driven migrations

djzager:
## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive design for Agent-Driven Migration Assistance: autonomous agent orchestration with fallback to prior single-step behavior.
  * Described user-facing permission/autonomy model, streaming UI behaviors, tool-call review flows, and file-change tracking with batch/inline review.
  * Documented local bridge approach, migration intelligence package layout, configuration options, state-management guidance, testing plan, and upgrade/fallback semantics.

### DIFF
--- a/enhancements/kai/agent-driven-migration/README.md
+++ b/enhancements/kai/agent-driven-migration/README.md
@@ -14,8 +14,8 @@ see-also:
   - "https://github.com/konveyor/editor-extensions/issues/1334"
   - "/enhancements/kai/solution_server.md"
   - "/enhancements/distributed-language-extensions/README.md"
-  - "/enhancements/kai/migration-intelligence/README.md"
   - "https://github.com/konveyor/enhancements/pull/259"
+  - "https://agentskills.io"
 replaces: []
 superseded-by: []
 ---
@@ -26,7 +26,10 @@ Evolve the Konveyor editor extension's AI solution pipeline from a single-shot
 LLM workflow to a pluggable agent architecture where autonomous AI agents
 (Goose, OpenCode, or others) can explore the codebase, execute tools, iterate on
 fixes, and produce multi-file migration solutions — all under user-controlled
-permission policies.
+permission policies. Migration intelligence (skills, prompts, and agent
+instructions) is extracted from the extension into a standalone, portable
+package that any agent can consume — in the IDE, from the CLI, or in
+containers.
 
 ## Release Signoff Checklist
 
@@ -37,38 +40,48 @@ permission policies.
 
 ## Open Questions
 
-1. **Credential refresh for Hub LLM proxy**: Agent subprocesses receive Hub
-   bearer tokens at spawn time. These tokens have a ~5 minute TTL. Once expired,
-   the agent cannot make LLM requests and there is no mechanism to deliver
-   refreshed tokens to the running subprocess. This blocks Hub LLM proxy
-   integration with agent mode. See
-   [editor-extensions#1334](https://github.com/konveyor/editor-extensions/issues/1334).
-   Options under consideration:
-   - Agent-side credential refresh (requires upstream Goose changes to OpenAI provider)
-   - Long-lived API keys from Hub
-   - Credential file sidecar pattern (extension writes refreshed tokens to disk,
-     agent re-reads on auth failure)
+1. ~~**Credential refresh for Hub LLM proxy**~~: **Resolved** — addressed by
+   the [credential file sidecar](https://github.com/konveyor/enhancements/pull/265)
+   enhancement. The extension writes refreshed tokens to disk; the agent
+   re-reads on auth failure.
 
-2. ~~**AI artifact distribution**~~: **Resolved** — addressed by the
-   [Migration Intelligence](/enhancements/kai/migration-intelligence/README.md)
-   enhancement. Skills follow the [Agent Skills](https://agentskills.io) format
-   and are distributed via hub profile bundles alongside rulesets.
-
-3. **Backend consolidation**: The POC supports Goose and OpenCode as agent
+2. **Backend consolidation**: The POC supports Goose and OpenCode as agent
    backends. Should the project commit to supporting multiple backends long-term,
    or converge on one? The pluggable architecture allows deferring this decision,
    but supporting multiple backends has ongoing maintenance cost.
 
+3. **Session management**: When does the orchestrator start a new agent session
+   vs. reuse an existing one? What context carries into a session? If session A
+   made file changes, does session B know not to revert them? Is there tracking
+   of context window usage? Can the user stop a running agent and redirect it
+   mid-task?
+
 ## Summary
 
-The Konveyor editor extension currently generates migration solutions by sending
-a single prompt to an LLM and receiving file modifications back. This works for
-straightforward changes but falls short for complex migrations that require the
-AI to explore the codebase, understand dependency relationships, iterate on
-failed attempts, and coordinate changes across multiple files.
+The Konveyor editor extension currently generates migration solutions through
+its `KaiInteractiveWorkflow`, a LangGraph pipeline orchestrated by
+`SolutionWorkflowOrchestrator`. The workflow has two modes controlled by the
+`genai.agentMode` setting:
 
-This enhancement introduces a pluggable agent architecture that replaces the
-single-shot LLM call with autonomous AI agents capable of:
+- **Without agent mode** (`genai.agentMode: false`): The workflow runs Phase 1
+  only — processing incidents per-file using LLM generation without tool access.
+  This is effectively single-shot-per-file: the LLM receives file contents and
+  violation details, generates a fix, and returns. No tool use, no knock-on fix
+  detection, no follow-up phase.
+
+- **With agent mode** (`genai.agentMode: true`): Phase 1 additionally captures
+  knock-on fix information. If identified, the user is prompted to continue.
+  Phase 2 then runs follow-up agents with full tool access (file read/write/search,
+  Maven dependency lookup) and a diagnostics loop for handling cascading changes.
+
+Even with agent mode enabled, the workflow operates on a pre-defined set of
+incidents, has no tool access during initial fix generation (Phase 1), and
+cannot re-analyze after making changes to detect new issues. The underlying
+agent is the `KaiInteractiveWorkflow` — there is no pluggable agent backend
+and no integration with external agent tools like Goose or OpenCode.
+
+This enhancement introduces a pluggable agent architecture that expands
+beyond this workflow to autonomous AI agents capable of:
 
 - **Codebase exploration**: Reading files, searching for patterns, understanding
   project structure before making changes
@@ -78,8 +91,8 @@ single-shot LLM call with autonomous AI agents capable of:
   refining the approach
 - **Multi-file coordination**: Making related changes across many files in a
   single session
-- **User-controlled autonomy**: Granular permission policies that let users
-  decide which agent actions require approval
+- **User-controlled autonomy**: Agent backends manage their own permission
+  models; the IDE surfaces configuration for them
 
 The architecture is backend-agnostic: a common `AgentClient` interface allows
 different agent implementations (Goose via ACP protocol, OpenCode via SDK, or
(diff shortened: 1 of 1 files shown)


djzager:
Tagging related issues:
https://github.com/konveyor/editor-extensions/issues/1243
https://github.com/konveyor/editor-extensions/issues/1334

enhancements/kai/agent-driven-migration/README.md
```diff
@@ -0,0 +1,627 @@
...
+   and are distributed via hub profile bundles alongside rulesets.
+
+3. **Backend consolidation**: The POC supports Goose and OpenCode as agent
+   backends. Should the project commit to supporting multiple backends long-term,
+   or converge on one? The pluggable architecture allows deferring this decision,
+   but supporting multiple backends has ongoing maintenance cost.
+
+## Summary
+
+The Konveyor editor extension currently generates migration solutions by sending
```
pranavgaikwad: This is inaccurate right? We also have an agent mode?

enhancements/kai/agent-driven-migration/README.md
```diff
@@ -0,0 +1,627 @@
...
+environments where autonomous agents are unavailable or undesirable.
+
+## Motivation
+
+### Current State Problems
+
+**Single-shot LLM limitations**: The current `KaiInteractiveWorkflow` sends one
+prompt with violation details and file contents, then receives one response with
+proposed changes. The LLM cannot:
+- Read additional files to understand context
```
pranavgaikwad: same as above. the interactive workflow does have access to file system tools

enhancements/kai/agent-driven-migration/README.md
```diff
@@ -0,0 +1,627 @@
...
+2. **Autonomous codebase exploration**: Agents can read files, search code, and
+   understand project structure before proposing changes, leading to
+   higher-quality migration solutions.
+
+3. **Tool integration via MCP**: Expose Konveyor's analysis capabilities
+   (run analysis, get incidents, apply changes) as MCP tools that agents can
+   invoke. Enable agents to use additional MCP servers for language-specific
+   tasks (Maven search, npm audit, etc.).
+
+4. **Granular permission control**: Users control what the agent can do
```
pranavgaikwad: Different agent backends handle permissions differently. Is that going to be a problem?

enhancements/kai/agent-driven-migration/README.md
```diff
@@ -0,0 +1,627 @@
...
+   autonomously vs. what requires approval. Three autonomy levels (auto, smart,
+   ask) with per-category overrides for file editing, command execution, web
+   access, and MCP tools.
+
+5. **File change review**: All agent-produced file modifications are presented
```
pranavgaikwad, line 172: I think we should investigate how different agents are handling file edits that we can show to the user. I have seen that agents switch between different tools to make file edits depending on context; sometimes they use a sed command; other times they will use a built in file edit tool. Have we thought about feasibility of this?

enhancements/kai/agent-driven-migration/README.md
```diff
@@ -0,0 +1,627 @@
...
+A developer prefers the focused, predictable behavior of single-shot LLM fixes.
+They set `agentMode: false` in settings. When they click "Get Solution," the
+extension sends one prompt to the configured LLM and presents the response as a
+set of file diffs for per-file accept/reject — the same behavior as before the
+agent architecture was introduced.
+
+#### Story 4: Agent uses Konveyor analysis tools
+
+During a migration fix, the agent realizes it needs to check whether its changes
+introduced new violations. It calls the `run_analysis` MCP tool, waits for
```
pranavgaikwad, line 362: I think the analysis MCP server should only ever have 1 exposed tool i.e. `get_analysis_results(file_paths=[], cache=True, limit=-1)`. I do not think that the agent should be calling `run_analysis`. The MCP server should have a file watcher or receive events from the IDE and keep triggering partial analysis every time a file changes. I know it is out of scope for this enhancement, just mentioning here for awareness.

enhancements/kai/agent-driven-migration/README.md
```diff
@@ -0,0 +1,627 @@
...
+    webAccess: "auto" | "ask" | "deny"
+    mcpTools: "auto" | "ask" | "deny"
+    other: "auto" | "ask" | "deny"
+```
+
+The **"smart"** default auto-approves read-only operations and web access, but
+requires approval for file writes and command execution. This balances agent
+productivity with user control.
+
+A `ToolClassifier` maps tool names to categories. For example, Goose reports
```
pranavgaikwad: I think this answers my previous question around tools: am I correct in assuming that we will maintain a mapping of tools per agent backend? If yes, does that affect our goal of allowing *any* agent backends to be used?

enhancements/kai/agent-driven-migration/README.md
```diff
@@ -0,0 +1,627 @@
...
+- OpenCode: Permission block configuration
+
+#### MCP Bridge Architecture (Interim)
+
+The extension exposes Konveyor's analysis capabilities to agents via MCP:
+
+```
+Agent subprocess
+  ↕ (stdio)
+Konveyor MCP Server (Node.js process)
```
pranavgaikwad, line 499: Who will start the konveyor MCP server? Note that it is going to need workspace specific things to start like providers, rules etc.

enhancements/kai/agent-driven-migration/README.md
```diff
@@ -0,0 +1,627 @@
...
+
+#### File Change Tracking
+
+The `AgentFileTracker` detects file modifications made by the agent:
+
+1. **Pre-caching**: Before the agent runs, cache the contents of all files
+   referenced in the incidents being fixed
+2. **Permission-time caching**: When a permission request arrives for a file
+   write tool, cache the file's current state before the write executes
+3. **Post-completion scanning**: After the agent finishes, scan all cached files
```
pranavgaikwad: why do we need this?

enhancements/kai/agent-driven-migration/README.md
djzager: You are correct.

enhancements/kai/agent-driven-migration/README.md
```diff
@@ -0,0 +1,627 @@
...
+   - Long-lived API keys from Hub
+   - Credential file sidecar pattern (extension writes refreshed tokens to disk,
+     agent re-reads on auth failure)
+
+2. ~~**AI artifact distribution**~~: **Resolved** — addressed by the
+   [Migration Intelligence](/enhancements/kai/migration-intelligence/README.md)
+   enhancement. Skills follow the [Agent Skills](https://agentskills.io) format
+   and are distributed via hub profile bundles alongside rulesets.
+
+3. **Backend consolidation**: The POC supports Goose and OpenCode as agent
```
shawn-hurley: Having support for pluggability, but having a default/preferred option is probably the correct thing to do here. Once we implement this, we should start to see which one works the best.

It would be interesting to me to know whether I could also use closed-source options as the agent? Is there a "must conform to this way of calling it?" or something?

enhancements/kai/agent-driven-migration/README.md
```diff
@@ -0,0 +1,627 @@
...
+- **Tool use**: Executing build commands, running tests, querying external
+  services via MCP servers
+- **Iterative problem-solving**: Attempting a fix, checking the result, and
+  refining the approach
+- **Multi-file coordination**: Making related changes across many files in a
+  single session
+- **User-controlled autonomy**: Granular permission policies that let users
+  decide which agent actions require approval
+
+The architecture is backend-agnostic: a common `AgentClient` interface allows
```
shawn-hurley: To confirm: to enable an agent backend, we need an AgentClient implementation we can use. 

Would something be able to register a new implementation that is not shipped with the core project/extension?

shawn-hurley (changes requested):
I don't know if this is a blocker, but I wanted to bring it up.

I feel like there is a potential problem with session management for the backend agent. 

When do we start a new session? What context would be in that session? If session A made changes, would session B know not to revert those changes? 

Is there some tracking of the context window that a session is using?

If a user wants to stop a fix and redirect the agent, is that possible?

enhancements/kai/agent-driven-migration/README.md
```diff
@@ -0,0 +1,627 @@
...
+
+1. **Credential refresh for Hub LLM proxy**: Agent subprocesses receive Hub
+   bearer tokens at spawn time. These tokens have a ~5 minute TTL. Once expired,
+   the agent cannot make LLM requests and there is no mechanism to deliver
+   refreshed tokens to the running subprocess. This blocks Hub LLM proxy
+   integration with agent mode. See
+   [editor-extensions#1334](https://github.com/konveyor/editor-extensions/issues/1334).
+   Options under consideration:
+   - Agent-side credential refresh (requires upstream Goose changes to OpenAI provider)
+   - Long-lived API keys from Hub
```
fabianvf: This is what we will do, and is covered in #265

enhancements/kai/agent-driven-migration/README.md
fabianvf: My guess is that's out of scope though I'd be ecstatic if someone wanted to. I think the backend agnosticism is for our benefit, so that we can integrate with a variety of tools and not get too tied to one, since the space is changing beneath us so quickly. Correct me if I'm wrong @djzager

enhancements/kai/agent-driven-migration/README.md
fabianvf: yeah, these criticisms are valid for the single shot, but a lot of this is at least somewhat supported in the existing agent mode

enhancements/kai/agent-driven-migration/README.md
```diff
@@ -0,0 +1,627 @@
...
+1. **Pluggable agent backends**: Define an `AgentClient` interface that any agent
+   implementation can satisfy. Ship with Goose (ACP protocol) and direct LLM
+   fallback. Architecture supports adding more backends without modifying core
+   orchestration.
+
+2. **Autonomous codebase exploration**: Agents can read files, search code, and
+   understand project structure before proposing changes, leading to
+   higher-quality migration solutions.
+
+3. **Tool integration via MCP**: Expose Konveyor's analysis capabilities
```
fabianvf: You should link this to #259

enhancements/kai/agent-driven-migration/README.md
```diff
@@ -0,0 +1,627 @@
...
+6. **Backward compatibility**: The existing direct LLM workflow remains available
+   as a configuration option (`agentMode: false`). Users who don't want
+   autonomous agents can use the familiar single-shot behavior.
+
+### Non-Goals
+
+1. **Building a new agent framework**: This enhancement integrates existing agent
+   tools (Goose, OpenCode), it does not create a new one.
+
+2. **Replacing the analysis pipeline**: The analyzer, rulesets, and violation
```
fabianvf, line 197: The way the MCP portion is phrased implies that implementation will be a part of this work, this non-goal explicitly says it isn't. We should correct the portion above to make it clear that this work has a dependency on the mcp proposal being implemented

enhancements/kai/agent-driven-migration/README.md
```diff
@@ -0,0 +1,627 @@
...
+│  ┌───────────────────────────────────────────────────────────────┐  │
+│  │ Agent Feature Module                                          │  │
+│  │                                                               │  │
+│  │  AgentOrchestrator ──┬── AgentClient (Goose/OpenCode/Direct)  │  │
+│  │       │              │         │                               │  │
+│  │       │              │    ToolPermissionHandler                │  │
+│  │       │              │         │                               │  │
+│  │       │              └── AgentFileTracker                     │  │
+│  │       │                                                       │  │
+│  │       ├── MCP Bridge Server (localhost HTTP)                   │  │
```
fabianvf, line 232: We have had users report they are unable to open ports in their environments, we will likely need to at least support MCP over stdio (or some other transport) for those users. If the AgentClients don't support this it could be a partial blocker

enhancements/kai/agent-driven-migration/README.md
djzager: There were really 2 ways that I saw this going:
1) The users steps through the process accepting file changes a long the way vis a vis claude
2) enable `isBatchReviewMode ` and we don't ask for approval for those changes as the LLM makes them, then we undo them and show them using the batch review

enhancements/kai/agent-driven-migration/README.md
djzager: My belief is that we need to delegate this.

enhancements/kai/agent-driven-migration/README.md
djzager: Yeah these two particular comments were me failing to properly describe the agentMode enabled flow...I'll fix that in the next version.

enhancements/kai/agent-driven-migration/README.md
djzager: updated as delegated.

enhancements/kai/agent-driven-migration/README.md
djzager: I believe I did a better job of pointing to the mcp server enhancement on this topic.

enhancements/kai/agent-driven-migration/README.md
djzager: some of this feels like more investigation is needed on what logic is necessary to be able to properly render tool use/approvals/reads/writes/etc. We can do what we need to for goose/opencode with a mind towards generic where possible.

enhancements/kai/agent-driven-migration/README.md
djzager: The IDE will own this process managed as part of initializeAgent where we can give the relevant workspace specifics

enhancements/kai/agent-driven-migration/README.md
djzager: I updated the language of the document to reflect that our agentclient implementation is an internal one.

enhancements/kai/agent-driven-migration/README.md
djzager: Fixed, now reference the other enhancement.

enhancements/kai/agent-driven-migration/README.md
djzager: done.

enhancements/kai/agent-driven-migration/README.md
djzager: done

enhancements/kai/agent-driven-migration/README.md
djzager: From claude:
> The extension owns the review pipeline for file changes. apply_file_changes gives agents an explicit MCP-based channel to propose modifications that route through batch review or inline approval — rather than writing to disk first and relying on the AgentFileTracker to catch them after the fact. In the future, this channel also gives us a hook to feed accepted/rejected changes back to the solution server for refining skills and prompts.

From me:
In our POC work we created an interface layer for the purpose of rendering and appropiately catching what the agent was trying to do (for rendering and supporting the review feedback mechanism)....however it seems like it'll be wise for us in the implementation phase to see how much this ~actually holds up~. That is, we want to be generic and if we don't have to do post-completion scanning or any special logic for a particular agent then we should avoid it to the maximum extent possible.

enhancements/kai/agent-driven-migration/README.md
djzager: Doc updated for this.

shawn-hurley (approved):

fabianvf (approved):

pranavgaikwad (approved):

enhancements/kai/agent-driven-migration/README.md
djzager: Updated, we will support any ACP compatible backend.

fabianvf (approved):
LGTM

PR #275, opened 2026-04-03
:sparkles: add ai assisted rules generation proposal

savitharaghunathan:
using skills to generate rules from migration guides

### DIFF
--- a/enhancements/ai-assisted-rules-generation/README.md
+++ b/enhancements/ai-assisted-rules-generation/README.md
@@ -19,9 +19,7 @@ see-also:
 
 # AI-Assisted Rules Generation
 
-Extract migration patterns from documentation via LLM and generate structurally
-valid Konveyor rules, with interactive IDE support (MCP server) and batch/CI
-pipelines (CLI + Agent Skill).
+Extract migration patterns from documentation via LLM and generate structurally valid Konveyor rules, with interactive IDE support (MCP server) and batch/CI pipelines (CLI + Agent Skill).
 
 ## Release Signoff Checklist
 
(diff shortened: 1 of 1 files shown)


enhancements/ai-assisted-rules-generation/README.md
```diff
@@ -0,0 +1,455 @@
...
+- **[analyzer-rule-generator](https://github.com/konveyor-ecosystem/analyzer-rule-generator)** — Closest prior art (Python, migration guides, Claude-oriented flows). We overlap on “guide in, rules out” but diverge on stack: AgentSkills.io + Go `cmd/` helpers, with **`patterns.json` as the stable contract** so deterministic steps stay testable without standardizing on one Python deployment.
+- **[scribe](https://github.com/sshaaf/scribe)** (Java MCP server for rules) — Overlap in “MCP + rules,” but we are not trying to own a general rule CRUD server in this enhancement; MCP remains optional alongside batch and agent workflows.
+- **[semver-analyzer](https://github.com/shawn-hurley/semver-analyzer)** — deterministic API diff in Rust. We can treat it as a **possible future signal** (semver/API changes feeding extraction) rather than something this proposal implements.
+
+### Security, Risks, and Mitigations
+
+- **Prompt injection**: Migration guides from URLs could contain adversarial content. The agent skills include structured extraction formats that constrain LLM output.
```
fabianvf, line 371: we should have a broader discussion on how we'll handle prompt injection as an organization, a lot of our features are vulnerable to it in some way. Could be worth creating a followup issue for

enhancements/ai-assisted-rules-generation/README.md
```diff
@@ -0,0 +1,455 @@
...
+- **URL ingestion**: URLs are fetched with a hardened HTTP client with SSRF mitigation that blocks loopback and private IP addresses.
+- **File system access**: Path traversal is restricted. Workspace directory names are sanitized (no `..`, `/`, `\`).
+- **Supply chain**: Generated rules affect how kantra analyzes applications. Human review before committing to rulesets is essential.
```
fabianvf, line 374: This was the driving idea behind my solution server proposal 👍

enhancements/ai-assisted-rules-generation/README.md
```diff
@@ -0,0 +1,455 @@
...
+
+- Invokes the `scaffold` helper to create the test directory structure and `manifest.json`
+- Reads the manifest to know exactly which files to generate (build file + source file per test group)
+- Generates language-appropriate source code that triggers rule conditions
+- Resolves dependencies if needed
+- Invokes the `sanitize` helper to fix bad XML comments that LLMs may generate
+
+##### `rule-validator`
+
+Fixes failing test data so rules pass kantra validation. Uses a reference-driven approach: reads each failing rule's condition type, looks up the known fix from language-specific provider references (`references/providers/java.md`, `go.md`, `nodejs.md`, `csharp.md`), and applies targeted fixes. Falls back to general LLM reasoning for failure modes not covered by the references. Verifies fixes by running tests via `cmd/test` and iterates up to `max_iterations` (default 1, max 3).
```
fabianvf, line 184: It could also be worth investing in an openapi based structural validator that could run before needing to invoke kantra, to avoid wasting time fixing rules that are structurally invalid rather than incorrect/incomplete

fabianvf (approved):

enhancements/ai-assisted-rules-generation/README.md
savitharaghunathan: +1. totally agree :)

I am also thinking, we can take advantage of analyzer-lsp MCP server's validate_rules tool.

enhancements/ai-assisted-rules-generation/README.md
savitharaghunathan: @coderabbitai create an issue for this please

djzager (approved):
I feel like a lot of the open questions will be answered as we actually use what's being implemented here. The only question that I'm left with is where exactly is this going to live?

savitharaghunathan:
> I feel like a lot of the open questions will be answered as we actually use what's being implemented here. The only question that I'm left with is where exactly is this going to live?

@djzager I have been wondering about how to package and version. Versioning also applies to other skills that we may distribute. This is something I can look into and come back with ideas :) In my head, this is going to be a repo on its own. But, I think it might be more valuable to ship it with migration intelligence package and Analyzer MCP server. what do you think? this could be a step 0 for some projects. Generate rules using this guide -> then use agent driven migration or non-agent workflow (I am guessing we support that). what do you think?

djzager:
I'm wondering if https://github.com/konveyor/rulesets is a good place for this.

1) @shawn-hurley has previously talked about augmenting the rulesets repository with rules so that particular violations include more help for agents to fix them.
2) This is for generating rules so it could be useful to potential rulesets repo contributors to have tooling included in the checkout to help with getting rules in.
3) I am just slow to create a nother repo that may not get eyeballs for changes to be made when they need them.

savitharaghunathan:
> I'm wondering if https://github.com/konveyor/rulesets is a good place for this.
> 
> @shawn-hurley has previously talked about augmenting the rulesets repository with rules so that particular violations include more help for agents to fix them.

I am under the assumption that every provider will ship with its own rulesets.. I might have assumed that this means all java rulesets will be under java provider.. like how c# is right now.

djzager:
I would think, if that is the route that we are going, then we may consider including it in the analyzer-lsp repository directly where those types are most readily available. Come to think of it, that may make the most sense as it is right now. WDYT?

savitharaghunathan:
> I would think, if that is the route that we are going, then we may consider including it in the analyzer-lsp repository directly where those types are most readily available. Come to think of it, that may make the most sense as it is right now. WDYT?

That makes sense to me for now. If needed, I can create a thread for discussion and then come back and update this proposal.

jmle (approved):
Looks good 👍

PR #274, opened 2026-04-03
:sparkles: migration intelligence

djzager:
## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive Migration Intelligence spec covering skill-based migration planning, reusable skills and prompts, portable migration package, phased control flow (skill creation, plan generation, phased execution), IDE/CLI/container interaction patterns, distribution and bundle format changes for profiles, security/risk mitigations, test plans, upgrade/downgrade strategy, compatibility expectations, and implementation checklist with POC history.

PR #279, opened 2026-04-14
Add enhancement: REST frontend and rule candidate staging for Kai Solution Server

fabianvf:
Proposes two additions to the Kai Solution Server:

1. A REST API at /api/v1/ alongside the existing MCP surface, providing documented HTTP endpoints for ingestion, browsing, filtering, and curation of solution-server data. All POST endpoints accept lists.

2. A rule candidate staging and curation data model. Rule candidates (with optional test fixtures and free-form provenance metadata) can be contributed by any producer and triaged through accept/reject workflows. Accepted rules are exported for promotion into the Hub's existing rule management systems (analysis profiles, custom migration targets) or a versioned git repository. The solution server is the staging funnel, not the permanent home for rules.

Also introduces collections as a cross-artifact grouping primitive that generalizes the existing client_id pattern used by MCP.

Key design decisions:
- Producer-agnostic: the API does not privilege any single source of rules or hints
- Provides the APIs needed to build a curation flow, not a curation flow itself
- Rulesets are an assembly/export mechanism; `/archive` is the primary promotion path
- Validation is optional and not a gate for acceptance
- Per-archetype RBAC is required before GA but not the initial implementation
- Review UI is out of scope; the API enables building one
- Git-based triage is discussed as a viable alternative in the Alternatives section

## Summary by CodeRabbit

* **Documentation**
  * Added a specification for an additive REST API at /api/v1/ covering incidents, solutions, violations (with success-rate), hints (including best-hint selection), and cross-artifact collections with association endpoints.
  * Documented rule candidate and ruleset data model and lifecycle (contribute, curate, accept/reject, supersede), pagination/filtering, ruleset assembly, YAML rendering, and exportable ruleset archives.
  * Included signoff checklist, test plan, upgrade/downgrade notes, and security/risk considerations.


abrugaro:
Regarding the rules section, I have a question:

We are currently synchronizing rules from the Hub via analysis profiles, which also allow for custom rules uploaded by the user through custom migration targets. Couldn't either of these two systems cover the need outlined in that enhancement?

I find it a bit confusing to add a new rule "repository" within the solution server itself, using a different database than the hub's and having new endpoints. I also understand that building a UI is out of scope for this enhancement, but if we later wanted to consume these rules from lets say, kantra, we would have to add logic to check if the solution server is available or centralize it through the Hub. The latter would be strange since the Hub already has its own endpoints and rule management.

I understand the proposed flow for rule proposal and acceptance, but I don't follow the part regarding how kantra / the analyzer consumes the served YAML

And regarding testing:

> An end-to-end test that contributes rule candidates, accepts a subset, assembles a ruleset, and confirms the served YAML is consumable without errors

 For the initial version of this enhancement (which does not include a UI), I assume it would be sufficient to include these tests in https://github.com/konveyor/go-konveyor-tests, right? 

And as for the extension tests, the goal would be to verify that changes made by the LLM using the MCP are reflected when queried via REST, and vice versa, is that correct?

fabianvf:
> We are currently synchronizing rules from the Hub via analysis profiles, which also allow for custom rules uploaded by the user through custom migration targets. Couldn't either of these two systems cover the need outlined in that enhancement?
> 
> I find it a bit confusing to add a new rule "repository" within the solution server itself, using a different database than the hub's and having new endpoints. I also understand that building a UI is out of scope for this enhancement, but if we later wanted to consume these rules from lets say, kantra, we would have to add logic to check if the solution server is available or centralize it through the Hub. The latter would be strange since the Hub already has its own endpoints and rule management.

Yeah, this is a fair criticism. I think there's a tension in the proposal that I didn't do a great job resolving. I was trying to keep things minimal and additive (rules are just another data type alongside incidents/solutions/hints), but in doing that I didn't really justify why this needs to live in the solution server instead of going through what the Hub already has.

Honestly, I think you're right that it doesn't make sense as a parallel rule system. Where I actually see this providing value is more as a staging/triage layer. There are a bunch of efforts exploring automated rule generation (commit mining, AI-assisted authoring, semver, etc) and the thing that's missing today is somewhere for those candidates to land before anyone's reviewed them. It doesn't make sense to dump unreviewed candidates directly into the Hub's analysis profiles. What you want is a place where candidates land, where you can filter by validation state, provenance, quality, etc., and then the accepted set gets promoted into the Hub or a git repo. The solution server is the funnel, not the final destination.

The proposal as written doesn't frame it that way, and I think that's what's making it read as unnecessary duplication. I'd like to rework the rules section to make it clearer that the solution server's job here is collection, staging, and triage of candidates, not long-term rule management, and that the expected end state for accepted rules is promotion into the Hub's existing systems (analysis profiles, custom migration targets, or into some repo somewhere). The `/yaml` and `/archive` endpoints would be bridge/export mechanisms for that promotion path.

> I understand the proposed flow for rule proposal and acceptance, but I don't follow the part regarding how kantra / the analyzer consumes the served YAML

That also gets cleaner with the staging framing, if accepted rules get promoted into the Hub, kantra never needs to know the solution server exists. Rules just show up through the channels it already understands. That's probably cleaner than using this as a second API for centralized config to retrieve rules from anyway

And FWIW, I'm not attached to where the service boundary falls here. If it makes more sense for this staging workflow to be part of the Hub's API surface I'd be happy to rewrite it in Go and put it there (@jortel), the data model and lifecycle matter more to me than which binary is serving the endpoints.

> > An end-to-end test that contributes rule candidates, accepts a subset, assembles a ruleset, and confirms the served YAML is consumable without errors
> 
> For the initial version of this enhancement (which does not include a UI), I assume it would be sufficient to include these tests in https://github.com/konveyor/go-konveyor-tests, right?

Yeah, we could do go-konveyor-tests, or just update the existing solution server tests in kai to include it.

> And as for the extension tests, the goal would be to verify that changes made by the LLM using the MCP are reflected when queried via REST, and vice versa, is that correct?

I don't know that it needs additional extension tests as proposed, this is more about laying the foundation for future features than it is about hooking it end to end. I think ideally this provides a mechanism for the solution server to integrate more closely with the hub and new tools than it does anything to change the behavior of the existing integration with the extension.

enhancements/kai/solution_server_rest_and_rules.md
```diff
@@ -0,0 +1,502 @@
...
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] User-facing documentation is created
+
+## Open Questions
+
+1. Per-archetype RBAC for rule approval needs a concrete design before GA. Do reviewers approve rules *for* an archetype, or is rule visibility scoped *by* archetype, or both? The initial implementation ships without per-role enforcement (matching the existing solution-server posture), but this must be resolved before the feature is GA-ready.
+2. The expected end state for accepted rules is promotion into the Hub's existing systems (analysis profiles, custom migration targets) or a versioned git repository. The specific promotion mechanism (manual export via `/archive`, automated sync to a git repo, direct Hub API integration) needs design. How much of this should be automated vs. manual?
+
```
savitharaghunathan, line 37: The solution server accumulates migration data (incidents, solutions, success rates) that could inform improvements to existing rules like refining patterns, adjusting effort ratings, improving messages., or adding an example pair. Should the API surface this signal (e.g., low success-rate rules, common fix patterns) to support existing rule improvement workflows alongside new rule staging?

djzager (approved):
Nice work. At a high level I agree with this enhancement and the direction we are going...trying to make solution server more useful (or at least so people will use it).

I may be aggressive wanting to pursue connections that don't quite exist yet for us to make use of but I think there are cool things we could consider for the solution server.

From my thinking with claude:

> There is an interesting precedent in [codejunkie99/agentic-stack](https://github.com/codejunkie99/agentic-stack) — a portable agent framework with a self-augmentation loop: experiences accumulate, a periodic "dream cycle" extracts candidate lessons, humans review and graduate them, and graduated lessons improve future agent behavior. Skills that fail repeatedly get flagged for rewrite. The solution server could do the organizational-scale version of this — learning from every developer and every agent doing migrations against a Hub, not just one person's experience.

> This also ties into the agent-driven migration work ([PR #274](https://github.com/konveyor/enhancements/pull/274)), where we're extracting migration intelligence (skills, prompts, agent instructions) into portable packages. And we've discussed embedding agent-facing guidance directly into the rulesets interface. If custom rules carry associated hints/skills, then the solution server becomes the learning engine that curates that guidance based on real outcomes: "This hint has a 20% success rate, here's a better one derived from the patterns we've seen in accepted solutions."

I just wonder if we can be more ambitious here. If we can pursue skills/agent specific hints in the rulesets, if we can have skills/prompts/rules/etc exposed in the hub, then we might be able to implement something interesting here with the solution server.

These thoughts are non-blocking so I'm giving it the approval.

enhancements/kai/solution_server_rest_and_rules.md
savitharaghunathan: this can be a follow up. Commented here to start a discussion and see if we can improve existing rules. this is non-blocking imo

savitharaghunathan (approved):

abrugaro (approved):

enhancements/kai/solution_server_rest_and_rules.md
```diff
@@ -0,0 +1,442 @@
...
+- Provide a stable, documented HTTP interface (`/api/v1/` with OpenAPI) for
+  every solution-server use case that is not an interactive LLM tool call.
+- Support browsing and querying incidents, solutions, violations, and hints
+  with pagination and filtering, suitable for UIs and offline analysis.
+- Introduce collections as a grouping primitive so that per-batch,
+  per-review, and per-source scoping work uniformly across incidents,
+  solutions, hints, and (in this enhancement) rules.
+- Provide ingestion endpoints (all `POST` endpoints accept lists)
+  suitable for any non-IDE producer without privileging any single
+  producer in the API shape.
```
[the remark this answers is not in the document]
jortel: Agreed.
Each route should accept/return a specific REST resource.

enhancements/kai/solution_server_rest_and_rules.md
```diff
@@ -0,0 +1,502 @@
...
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] User-facing documentation is created
+
+## Open Questions
+
+1. Per-archetype RBAC for rule approval needs a concrete design before GA. Do reviewers approve rules *for* an archetype, or is rule visibility scoped *by* archetype, or both? The initial implementation ships without per-role enforcement (matching the existing solution-server posture), but this must be resolved before the feature is GA-ready.
```
jortel, line 35: Currently the hub does not support per-resource (id) scopes.  Mainly because it quickly turns into a _can-of-worms_.  Can you elaborate on why this is needed for GA?

jortel (approved):

PR #288, opened 2026-05-05
Add Crane 2.0 enhancement proposals

istein1:
## Summary

- Archive previous crane-2.0 enhancements to `crane-2.0-archived/` to preserve history while making room for the next phase of Crane development
- Add three new enhancement proposals under `crane-2.0/`:
  - **multi-stage-kustomize-transforms**: Replaces the custom JSONPatch pipeline with a Kustomize-native multi-stage transformation pipeline where each plugin operates in an isolated, inspectable stage
  - **migrate-cluster-level-application-workload-dependencies**: Extends the Crane pipeline to discover, export, transform, and apply cluster-scoped resources (CRDs, SCCs, ClusterRoleBindings) referenced by namespace workloads
  - **validate-api-compatibility**: Adds a `crane validate` preflight command that checks rendered manifests against the target cluster's API surface before applying

## Test plan

- [ ] Verify all three README.md files render correctly on GitHub (front matter table, GFM tables, checkboxes, code blocks, links)
- [ ] Confirm archived enhancements are accessible under `crane-2.0-archived/`
- [ ] Review each enhancement for completeness against the [enhancement template](https://github.com/konveyor/enhancements/blob/master/guidelines/enhancement_template.md)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Documentation**
  * Added design for exporting, transforming, and applying cluster-scoped workload dependencies with ordering, opt-out flags, directory/layout conventions, security/compatibility guidance, and test/upgrade plans.
  * Added multi-stage Kustomize transform spec with stage discovery, deterministic outputs, plugin compatibility, CLI behavior changes, and test/rollout guidance.
  * Added read-only validate-api-compatibility preflight design (live and planned offline modes), reporting formats, exit codes, and failure artifact handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/konveyor/enhancements/pull/288?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


enhancements/crane-2.0/multi-stage-kustomize-transforms/README.md
```diff
@@ -0,0 +1,249 @@
...
+  50_CustomModifications/
+    kustomization.yaml
+    resources/
+```
+
+Each stage consumes the `resources/` directory from the previous stage as input. The first stage consumes the export directory.
+
+#### CLI Changes
+
+**`crane transform` — new flags:**
```
rromannissen, line 165: Can you explain what is the expected command and input when the user intends to perform a transformation chaining multiple plugins? I get the impression that the directory structure is being mixed as both input and output.

enhancements/crane-2.0/multi-stage-kustomize-transforms/README.md
```diff
@@ -0,0 +1,249 @@
...
+    resources/
+    patches/
+  50_CustomModifications/
+    kustomization.yaml
+    resources/
+```
+
+Each stage consumes the `resources/` directory from the previous stage as input. The first stage consumes the export directory.
+
+#### CLI Changes
```
rromannissen, line 144: Crane is a new project to the Konveyor ecosystem. It would be convenient to explain what the full list of expected commands are after this enhancement, not only the ones that are new or have been changed.

enhancements/crane-2.0/multi-stage-kustomize-transforms/README.md
```diff
@@ -0,0 +1,249 @@
...
+* **Inspection:** At any time, I can preview the final cleaned YAML before applying it to the cluster by simply running `kubectl kustomize transform/10_KubernetesPlugin/`.
+* **Debugging:** If I need to see exactly what went into the transformation or what came out, I can check the intermediate raw snapshots inside the `transform/.work/10_KubernetesPlugin/input/` and `output/` directories.
+
+#### Story 2: Multi-Stage Transformation with Multiple Existing Plugins
+
+**As a** platform engineer performing a cross-platform migration (such as moving from OpenShift to vanilla Kubernetes), **I want to** run a multi-stage pipeline where each plugin has its own isolated directory (e.g., `10_KubernetesPlugin` and `20_OpenShiftPlugin`), **So that** I can accurately trace patch generation across plugins, inspect intermediate states, and debug unexpected changes without dealing with a single massive patch file.
```
rromannissen, line 99: This feels like design decisions (directory layout) bleeding into use cases.

enhancements/crane-2.0/multi-stage-kustomize-transforms/README.md
```diff
@@ -0,0 +1,249 @@
...
+## Release Signoff Checklist
+
+- [x] Enhancement is `implementable`
+- [x] Design details are appropriately documented from clear requirements
+- [x] Test plan is defined
+- [ ] User-facing documentation is created
+
+## Open Questions
+
+1. **Migration descriptor file vs. CLI flags + conventions (dev preferred):** Should Crane require a declarative descriptor file that specifies which transformers run and in what order, or is the current approach sufficient — CLI commands with flags (`--stage`) combined with the convention-based directory structure (`10_PluginName/`, `20_PluginName/`, `50_CustomChange/`)? The current approach keeps configuration minimal and co-located with the output. A future migration pipeline abstraction may introduce a migration plan file (including these descriptors) naturally, but it is unclear whether the transform descriptor file is needed at the single-tool level.
```
rromannissen, line 33: Can you provide examples of the intended input command for a multi stage transformation? Also related to that, how does the user know about the available plugins? How does the user define the order in which they are executed for a certain transformation?

Regarding the question itself, I think a migration descriptor approach is the best choice, as it would provide a standard file that could be versioned and stored in a repository to handle the repeatable transformation configuration of multiple resources at scale. For example, think about a resource transformation pipeline that has a stage that retrieves the transformation configuration from a repository (our migration descriptor) to then execute the transform command and store the adapted results on a different repository.

enhancements/crane-2.0/multi-stage-kustomize-transforms/README.md
```diff
@@ -0,0 +1,249 @@
...
+* Preflight checks: validates `kubectl` availability and `kustomization.yaml` existence before execution
+* Idempotent by design: `crane apply` delegates to `kubectl apply` with declarative semantics, so repeated runs produce the same result without side effects. Any future additions to the apply flow (pre/post hooks, logging, etc.) must preserve this property.
+
+#### Plugin Compatibility
+
+Existing plugins are unaffected. They continue to return JSONPatch operations via the existing plugin contract. Crane converts these operations into Kustomize Strategic Merge Patches or JSON6902 patch files during the serialization phase. This conversion is internal to `crane-lib`.
+
+#### Stage Discovery
+
+Stages are discovered by convention: Crane scans the transform directory for subdirectories matching `<number>_<pluginName>`. When no explicit configuration is provided, the KubernetesPlugin is assigned priority 10, and remaining plugins are assigned priorities starting at 20 in alphabetical order, with gaps of 10 to allow user-defined custom stages.
```
rromannissen: Does this mean that users have to put together an entire directory structure by themselves in order to configure the transformations to be executed?

enhancements/crane-2.0/multi-stage-kustomize-transforms/README.md
```diff
@@ -0,0 +1,249 @@
...
+| `--stage <dir>` | Transform only one stage |
+| `--force` | Overwrite modified (dirty) stage directories |
+
+**`crane apply` — behavioral changes:**
+
+* Renders output with `kubectl kustomize`
+* Discovers stages automatically from directory structure
+* Supports same stage-selection flags as `transform`
+* Preflight checks: validates `kubectl` availability and `kustomization.yaml` existence before execution
+* Idempotent by design: `crane apply` delegates to `kubectl apply` with declarative semantics, so repeated runs produce the same result without side effects. Any future additions to the apply flow (pre/post hooks, logging, etc.) must preserve this property.
```
rromannissen: Does this mean that `crane apply` executes both `kubectl kustomize` and `kubectl apply`? Why?

enhancements/crane-2.0/validate-api-compatibility/README.md
```diff
@@ -0,0 +1,380 @@
...
+
+An application uses a CRD-backed type `widgets.example.com/v1`. The CRD is not installed on the target cluster. `crane validate` reports the incompatibility. The operator installs the CRD on the target before applying.
+
+#### Story 5: Air-Gapped / Disconnected Environment (Phase 2 -- Offline Mode)
+
+The target cluster is on a restricted network. A team member with cluster access captures the API surface with `kubectl api-resources -o json`. The migration team validates without cluster access using `crane validate --api-resources api-resources.json`. The matching logic, report format, and exit codes are identical to live mode.
+
+#### Story 6: Pre-Provisioning Check (Phase 2 -- Offline Mode)
+
+The target cluster doesn't exist yet, but the team knows it will be OCP 4.14. Someone captures the API surface from a reference cluster and uses it for validation.
```
rromannissen, line 119: I think we should consider the possibility of packaging API resources bundles of well known Kubernetes distributions and versions. In your example, if the user doesn't have access to an OCP 4.14 cluster, but knows that the target is effectively OCP 4.14, it should be possible to just specify that with a flag for simplicity sake.

enhancements/crane-2.0/migrate-cluster-level-application-workload-dependencies/README.md
```diff
@@ -0,0 +1,225 @@
...
+
+Document which cluster-level resource types mig-controller discovers during namespace migration, how it resolves references (e.g., pod spec -> SCC, ServiceAccount -> ClusterRoleBinding), and when elevated permissions are required. This deliverable serves as the reference specification for Crane's behavior.
+
+#### Phase 1 -- Export (#186, closed)
+
+Extend `crane export` to:
+
+* Traverse exported namespace resources and identify references to cluster-scoped objects (SCCs, ClusterRoleBindings, CRDs, RoleAggregations, etc.)
+* Attempt to GET/list referenced cluster-scoped resources
+* On success: write them to the export directory alongside namespace resources
```
rromannissen, line 116: I would suggest to log some INFO or have some differentiation on cluster scoped resources, so it's clear for the user which part of the export corresponds to namespace scoped resources, and which one comes from cluster scoped resources.

enhancements/crane-2.0/migrate-cluster-level-application-workload-dependencies/README.md
```diff
@@ -0,0 +1,225 @@
...
+2. How should `crane apply` behave when cluster-scoped resources already exist on the target with conflicting definitions (merge, skip, fail)?
+3. **Role Aggregation depth:** When a ServiceAccount is bound to an aggregated ClusterRole, should Crane export only the parent ClusterRole, or the full aggregation tree (the parent plus all child ClusterRoles carrying matching `aggregationRule` labels)? Exporting only the parent means the target cluster must already have the child roles installed for aggregation to produce the same effective permissions. Exporting the full tree is more complete but may pull in shared/system ClusterRoles that are not application-specific.
+4. **CRD export vs. CRD-assumed-on-target:** The current implementation exports CRDs referenced by CR instances in the namespace. If the user lacks permissions to GET the CRD, it is not exported and the user is responsible for ensuring the CRD exists on the target. Should the enhancement explicitly state this two-tier behavior (export when possible, assume-on-target when forbidden), or should a CRD that cannot be exported be treated as a hard failure?
```
rromannissen, line 41: I guess the answer to this question depends on whether the discovered CRDs are managed by an operator or not. If the CRDs were managed by an operator in the source cluster, I don't think it would make sense to copy them and create them in the target cluster just to leave them orphan there. I think the enhancement should dive deeper on this scenario.

rromannissen (changes requested):

enhancements/crane-2.0/multi-stage-kustomize-transforms/README.md
aufi: No, this is a typo, thanks for reporting, should be "delegates to `kubectl kustomize`".

enhancements/crane-2.0/multi-stage-kustomize-transforms/README.md
aufi: Sure, multistage commands:

(having exported data from source cluster already)

```
crane transform 
```
^ generates default Kubernetes plugin stage into transform/10_KubernetesPlugin

```
crane transform list-plugins
```
^ outputs available plugins list (default, crane-embeded and also custom from local dirs if available)

```
crane transform --stage 20_OpenShiftPlugin
```
^ creates a new stage with OpenShiftPlugin run after 10_Kubernetes (ordering by lexical order of stage name), with resources from previous stage output

```
crane transform --stage 50_CustomLabelsOrSomething # (no Plugin suffix)
```
^ creates a new stage with data from previous stage and kustomize skeleton files that changes nothing by default, but user can do their custom changes

```
crane apply
```
^ creates output directory with transformed manifests, note: in multistage, each new stage needs applied/materialized output of previous stage, so internally transform runs apply on previous stage to get input resources for a new one

More background on multistage transform with examples is in one of our doc PRs https://github.com/aufi/crane/blob/b53e2565959c54f9f677f2b622c433fce4f0a336/docs/transform-scenarios/03-multistage.md

enhancements/crane-2.0/multi-stage-kustomize-transforms/README.md
aufi: Yes, thanks for suggestion, will update.

enhancements/crane-2.0/multi-stage-kustomize-transforms/README.md
aufi: Correct, each stage contains `resources/` which is its input (that came from either export/ directory for the first stage or from the output of previous stage) and the result of given stage goes to following one (or to output/ directory for final stage when call `apply`), so right, inputs and outputs exist in within the transform directory. 

Reason for that each stage need to work on something real (if there is not yet output of previous stage, it is hard to generate new transformations on unknown data). This bring flexibility and isolation between stages. A limitation it brings is not recommended modification of "previous" stages, that could potentialy break following, already existing ones.

E.g. existing stage 20_... wants patch a resource, but someone does a manual change in already existing previous stage like 10_... that whiteouts such resource, then final `apply` will fail and the broken stage needs to be regenerated again with `transform`, easy for plugin-based stages, or fixed manually), so it make sense suggest users to manually modify just last existing stage (or pay attention to following stage update too).

More on multistage concepts and examples on https://github.com/migtools/crane/pull/307/changes#diff-5e44e00052da9942963c922d4e9f19a89db3d9da4e07d3fca9abec71d76266ce

enhancements/crane-2.0/multi-stage-kustomize-transforms/README.md
aufi: The directory structure is created with `crane transform` commands, user don't have to use commands like mkdir, etc. User can run multiple `transform` commands with for different stages (one line with comand, one stage).

```
crane transform      # uses default Kubernetes Plugin to create 10_KubenetesPlugin, typical (but not required) first step
crane transform --stage 20_SomePlugin
crane transform --stage 50_CustomEdits   
```
Then manual custom changes in stage 50_... and `crane apply` to get the final output.

enhancements/crane-2.0/multi-stage-kustomize-transforms/README.md
aufi: Added on lines below with https://github.com/konveyor/enhancements/pull/288/commits/33ec87c51885a1006597ba4f391dead32bb56ee0

enhancements/crane-2.0/multi-stage-kustomize-transforms/README.md
aufi: On migration descriptor file, thanks for the example, it makes a perfect sense.

My/engineering _soft_ push-back comes from the approach to build crane as a set of small tools doing just their job as good as they can (unix-like style) and under influence of crane-runner project (https://github.com/migtools/crane-runner for pipeline based in-cluster migration orchestration and later potentially backend for UI). From this perspective, `crane transform` and `crane apply` are low-level commands, that (IMO) just get args and do their part of the job.

There is no a technical problem with implementing the descriptor file for transformations (crane transform could get a file with stages/plugins instead of single stage and generate those). The "dirty" part from my point of view is about having this descriptor file and then add (or extend it) to a "bigger" migration descriptor/plan file.

In the end (in later releases), I believe, there should be a single migration descriptor/plan file, that will be consumed by a single endpoint in crane (or related tooling) and it should cover all options needed for the full migration. Having a migration descriptor/migration plan file (that can contain src/dest clusters, transformers/plugins list, etc.) would fit better to an orchestration layer "above" those (`transform` and `apply`) commands.

This file will not be directly visible to `transform` or `apply` subcommands, that are just "cogs" in the migration process.

enhancements/crane-2.0/migrate-cluster-level-application-workload-dependencies/README.md
stillalearner: Cluster resources are stored in different directory , _cluster, under the namespace dir

enhancements/crane-2.0/migrate-cluster-level-application-workload-dependencies/README.md
```diff
@@ -0,0 +1,225 @@
...
+  resources/
+    my-namespace/
+      Deployment_my-namespace_frontend.yaml
+      Service_my-namespace_frontend.yaml
+    _cluster/
+      SecurityContextConstraints_clusterscoped_myapp-scc.yaml
+      ClusterRoleBinding_clusterscoped_myapp-binding.yaml
+      CustomResourceDefinition_clusterscoped_widgets.example.com.yaml
+  failures/
+    clusterscoped/
```
stillalearner: we can remove this from here, 
```suggestion
```

enhancements/crane-2.0/migrate-cluster-level-application-workload-dependencies/README.md
```diff
@@ -0,0 +1,225 @@
...
+Document which cluster-level resource types mig-controller discovers during namespace migration, how it resolves references (e.g., pod spec -> SCC, ServiceAccount -> ClusterRoleBinding), and when elevated permissions are required. This deliverable serves as the reference specification for Crane's behavior.
+
+#### Phase 1 -- Export (#186, closed)
+
+Extend `crane export` to:
+
+* Traverse exported namespace resources and identify references to cluster-scoped objects (SCCs, ClusterRoleBindings, CRDs, RoleAggregations, etc.)
+* Attempt to GET/list referenced cluster-scoped resources
+* On success: write them to the export directory alongside namespace resources
+* On forbidden: log a warning with the specific resource and required permission, and continue the export. The resource is recorded under `failures/clusterscoped/` so administrators can identify what was inaccessible.
```
stillalearner:
```suggestion
* On forbidden: log a warning with the specific resource and required permission, and continue the export. The resource is recorded under `failures/` so administrators can identify what was inaccessible.
```

enhancements/crane-2.0/validate-api-compatibility/README.md
stillalearner: Good idea for a future enhancement. 

For now, anyone with access to a reference cluster of the target version can capture and share the API surface file. 

Bundling well-known distributions is interesting but i think it carries a maintenance burden. Also what all distributions we will be maintaining? Only for OCP?

enhancements/crane-2.0/migrate-cluster-level-application-workload-dependencies/README.md
stillalearner: Yes, thats a valid point. We can add a filter for operator-managed CRDs, and either skip those entirely during export OR just put them in a separate directory operatormanaged/ lets ssay.

For other CRDs, we can proceed with the same logic as today i believe. 

Let me discuss it and make changes here.

aufi: Skip entirely the operator-managed CRDs makes sense to me.

shawn-hurley: CRDs are inherently cluster-scoped and should be the responsibility of the cluster administrator.

If they are not an administrator, then I don't think there is anything they can do. We should warn users, but they will also get an error when applying. 

I think the same thing applies to SCC. 


IMO, there are two cases:

1. Cluster admin is moving from cluster A to cluster B; they will have access, and we do have to apply cluster-scoped resources before namespaced-scoped resources when adding to the target cluster.
2. namespace-scoped admin/user will not have access to CRD/SCC's and will need to coordinate with the appropriate team/user to get the dependencies installed before pushing their app to the target cluster.

Does that make sense to folks? Am I missing a case?

enhancements/crane-2.0/migrate-cluster-level-application-workload-dependencies/README.md
```diff
@@ -0,0 +1,225 @@
...
+4. **CRD export vs. CRD-assumed-on-target:** The current implementation exports CRDs referenced by CR instances in the namespace. If the user lacks permissions to GET the CRD, it is not exported and the user is responsible for ensuring the CRD exists on the target. Should the enhancement explicitly state this two-tier behavior (export when possible, assume-on-target when forbidden), or should a CRD that cannot be exported be treated as a hard failure?
+5. **Cluster-scoped apply opt-in vs. opt-out:** The current implementation exports cluster-scoped resources by default. For apply, should `crane apply` include cluster-scoped resources by default (opt-out via `--namespace-scoped-only`), or exclude them by default (opt-in via `--cluster-scoped-only` or `--include-cluster-scoped`)? The default matters because applying cluster-scoped resources affects the entire cluster and many admins will want to review them separately.
```
shawn-hurley, line 43: I wonder if the default for both should be that we only deal with namespaced scoped resources?

shawn-hurley (commented):
One thing to note,

Say a namespace admin has a service account A. This service account is granted RBAC permissions at the cluster scope to access files in another namespace.

In this situation, it would be namespaced, scoped resources (not cluster-scoped), but you would hit the same problem on apply.


I am going to advocate that we defer all of these problems and assume that when a namespace admin attempts to migrate an application, the cluster administrator has already created the target cluster environment. 

We do need good logging and error messages to explain this, but there is a bigger class of problems that you should be aware of this shape.

enhancements/crane-2.0/migrate-cluster-level-application-workload-dependencies/README.md
dymurray: Or could a user explicitly define which cluster-scoped resources they want to include in a comma separated list?

stillalearner: currently, what we implemented for cluster-scoped resources is, 

export whatever cluster resources are needed (if you have permissions), else just proceed with exporting namespaced scoped resources. 
So that it works in both cases... and extra flag is not needed.

enhancements/crane-2.0/migrate-cluster-level-application-workload-dependencies/README.md
istein1: @stillalearner , ack

enhancements/crane-2.0/migrate-cluster-level-application-workload-dependencies/README.md
istein1: @stillalearner , ack

enhancements/crane-2.0/validate-api-compatibility/README.md
istein1: @stillalearner , This is about OCP to OCP for coming version.

enhancements/crane-2.0/multi-stage-kustomize-transforms/README.md
aufi: For better understading each other & discussion, there is a draft of possible migration descriptor / instructions file https://github.com/aufi/crane/tree/transform-instructions-file/instructions_file_notes#readme

enhancements/crane-2.0/multi-stage-kustomize-transforms/README.md
```diff
@@ -0,0 +1,268 @@
...
+## Implementation History
+
+* **2026-03:** [Issue #207](https://github.com/migtools/crane/issues/207) opened with detailed proposal for Kustomize-native output and multi-stage pipeline.
+* **2026-03:** [Multi-Stage Transform Specification](https://github.com/aufi/move-crane/blob/main/drafts/spec-transform-multistep-kustomize/SUMMARY.md) drafted with full on-disk layout, CLI spec, and implementation breakdown.
+* **2026-04 - present:** Implementation - open for testing, feedback and change requests.
+
+## Drawbacks
+
+* **Breaking change.** Removing `transform-*.json` files will break any user scripts or CI pipelines that parse the old format. This is mitigated by documentation and a deprecation period, but is still disruptive.
+* **kubectl dependency.** Adding a hard runtime requirement on `kubectl` increases the tool's prerequisites. Users in air-gapped environments must ensure `kubectl` is available.
```
dymurray: Is there no way to consume this functionality as a library? I think it would be a mistake to require users to need `kubectl` as a hard requirement (though for most users it's probably not a huge ask)

enhancements/crane-2.0/multi-stage-kustomize-transforms/README.md
```diff
@@ -0,0 +1,268 @@
...
+* **Complexity.** The multi-stage pipeline adds conceptual and implementation complexity. For users running a single plugin, the output still uses the staged directory layout (e.g., `10_KubernetesPlugin/`) for consistency, but only a single stage directory is created.
+
+## Alternatives
+
+1. **Embed a Kustomize library instead of shelling out to kubectl.** Pro: removes the `kubectl` runtime dependency. Con: Kustomize's Go library API is not stable and has historically been difficult to vendor. The `kubectl kustomize` path is simpler and always tracks the user's installed version.
```
dymurray: I would push back on this; we control the version we pull in and kubectl already pulls this in as a dep, so we can just map to the latest version that kube ships.

enhancements/crane-2.0/multi-stage-kustomize-transforms/README.md
dymurray: So if a user has custom plugins and they want `transform` to always run it they use the existing `plugin-manager` to add these so they don't need to chain a bunch of `--stage` commands each time?

enhancements/crane-2.0/migrate-cluster-level-application-workload-dependencies/README.md
stillalearner: @shawn-hurley Yes, these are the 2 cases. 

How the current implementation handles it?
 -> If cluster admin is running the export, all the related cluster resources are exported and stored in a different directory _cluster/   , If the admin wants to do kubectl apply on those cluster resources, he certainly can. If not, just skip applying that directory during "kubectl apply " and handle manually.

-> If namespace admin is running the export, all the related cluster resources are tried to export, but ofcourse not exported finally due to permission issues (which are also logged in failures/) , and the export only exports the namespaced resources. So it becomes the responsibility of admin to handle cluster resources manually.


This way we can keep it simple and handle both scenarios, what do you think about it? 
Do you think there is a need to change this behaviour, or any gap here?

enhancements/crane-2.0/multi-stage-kustomize-transforms/README.md
aufi: It is technically possible import kustomize as go dependency to crane and run kustomize transformations within crane itself. Downsides are large list of added dependencies (possible conflicts primary in client-go and kyaml) and large size of the produced binary.

For me it looks cleaner and less tricky to not embed kustomize to crane, but both options are possible.

enhancements/crane-2.0/multi-stage-kustomize-transforms/README.md
aufi: I'm not sure about overhead with linking correct downstream kubectl/oc dependency according to given release, separated binaries don't have to solve this, but keeping this as a right question to be discussed.

Put together few notes at https://github.com/aufi/move-crane/blob/main/notes/kubectl-call-or-embed.md

enhancements/crane-2.0/multi-stage-kustomize-transforms/README.md
aufi: Existing plugin manager could be used, but this implementation introduced rule _one plugin per stage_ (primary for better debugging), so bunch of `transform --stage` are needed (or the migration description file is needed).

enhancements/crane-2.0/validate-api-compatibility/README.md
```diff
@@ -0,0 +1,380 @@
...
+Validate operates on the **final rendered manifests** from `crane apply`'s output directory, not the raw export or intermediate transform output. This is deliberate:
+
+* **Export** produces raw resources from the source cluster
+* **Transform** runs plugins that may modify apiVersions, add/remove fields
+* **Apply** renders final YAML via `kubectl kustomize` into the output directory
+* **Validate** checks those final manifests against the target cluster's API surface (live or offline)
+* **`kubectl apply`** deploys to the target cluster
+
+Validating after apply (rendering) ensures that transform plugins' changes are reflected in the check. Validating raw exports would produce false positives for resources that transforms will fix.
+
```
nachandr, line 166: Suggestion: For better clarity, could we add this section on resource filtering ? 

**Resource Filtering:**  The runtime-generated resources like Pods, ReplicaSets, Endpoints etc. are  filtered out by `crane transform`. The default KubernetesPlugin (along with other transform plugins if configured) identifies
  and excludes controller-managed resources from the final `kustomization.yaml` , keeping only declarative
  resources that users should apply to the target cluster.

  Here's an example of resource count through the pipeline 
  * **Export stage:** 13 resources (includes Pod, ReplicaSet, Endpoints, EndpointSlice, default ServiceAccount, kube-root-ca.crt ConfigMap)
  * **Transform stage:** Filters to 7 declarative resources (removes the 6 runtime-generated resources above)
  * **Validate stage:** Reports 7 resources scanned

enhancements/crane-2.0/migrate-cluster-level-application-workload-dependencies/README.md
istein1: Resolving this comment as this is already addressed as Sachin mentioned above (see line 151 in this enhancement).

enhancements/crane-2.0/multi-stage-kustomize-transforms/README.md
istein1: After discussing it we agreed on updating the transform with multiple stages generated by a single command and add simple transform descriptor file. Tracked in https://github.com/migtools/crane/issues/370

enhancements/crane-2.0/multi-stage-kustomize-transforms/README.md
istein1: In a discussion we had on this it was agreed to evaluate embedding Kustomize directly into the Crane binary to remove the kubectl dependency on the user's system. This is tracked in https://github.com/migtools/crane/issues/369

enhancements/crane-2.0/validate-api-compatibility/README.md
istein1: In a discussion we had it was agreed to implement "Enable the CLI to fetch relevant metadata automatically based on a provided "OCP version" input" in a later version.

enhancements/crane-2.0/validate-api-compatibility/README.md
stillalearner: yes, we can have something like this

enhancements/crane-2.0/multi-stage-kustomize-transforms/README.md
istein1: Resolving as there is no action required here.

enhancements/crane-2.0/multi-stage-kustomize-transforms/README.md
istein1: In a discussion we had on this it was agreed to evaluate embedding Kustomize directly into the Crane binary to remove the kubectl dependency on the user's system. This is tracked in https://github.com/migtools/crane/issues/369

enhancements/crane-2.0/multi-stage-kustomize-transforms/README.md
istein1: After discussing it we agreed on updating the transform with multiple stages generated by a single command and add simple transform descriptor file. Tracked in https://github.com/migtools/crane/issues/370

enhancements/crane-2.0/multi-stage-kustomize-transforms/README.md
istein1: After discussing it we agreed on updating the transform with multiple stages generated by a single command and add simple transform descriptor file. Tracked in https://github.com/migtools/crane/issues/370

enhancements/crane-2.0/migrate-cluster-level-application-workload-dependencies/README.md
istein1: In a discussion it was decided that we will keep the export as it is right now, following "export whatever we can " rule. 
In addition, Apply command to have an option to skip cluster scoped resources (likely a flag such as --skip-cluster-scoped)

enhancements/crane-2.0/validate-api-compatibility/README.md
```diff
@@ -0,0 +1,390 @@
...
+This enhancement adds a new **`crane validate`** command -- a read-only preflight check that scans the final rendered manifests from `crane apply`'s output directory, queries the **target** cluster's discovery API, and reports which apiVersion+kind combinations are not available on the target under strict group/version matching. When an incompatible resource's kind exists under a different apiVersion on the target, validate suggests the alternative. This brings Crane toward feature parity with mig-controller's GVK compatibility warnings, giving operators early visibility into API skew between clusters.
+
+The enhancement is delivered in phases:
+
+* **Phase 1 (implemented):** Live-cluster validation -- `crane validate` queries the target cluster's discovery API directly via kubeconfig.
+* **Phase 2 (planned):** Offline/disconnected validation -- `crane validate --api-resources api-resources.json` accepts the JSON output of `kubectl api-resources -o json` for air-gapped environments, pre-provisioning checks, and CI/CD pipelines without cluster access.
```
stillalearner:
```suggestion
* **Phase 2 (planned):** Offline/disconnected validation -- ` crane validate --api-resources api-surface.json accepts a captured API surface JSON file (produced by scripts/capture-api-surface.sh)` for air-gapped environments, pre-provisioning checks, and CI/CD pipelines without cluster access.
```

enhancements/crane-2.0/validate-api-compatibility/README.md
```diff
@@ -0,0 +1,390 @@
...
+* **CI-friendly:** Provide deterministic exit codes so `crane validate` can gate a migration pipeline -- non-zero exit when incompatible GVKs are found.
+* **Read-only:** Validate never modifies manifests or cluster state. It is a pure reporting command.
+* **Low privilege:** Validate only requires basic cluster authentication -- it uses the Kubernetes discovery API (`/api`, `/apis`) which is accessible to all authenticated users via the default `system:discovery` ClusterRole. No special RBAC is needed.
+* **Pipeline composability:** Validate fits naturally into the export -> transform -> apply -> validate flow, reading from `crane apply`'s output directory without requiring changes to other commands.
+* **Suggestion engine:** When an incompatible resource's kind is available under a different apiVersion on the target, suggest the alternative to guide the user.
+* **Offline/disconnected support (Phase 2):** Support validation against air-gapped clusters by accepting `kubectl api-resources -o json` output as input, using the same matching logic as live mode.
```
stillalearner:
```suggestion
* **Offline/disconnected support (Phase 2):** Support validation against air-gapped clusters by aaccepting an API surface JSON file as input, using the same matching logic as live mode.
```

enhancements/crane-2.0/validate-api-compatibility/README.md
```diff
@@ -0,0 +1,390 @@
...
+
+After running `crane transform` with plugins that remap deprecated APIs, the operator runs the full pipeline and validates the final output. All entries report OK; exit code 0. The operator proceeds to `kubectl apply` with confidence.
+
+#### Story 4: CRD-Backed Custom Resources
+
+An application uses a CRD-backed type `widgets.example.com/v1`. The CRD is not installed on the target cluster. `crane validate` reports the incompatibility. The operator installs the CRD on the target before applying.
+
+#### Story 5: Air-Gapped / Disconnected Environment (Phase 2 -- Offline Mode)
+
+The target cluster is on a restricted network. A team member with cluster access captures the API surface with `kubectl api-resources -o json`. The migration team validates without cluster access using `crane validate --api-resources api-resources.json`. The matching logic, report format, and exit codes are identical to live mode.
```
stillalearner:
```suggestion
The target cluster is on a restricted network. A team member with cluster access  captures the API surface using scripts/capture-api-surface.sh . The migration team validates without cluster access using `crane validate --api-resources api-surface.json`. The matching logic, report format, and exit codes are identical to live mode.
```

enhancements/crane-2.0/validate-api-compatibility/README.md
```diff
@@ -0,0 +1,390 @@
...
+
+#### Story 6: Pre-Provisioning Check (Phase 2 -- Offline Mode)
+
+The target cluster doesn't exist yet, but the team knows it will be OCP 4.14. Someone captures the API surface from a reference cluster and uses it for validation.
+
+#### Story 7: CI/CD Pipeline Without Cluster Access (Phase 2 -- Offline Mode)
+
+For multi-cluster targeting, the pipeline runs validate once per cluster using the corresponding `api-resources.json` file.
```
stillalearner:
```suggestion
For multi-cluster targeting, the pipeline runs validate once per cluster using the corresponding API surface JSON file.
```

enhancements/crane-2.0/validate-api-compatibility/README.md
```diff
@@ -0,0 +1,390 @@
...
+| `--validate-dir` | Directory where validation report and failure artifacts are saved (default: `"validate"`) |
+| `--kubeconfig` | Path to target cluster kubeconfig |
+| `--context` | Kubeconfig context for target cluster |
+| `--output` / `-o` | Report file format: `json` (default) or `yaml`. Controls the format of the persisted report file. Table output is always printed to the terminal. |
+
+**Additional flag (Phase 2 -- planned):**
+
+| Flag | Description |
+|------|-------------|
+| `--api-resources` | Path to `kubectl api-resources -o json` output file. Enables offline validation. Mutually exclusive with `--kubeconfig`/`--context`. |
```
stillalearner:
```suggestion
| `--api-resources` | Path to API surface JSON file for offline validation. Mutually exclusive with `--kubeconfig`/`--context`. |
```

enhancements/crane-2.0/validate-api-compatibility/README.md
```diff
@@ -0,0 +1,390 @@
...
+|------------|---------|----------|
+| `group` + `version` | `groupVersion` (e.g., `apps/v1`; core resources omit `group`, so groupVersion = `v1`) | GVK matching |
+| `kind` | `kind` | GVK matching |
+| `name` | `resourcePlural` | Report output |
+| `namespaced` | `namespaced` | Future: scope validation |
+
+**Permissions (live mode):** Only requires the Kubernetes discovery API (`/api`, `/apis`), which is accessible to all authenticated users via the default `system:discovery` ClusterRoleBinding -> `system:authenticated` group. No namespace-level RBAC or resource-read permissions are needed.
+
+**Permissions (offline mode):** No cluster access needed from the migration workstation. The person capturing the API resources file needs only basic authentication on the target cluster.
+
```
stillalearner, line 208:
```suggestion

### Capture Script

  The `scripts/capture-api-surface.sh` script captures the full API surface of a target cluster into a single JSON file for offline validation. It queries every served group-version via the raw API endpoints and produces
   the `{"apiResourceLists": [...]}` format that `crane validate --api-resources` expects.

  <details>
  <summary>View script</summary>

  ```bash
  #!/bin/bash
  # capture-api-surface.sh
  # Run this on a machine with kubectl access to the target cluster.
  # It captures the full API surface (all served versions, not just preferred).
  # Pass the output file to: crane validate --api-resources api-surface.json
  #
  # Usage:
  #   bash capture-api-surface.sh --context <context-name> -o <output-file>
  #   bash capture-api-surface.sh --kubeconfig <path> -o <output-file>
  #   bash capture-api-surface.sh --kubeconfig <path> --context <context-name> -o <output-file>
  #
  # Examples:
  #   bash capture-api-surface.sh --context my-target-cluster -o api-surface.json
  #   bash capture-api-surface.sh --kubeconfig /path/to/kubeconfig -o api-surface.json

  set -euo pipefail

  CONTEXT=""
  KUBECONFIG_PATH=""
  OUTPUT=""

  while [[ $# -gt 0 ]]; do
    case "$1" in
      --context)    CONTEXT="$2"; shift 2 ;;
      --kubeconfig) KUBECONFIG_PATH="$2"; shift 2 ;;
      -o)           OUTPUT="$2"; shift 2 ;;
      *)            echo "Unknown flag: $1"; exit 1 ;;
    esac
  done

  if [ -z "$OUTPUT" ]; then
    echo "Usage: bash capture-api-surface.sh [--context <name>] [--kubeconfig <path>] -o <output-file>"
    exit 1
  fi

  # Build kubectl flags
  KUBECTL_FLAGS=""
  [ -n "$CONTEXT" ] && KUBECTL_FLAGS="$KUBECTL_FLAGS --context=$CONTEXT"
  [ -n "$KUBECONFIG_PATH" ] && KUBECTL_FLAGS="$KUBECTL_FLAGS --kubeconfig=$KUBECONFIG_PATH"

  echo "Capturing API surface..."
  [ -n "$CONTEXT" ] && echo "  Context: $CONTEXT"
  [ -n "$KUBECONFIG_PATH" ] && echo "  Kubeconfig: $KUBECONFIG_PATH"

  # Step 1: Get all API versions served by the cluster
  API_VERSIONS=$(kubectl api-versions $KUBECTL_FLAGS)

  # Step 2: For each API version, fetch the list of resources it serves
  echo '{"apiResourceLists":[' > "$OUTPUT"
  FIRST=true

  for GV in $API_VERSIONS; do
    # Core API (v1) lives at /api/v1, everything else at /apis/<group>/<version>
    if [ "$GV" = "v1" ]; then
      ENDPOINT="/api/v1"
    else
      ENDPOINT="/apis/$GV"
    fi

    RESOURCES=$(kubectl get --raw "$ENDPOINT" $KUBECTL_FLAGS 2>/dev/null) || continue

    if [ "$FIRST" = true ]; then
      FIRST=false
    else
      echo "," >> "$OUTPUT"
    fi
    echo "$RESOURCES" >> "$OUTPUT"
  done

  echo ']}' >> "$OUTPUT"

  COUNT=$(echo "$API_VERSIONS" | wc -l | tr -d ' ')
  echo "Done. Captured $COUNT API versions to $OUTPUT"

  Usage:

  # On a machine with target cluster access:
  bash scripts/capture-api-surface.sh --context my-target-cluster -o api-surface.json

  # On the migration workstation (no cluster access needed):
  crane validate --api-resources api-surface.json
  
```

enhancements/crane-2.0/validate-api-compatibility/README.md
```diff
@@ -0,0 +1,390 @@
...
+#### Target Discovery
+
+**Live mode (Phase 1):** Use client-go discovery client:
+
+* Call `ServerGroupsAndResources()` on the target cluster
+* Handle partial discovery failures gracefully (continue with available groups, log warning)
+* Call `discoveryClient.Invalidate()` before querying to ensure fresh results (no stale cache)
+* Build a two-level lookup map: groupVersion -> kind -> APIResource
+
+**Offline mode (Phase 2):** Parse `kubectl api-resources -o json` output:
```
stillalearner:
```suggestion
**Offline mode (Phase 2):**  Parse the API surface JSON file (format: {"apiResourceLists": [<APIResourceList>, ...]} containing native metav1.APIResourceList objects per group-version):
```

enhancements/crane-2.0/validate-api-compatibility/README.md
```diff
@@ -0,0 +1,390 @@
...
+
+* Deserialize the native Kubernetes `APIResourceList` JSON with `json.Unmarshal`
```
stillalearner:
```suggestion
* Deserialize with json.Unmarshal into the same Go types the discovery client return
```

enhancements/crane-2.0/validate-api-compatibility/README.md
```diff
@@ -0,0 +1,390 @@
...
+* Build the same two-level lookup map as live mode
+* Zero custom parsing -- the JSON output uses the same Go types (`metav1.APIResource`) the discovery client returns internally
```
stillalearner:
```suggestion
```

enhancements/crane-2.0/validate-api-compatibility/README.md
```diff
@@ -0,0 +1,390 @@
...
+
+The JSON output from `kubectl api-resources -o json` provides everything needed:
```
stillalearner:
```suggestion
```

enhancements/crane-2.0/validate-api-compatibility/README.md
```diff
@@ -0,0 +1,390 @@
...
+
+| JSON field | Maps to | Used for |
```
stillalearner:
```suggestion
```

enhancements/crane-2.0/validate-api-compatibility/README.md
```diff
@@ -0,0 +1,390 @@
...
+|------------|---------|----------|
+| `group` + `version` | `groupVersion` (e.g., `apps/v1`; core resources omit `group`, so groupVersion = `v1`) | GVK matching |
+| `kind` | `kind` | GVK matching |
+| `name` | `resourcePlural` | Report output |
+| `namespaced` | `namespaced` | Future: scope validation |
```
stillalearner:
```suggestion

```

enhancements/crane-2.0/validate-api-compatibility/README.md
```diff
@@ -0,0 +1,390 @@
...
+| **CRDs installed after capture** | Same as above -- re-run the kubectl command. |
+| **Empty resources array** | Error: "api-resources file contains no resources" |
+| **Malformed JSON** | Error from `json.Unmarshal` with clear message. |
+| **Missing `kind` field at top level** | Warn if `kind` is not `APIResourceList`, but still attempt to parse. |
+| **Core resources (no `group` field)** | `group` is omitted for core API resources. Parser uses `version` alone as `groupVersion` (e.g., `"v1"`). |
+| **Duplicate resources across groups** | Multiple entries with the same kind but different group/version are valid (e.g., `Event` in `v1` and `events.k8s.io/v1`). Index handles this naturally since it's keyed by `groupVersion`. |
+| **Verbs / RBAC** | Out of scope. `verbs` field is present in JSON but ignored. |
+| **Multi-cluster** | `--api-resources` takes one file. For N clusters, run validate N times with the corresponding file. |
+| **kubectl version compatibility** | `kubectl api-resources -o json` is available since K8s 1.11+. The `APIResourceList` schema is stable. |
```
stillalearner:
```suggestion
| **kubectl version compatibility** | `The API surface JSON format wraps native metav1.APIResourceList objects. The capture script uses standard discovery APIs available since K8s 1.11+. |
```

enhancements/crane-2.0/validate-api-compatibility/README.md
```diff
@@ -0,0 +1,390 @@
...
+* **Integration tests:** Capture `kubectl api-resources -o json` from a test cluster, run `crane validate --api-resources` with it -- ensure results match live validate against the same cluster
+* **E2E tests:** Cross-platform scenarios (EKS api-resources validated against OCP manifests)
+* **Regression:** Live-mode validate behavior unchanged
+
+### Offline Mode: Edge Cases (Phase 2)
+
+| Concern | Handling |
+|---------|----------|
+| **Stale data** | Document that the file is a point-in-time snapshot. Users should re-capture after cluster upgrades or CRD changes. |
+| **CRDs installed after capture** | Same as above -- re-run the kubectl command. |
```
stillalearner:
```suggestion
| **CRDs installed after capture** | Same as above -- re-run the capture script. |
```

enhancements/crane-2.0/validate-api-compatibility/README.md
```diff
@@ -0,0 +1,390 @@
...
+| **Empty resources array** | Error: "api-resources file contains no resources" |
+| **Malformed JSON** | Error from `json.Unmarshal` with clear message. |
+| **Missing `kind` field at top level** | Warn if `kind` is not `APIResourceList`, but still attempt to parse. |
+| **Core resources (no `group` field)** | `group` is omitted for core API resources. Parser uses `version` alone as `groupVersion` (e.g., `"v1"`). |
```
stillalearner:
```suggestion
| **Core resources (no `group` field)** | Core API resources use groupVersion: "v1" (no group prefix). The parser handles this naturally since groupVersion is provided directly. |
```

enhancements/crane-2.0/validate-api-compatibility/README.md
```diff
@@ -0,0 +1,390 @@
...
+2. **Validate at export time against both source and target:** Rejected because export is source-only by design, and requiring target access during export adds a dependency that doesn't exist today. Validate as a separate command keeps concerns cleanly separated.
+3. **Transform plugin that rewrites apiVersions:** Complementary, not a replacement. A transform plugin can fix known version mappings, but the user still needs a way to discover *which* mappings are needed. Validate provides that discovery.
+4. **Static compatibility database:** Ship a hardcoded map of "version X removed in Kubernetes Y." Rejected because it cannot cover CRDs, custom API servers, or OpenShift-specific APIs, and requires constant maintenance.
+5. **Custom offline format (Phase 2 alternative):** Define a crane-specific YAML format for target API surface. Rejected in favor of `kubectl api-resources -o json` because it requires zero custom parsing, uses native Kubernetes types, and needs no crane tooling on the target cluster side.
```
stillalearner:
```suggestion
5. **Custom offline format (Phase 2 alternative):** Define a crane-specific YAML format for target API surface. Rejected in favor of the API surface JSON format which wraps native Kubernetes APIResourceList types and only requires a lightweight capture script (scripts/capture-api-surface.sh) on the target cluster side.
```

enhancements/crane-2.0/validate-api-compatibility/README.md
```diff
@@ -0,0 +1,390 @@
...
+
+**No write operations:** Validate is strictly read-only on the cluster. It only writes to the local filesystem (report and failure artifacts under `--validate-dir`).
+
+**Path traversal prevention:** Failure artifact filenames are derived from manifest content (kind, apiVersion, namespace). All components are sanitized through `safeFilePart()` which strips path separators and non-alphanumeric characters, preventing directory escape.
+
+**False positives with CRDs:** If a CRD is installed on the target but its API group hasn't been aggregated yet (e.g., during CRD installation), validate may report a false incompatibility. Documentation should note this timing consideration.
+
+**Stale discovery cache (live mode):** The command calls `discoveryClient.Invalidate()` before querying to ensure fresh results from the target cluster's API server.
+
+**Stale data (offline mode):** The `kubectl api-resources -o json` file is a point-in-time snapshot. Users should re-capture after cluster upgrades or CRD changes. The report includes the source file path and mode for traceability.
```
stillalearner:
```suggestion
**Stale data (offline mode):** The API surface JSON file is a point-in-time snapshot. Users should re-capture after cluster upgrades or CRD changes.
```

enhancements/crane-2.0/validate-api-compatibility/README.md
```diff
@@ -0,0 +1,390 @@
...
+* **#302:** Phase 1 implementation PR -- crane validate command with live-cluster validation, 23 unit tests. *(Merged)*
+* **#319:** Phase 2 feature request -- offline/disconnected cluster support. *(Open)*
+
+## Drawbacks
+
+* **Additional step in the pipeline:** Users must remember to run `crane validate` -- it is not automatically invoked by `crane apply`. This is deliberate (composability), but means users can still skip it.
+* **Strict matching may over-report:** Some resources may work on the target under a different version of the same group (e.g., v1beta1 -> v1), but strict matching flags them as incompatible. The suggestion engine mitigates this by showing available alternatives.
+* **Discovery limitations (live mode):** The Kubernetes discovery API may not surface all available API versions in all configurations (e.g., aggregated API servers with intermittent availability). The command handles partial discovery failures gracefully.
+* **Stale data (offline mode):** The api-resources JSON file is a point-in-time snapshot that may drift from the actual cluster state. Documentation should emphasize re-capturing after changes.
```
stillalearner:
```suggestion
* **Stale data (offline mode):** The API surface JSON file is a point-in-time snapshot that may drift from the actual cluster state. Documentation should emphasize re-capturing after changes.
```

enhancements/crane-2.0/migrate-cluster-level-application-workload-dependencies/README.md
```diff
@@ -32,8 +32,11 @@ see-also:
 ## Open Questions
 
 1. What is the ordering strategy when applying cluster-scoped resources before namespace-scoped ones? CRDs and SCCs must exist before workloads that reference them.
+    * **Decision:** Already answered in the [Apply Ordering](#apply-ordering) design section: 1) CRDs 2) Namespaces 3) ClusterRoles/ClusterRoleBindings/SCCs 4) Namespace-scoped resources.
 2. How should `crane apply` behave when cluster-scoped resources already exist on the target with conflicting definitions — skip or fail? (Merge/reconcile is excluded per Non-Goals.)
+    * **Decision:** Fail by default and report the conflict. Since `crane apply` delegates to `kubectl apply` which uses declarative semantics (update-if-exists), the practical behavior is that existing resources get updated. Add a `--dry-run` recommendation in documentation so users can preview before applying.
```
stillalearner: crane apply will have an option to skip cluster scoped resources in output/ dir 

--skip-cluster-scoped-resources

enhancements/crane-2.0/migrate-cluster-level-application-workload-dependencies/README.md
istein1: Updated Open Question 2 to reflect the current implementation: `crane apply` won't fail on conflicting cluster-scoped resources, but `kubectl apply` will — blocking remaining namespaced resources. Added Context section explaining this behavior and updated the Decision to introduce `--skip-cluster-scoped` flag so users can bypass cluster-scoped resources and apply non-conflicting namespaced resources successfully.

enhancements/crane-2.0/migrate-cluster-level-application-workload-dependencies/README.md
istein1: Thanks @stillalearner, 
Fixed in commit [4c0f7e2](https://github.com/konveyor/enhancements/pull/288/commits/4c0f7e2915005e0c220028750bbeb234e8268aed)

enhancements/crane-2.0/validate-api-compatibility/README.md
istein1: Created https://github.com/migtools/crane/issues/387 to track this feature for a later version.

enhancements/crane-2.0/multi-stage-kustomize-transforms/README.md
```diff
@@ -0,0 +1,278 @@
...
+4. Is it right to assume:
+    1. that user-created scripts around the previous `transform-*.json` files are not supported anymore? (not affecting plugins)
+        * **Decision:** Yes — `transform-*.json` files are replaced by the Kustomize overlay layout. Existing user scripts that parse `transform-*.json` will break. Document the migration path and consider a one-release deprecation period (already noted in Breaking Change section).
+    2. that `crane apply` can call `kubectl kustomize` internally (or `oc`)? The binary itself will not have kustomize evaluation logic embedded.
+        * **Update:** Embedding Kustomize directly into the Crane binary is being evaluated to remove the `kubectl` runtime dependency ([crane#369](https://github.com/migtools/crane/issues/369)). The current implementation shells out to `kubectl kustomize`; this may change.
+        * **Decision:** Currently yes — `crane apply` shells out to `kubectl kustomize`. Embedding Kustomize as a Go library is being evaluated ([crane#369](https://github.com/migtools/crane/issues/369)) to remove the `kubectl` runtime dependency. The answer may change based on that evaluation.
```
stillalearner:
```suggestion
        * **Decision:** Resolved — Kustomize is now embedded directly into the Crane binary as a Go library (`sigs.k8s.io/kustomize/api/krusty`). There is no runtime dependency on `kubectl` or `oc` for transform or apply operations ([crane#369](https://github.com/migtools/crane/issues/369), [crane#388](https://github.com/migtools/crane/pull/388)).
```

enhancements/crane-2.0/multi-stage-kustomize-transforms/README.md
```diff
@@ -0,0 +1,278 @@
...
+* **No traceability for multiple plugins.** When multiple plugins run, their patches are merged into a single file per resource. If a transformation produces unexpected results, there is no way to determine which plugin was responsible without re-running plugins individually.
+
+### Why Now
+
+The earlier [kustomize-support](https://github.com/konveyor/enhancements/tree/master/enhancements/crane-2.0/kustomize-support) enhancement was rejected because it offered "no more than `kustomize init --autodetect`." Since then, the plugin system has matured, the need for traceability has intensified (users run 3-5 plugins in production pipelines), and a detailed specification for multi-stage execution has been drafted. This proposal addresses the deeper integration that the earlier enhancement deferred.
+
+### Goals
+
+1. **Kustomize-native output.** `crane transform` produces a valid Kustomize overlay (`kustomization.yaml` + patch files) instead of flat JSONPatch files. Users can run `kubectl kustomize <transform-dir>` to preview the final manifests at any time.
```
stillalearner:
```suggestion
1. **Kustomize-native output.** `crane transform` produces a valid Kustomize overlay (`kustomization.yaml` + patch files) instead of flat JSONPatch files. Crane renders these internally via the embedded kustomize library. Users can also run `kubectl kustomize <transform-dir>` externally to preview the final manifests.
```

enhancements/crane-2.0/multi-stage-kustomize-transforms/README.md
```diff
@@ -0,0 +1,278 @@
...
+3. **Debug-friendly stages.** Each stage (plugin and custom) produces its own input and output manifests under directory `transform/.work/10_ThePlugin`.
+4. **Deterministic, Git-friendly output.** Patch files use stable naming (`<namespace>--<group>-<version>--<Kind>--<name>.patch.yaml`) and deterministic ordering so that repeated runs produce identical output when input is unchanged.
+5. **Zero plugin changes.** Existing plugins continue to return JSONPatch operations. Crane handles the conversion to Kustomize patches transparently.
+6. **Standard toolchain integration.** The output works with `kubectl`, `kustomize`, ArgoCD, Flux, OPA/Gatekeeper, and other tools that consume Kustomize overlays.
+
+### Non-Goals
+
+* **Replacing Kustomize.** Crane is not building its own overlay engine. It produces standard Kustomize artifacts and delegates rendering to `kubectl kustomize`.
```
stillalearner:
```suggestion
* **Replacing Kustomize.** Crane is not building its own overlay engine. It produces standard Kustomize artifacts and renders them via the embedded kustomize library.
```

enhancements/crane-2.0/multi-stage-kustomize-transforms/README.md
```diff
@@ -0,0 +1,278 @@
...
+* **Plugin distribution.** How plugins are discovered, installed, and updated is covered by the existing crane [plugin-management](https://github.com/konveyor/enhancements/tree/master/enhancements/crane-2.0/plugin-management) system.
+* **Operator/controller integration.** This enhancement covers the CLI workflow only. Integration with the Crane operator or mig-controller is out of scope.
+* **Semantic validation between stages.** Cross-stage content validation is delegated to `kubectl kustomize` at apply time.
```
stillalearner:
```suggestion
* **Semantic validation between stages.** Cross-stage content validation is delegated to the embedded kustomize at apply time.
```

enhancements/crane-2.0/multi-stage-kustomize-transforms/README.md
```diff
@@ -0,0 +1,278 @@
...
+## Proposal
+
+### User Stories
+
+#### Story 1: Transformation with the Default Plugin
+
+**As a** platform engineer migrating an application, **I want to** run `crane transform` to execute the default `10_KubernetesPlugin`, **So that** server-managed metadata is removed, generating standard Kustomize patches that I can easily inspect and debug.
+
+* **Patch Generation:** The plugin automatically generates JSONPatch operations that Crane transparently converts into Kustomize patches, saving them in `transform/10_KubernetesPlugin/patches/` using a predictable, Git-friendly naming convention (`<namespace>--<group>-<version>--<Kind>--<name>.patch.yaml`).
+* **Inspection:** At any time, I can preview the final cleaned YAML before applying it to the cluster by simply running `kubectl kustomize transform/10_KubernetesPlugin/`.
```
stillalearner:
```suggestion
* **Inspection:** At any time, I can preview the final cleaned YAML before applying it to the cluster. Crane renders this internally, or I can run `kubectl kustomize transform/10_KubernetesPlugin/` externally for manual inspection.
```

enhancements/crane-2.0/multi-stage-kustomize-transforms/README.md
```diff
@@ -0,0 +1,278 @@
...
+* **Debugging:** If I need to see exactly what went into the transformation or what came out, I can check the intermediate raw snapshots inside the `transform/.work/10_KubernetesPlugin/input/` and `output/` directories.
+
+#### Story 2: Multi-Stage Transformation with Multiple Existing Plugins
+
+**As a** platform engineer performing a cross-platform migration (such as moving from OpenShift to vanilla Kubernetes), **I want to** run a multi-stage pipeline where each plugin has its own isolated directory (e.g., `10_KubernetesPlugin` and `20_OpenShiftPlugin`), **So that** I can accurately trace patch generation across plugins, inspect intermediate states, and debug unexpected changes without dealing with a single massive patch file.
+
+* **Patch Generation:** Each plugin's output is isolated into its own stage directory with its own `kustomization.yaml` and `patches/` folder, meaning patches from different plugins are never merged into an opaque file.
+* **Inspection:** I can inspect the output of any specific stage individually using standard tools like `kubectl kustomize <stage-dir>`.
```
stillalearner:
```suggestion
* **Inspection:** I can inspect the output of any specific stage individually. Crane renders each stage internally, or I can use `kubectl kustomize <stage-dir>` externally for manual verification.
```

enhancements/crane-2.0/multi-stage-kustomize-transforms/README.md
```diff
@@ -0,0 +1,278 @@
...
+
+#### Story 3: Custom Transformation Stage with User-Created Modifications
+
+**As a** platform engineer adapting migrated resources to new environment standards, **I want to** append a custom transformation stage (e.g., `50_CustomModifications`) to the pipeline, **So that** I can layer GitOps-friendly declarative patches and overrides (like swapping image tags or updating namespaces) directly on top of the automated Crane output.
+
+* **Modification Generation:** I can manually change the `kustomization.yaml` inside my custom stage directory and define custom modifications, like `namespace: migrated-app` or an `images:` block to swap container image tags, or changing annotation labels, or write custom patches without needing to write or compile a Go-based Crane plugin.
+* **Inspection:** I can verify my custom changes alongside the automated ones by running `kubectl kustomize` on my custom stage directory, ensuring my edits correctly apply over the output of the previous stage.
```
stillalearner:
```suggestion
* **Inspection:** I can verify my custom changes alongside the automated ones by running `crane apply` or `kubectl kustomize` on my custom stage directory, ensuring my edits correctly apply over the output of the previous stage.
```

enhancements/crane-2.0/multi-stage-kustomize-transforms/README.md
```diff
@@ -0,0 +1,278 @@
...
+**`crane transform` — new flags:**
+
+| Flag | Purpose |
+|------|---------|
+| `--stage <dir>` | Transform only one stage |
+| `--force` | Overwrite modified (dirty) stage directories |
+
+**`crane apply` — behavioral changes:**
+
+* Renders output with `kubectl kustomize`
```
stillalearner:
```suggestion
* Renders output via embedded kustomize library (no external `kubectl` or `oc` dependency)```

File: enhancements/crane-2.0/multi-stage-kustomize-transforms/README.md [hunk 43]
```diff
@@ -0,0 +1,278 @@
...
+* Discovers stages automatically from directory structure
+* Supports same stage-selection flags as `transform`
+* Preflight checks: validates `kubectl` availability and `kustomization.yaml` existence before execution
+* Deterministic by design: `crane apply` delegates manifest rendering to `kubectl kustomize`, so repeated runs with unchanged inputs produce identical rendered output. Any future additions to the apply flow (pre/post hooks, logging, etc.) must preserve this property.
```
stillalearner:
```suggestion
* Preflight checks: validates `kustomization.yaml` existence before execution
* Deterministic by design: `crane apply` renders manifests via the embedded kustomize library, so repeated runs with unchanged inputs produce identical rendered output. Any future additions to the apply flow (pre/post hooks, logging, etc.) must preserve this property.
```

enhancements/crane-2.0/multi-stage-kustomize-transforms/README.md
```diff
@@ -0,0 +1,278 @@
...
+For stable Git diffs and reproducible CI runs:
+
+* Resources within type files preserve creation order
+* `resources` and `patches` lists in `kustomization.yaml` are sorted lexicographically by filename
+* Patch filenames follow the pattern `<namespace>--<group>-<version>--<Kind>--<name>.patch.yaml`
+* Stage directories follow `<priority>_<pluginName>`
+
+### Security, Risks, and Mitigations
+
+**kubectl dependency.** `crane apply` currently delegates to `kubectl kustomize` at runtime. Embedding Kustomize as a Go library is being evaluated ([crane#369](https://github.com/migtools/crane/issues/369)) to remove this external dependency. Until then, a preflight check verifies `kubectl` availability with a clear error message.
```
stillalearner:
```suggestion
**Kustomize embedding.** Kustomize is embedded directly into the Crane binary as a Go library (`sigs.k8s.io/kustomize/api/krusty`). There is no runtime dependency on `kubectl` or `oc` for transform or apply operations ([crane#369](https://github.com/migtools/crane/issues/369), [crane#388](https://github.com/migtools/crane/pull/388)).
```

enhancements/crane-2.0/multi-stage-kustomize-transforms/README.md
```diff
@@ -0,0 +1,278 @@
...
+## Design Details
+
+### Implementation Epics
+
+| Epic | Scope | Key Deliverables |
+|------|-------|------------------|
+| **A — crane-lib: Kustomize Foundations** | Library | `TransformArtifact`/`PatchTarget` types, Kustomize serializer, resource grouping, report structs |
+| **B — crane: Transform Refactor** | CLI | Kustomize overlay generation, resource type files, deterministic ordering, dirty check |
+| **C — Multi-Stage Pipeline** | CLI + Lib | Stage discovery, per-stage execution, stage-aware CLI flags, priority auto-assignment |
+| **D — crane: Apply Refactor** | CLI | `kubectl kustomize` delegation, preflight checks, output handling |
```
stillalearner:
```suggestion
| **D — crane: Apply Refactor** | CLI | Embedded kustomize rendering, output handling |
```

enhancements/crane-2.0/multi-stage-kustomize-transforms/README.md
```diff
@@ -0,0 +1,278 @@
...
+## Implementation History
+
+* **2026-03:** [Issue #207](https://github.com/migtools/crane/issues/207) opened with detailed proposal for Kustomize-native output and multi-stage pipeline.
+* **2026-03:** [Multi-Stage Transform Specification](https://github.com/aufi/move-crane/blob/main/drafts/spec-transform-multistep-kustomize/SUMMARY.md) drafted with full on-disk layout, CLI spec, and implementation breakdown.
+* **2026-04 - present:** Implementation - open for testing, feedback and change requests.
+
+## Drawbacks
+
+* **Breaking change.** Removing `transform-*.json` files will break any user scripts or CI pipelines that parse the old format. This is mitigated by documentation and a deprecation period, but is still disruptive.
+* **kubectl dependency (under evaluation).** The current implementation requires `kubectl` at runtime. Embedding Kustomize as a library is being evaluated ([crane#369](https://github.com/migtools/crane/issues/369)) to eliminate this prerequisite.
```
stillalearner:
```suggestion
* **~~kubectl dependency.~~** Resolved — Kustomize is now embedded. No `kubectl` runtime dependency for transform or apply ([crane#369](https://github.com/migtools/crane/issues/369), [crane#388](https://github.com/migtools/crane/pull/388)).
```

enhancements/crane-2.0/multi-stage-kustomize-transforms/README.md
```diff
@@ -0,0 +1,278 @@
...
+* **Complexity.** The multi-stage pipeline adds conceptual and implementation complexity. For users running a single plugin, the output still uses the staged directory layout (e.g., `10_KubernetesPlugin/`) for consistency, but only a single stage directory is created.
+
+## Alternatives
+
+1. **Embed a Kustomize library instead of shelling out to kubectl.** Pro: removes the `kubectl` runtime dependency. Con: Kustomize's Go library API is not stable and has historically been difficult to vendor. **This option is being actively evaluated** ([crane#369](https://github.com/migtools/crane/issues/369)).
```
stillalearner:
```suggestion
1. **~~Embed a Kustomize library instead of shelling out to kubectl.~~** **Adopted.** Kustomize is now embedded via `sigs.k8s.io/kustomize/api/krusty` ([crane#369](https://github.com/migtools/crane/issues/369), [crane#388](https://github.com/migtools/crane/pull/388)). The `kustomize/api` module was already a transitive dependency, so binary size did not increase.
```

enhancements/crane-2.0/multi-stage-kustomize-transforms/README.md
```diff
@@ -0,0 +1,278 @@
...
+2. **Keep JSONPatch files but add a `crane render` command.** Pro: no breaking change. Con: still requires users to learn Crane-specific tooling; does not integrate with the broader Kustomize/GitOps ecosystem.
+3. **Use Helm charts instead of Kustomize overlays.** Pro: Helm provides great flexibility with templating system and it is widely adopted. Con: Typical k8s workload migration process aims to clean resources with minimal needed changes to make possible to be imported to the target cluster, templating should not be required in most cases. Helm's templating model is fundamentally different from patch-based transformation.
+
+## Infrastructure Needed
+
+* No new repositories required. Changes land in [migtools/crane](https://github.com/migtools/crane) and [konveyor/crane-lib](https://github.com/konveyor/crane-lib).
+* CI must have `kubectl` available for integration tests (already the case in the existing test infrastructure).
```
stillalearner:
```suggestion
* CI must have `kubectl` available for E2E tests that deploy to clusters (already the case in the existing test infrastructure). Transform and apply unit tests do not require `kubectl`.
```

rromannissen (approved):
LGTM

PR #290, opened 2026-05-15
:book: Enhancement for bundling rules directly in each provider

### LINKED ISSUE #289: [RFE] Segregate rules by provider
Currently we pull all rules to each provider. This causes issues, like the c-sharp provider running java rules with "run always" selectors.

We need to create a proposal for an enhancement for each provider to package its own rules and nothing else. Some points of concern are:
- This should work for IDE, hub and CLI.
- How to distribute the rules' code: should they be scattered on each provider's repo, or should they still be kept in the rulesets/seed repo?
- How to distribute the rules at runtime.
- We should make it as easy as possible for contributors.
- Many rules have `builtin` provider conditions _and_ their own provider's. builtin conditions should still work.
- For the moment we can drop the possibility of having rules with multiple providers other than builtin.`

jmle:
This is a proposal for bundling rules along their respective providers.

Main points:
- From my point of view, the rulesets repo should still be the origin of the rules, and the seeding process is still legitimate. This wouldn't change.
- There is the question of how we distribute rules. Should rules be bundled along each provider inside the image? This would be good for kantra with containers, but weird for the hub, which keeps its own database of the rules.
  - The solution proposed for this is that we do bundle the rules inside each provider, but first we check on a given mounted path for existing rules. Those would take priority, allowing the hub to keep doing its thing.
  - For kantra, each provider would take its rules from inside the image, except for containerless java, which would have its rules in the filesystem.
  
Fixes https://github.com/konveyor/enhancements/issues/289

Claude-assisted

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive design/proposal README for provider-specific rule packaging and runtime loading. It defines provider-to-ruleset mapping, distribution architecture, container behaviors (bundled vs extracted rules), repository/build/CICD requirements, seeding/packaging/versioning guidance, upgrade/downgrade fallback plan, testing matrix (unit/integration/e2e/regression), security risk mitigations, expected repository layout, and noted trade-offs/rejected alternatives.

[Review Change Stack](https://app.coderabbit.ai/change-stack/konveyor/enhancements/pull/290?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/enhancements/provider-specific-rule-packaging/README.md
+++ b/enhancements/provider-specific-rule-packaging/README.md
@@ -97,6 +97,10 @@ additional configuration required on my part.
 As a user running kantra for Java analysis, I want the bundled Java provider to include all Java rules without requiring
 external network access or additional downloads.
 
+#### Story 6: Custom Rules User
+As a user with organization-specific migration rules, I want to provide my custom rules in addition to the default
+bundled rules so that my analysis includes both standard and custom checks.
+
 ### Implementation Details/Notes/Constraints
 
 #### Current Rule Distribution Architecture
@@ -164,26 +168,35 @@ tackle2-seed/resources/rulesets/
 │                                       │
 │ Spawns provider containers            │
 │ (rules bundled in provider images)    │
+│                                       │
+│ Optional: mount custom rules          │
+│  -v /custom/rules:/opt/custom-rulesets:ro
 └───────────────────────────────────────┘
 
 ┌─────────────────────────────────────┐
 │ Kantra Distribution (ZIP)           │
 │ rulesets/java/ (for containerless)  │
+│                                     │
+│ Optional: custom rules on host      │
+│ mounted to provider containers      │
 └─────────────────────────────────────┘
 
 ┌─────────────────────────────────────┐
 │ Java Provider Container             │
-│ /opt/rulesets/ (Java only)          │
+│ /opt/rulesets/ (Java only - bundled)│
+│ /opt/custom-rulesets/ (optional)    │
 └─────────────────────────────────────┘
 
 ┌─────────────────────────────────────┐
 │ Node.js Provider Container          │
-│ /opt/rulesets/ (Node.js only)       │
+│ /opt/rulesets/ (Node.js - bundled)  │
+│ /opt/custom-rulesets/ (optional)    │
 └─────────────────────────────────────┘
 
 ┌─────────────────────────────────────┐
 │ .NET Provider Container             │
-│ /opt/rulesets/ (dotnet only)        │
+│ /opt/rulesets/ (dotnet - bundled)   │
+│ /opt/custom-rulesets/ (optional)    │
 └─────────────────────────────────────┘
 ```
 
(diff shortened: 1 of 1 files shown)


enhancements/provider-specific-rule-packaging/README.md
```diff
@@ -0,0 +1,550 @@
...
+│ Reads: /opt/rules (if mounted)       │
+│    OR: /opt/bundled-rules (fallback) │
+└──────────────────────────────────────┘
+```
+
+#### Rule Source of Truth
+
+The `rulesets` repository will remain the authoritative source for all rule definitions. This provides:
+- Centralized rule management and versioning
+- Single location for rule contributions and reviews
```
eemcmullan, line 208: +1

enhancements/provider-specific-rule-packaging/README.md
```diff
@@ -0,0 +1,550 @@
...
+only relevant rules to each provider.
+
+### Provider-to-Ruleset Mapping
+
+The following mapping will be established:
+
+| Provider | Ruleset Directory | Notes |
+|----------|------------------|-------|
+| Java | `rulesets/stable/java/` | Most mature ruleset |
+| .NET (C#) | Currently in provider repo | Future: migrate to `rulesets/stable/dotnet/` |
```
eemcmullan, line 68: I think we should use `csharp` rather than `dotnet` to align with the provider name

enhancements/provider-specific-rule-packaging/README.md
```diff
@@ -0,0 +1,550 @@
...
+#### Required Changes to Current Architecture
+
+To achieve provider-specific rule packaging, the following changes are needed:
+
+**Provider Container Changes (Core Change):**
+- Currently: Provider containers do NOT bundle any rules
+- Proposed: Each provider bundles its language-specific seeded rules from tackle2-seed
+  - Dockerfile copies relevant ruleset directory: e.g., `COPY --from=rulesets /tackle2-seed/resources/rulesets/java/ /opt/bundled-rules/`
+- **Rule Loading Logic**: Providers support both bundled and external rules with simple precedence:
+  1. Check for externally-provided rules (e.g., `/opt/rules/` via volume mount) - use if present
```
eemcmullan: Would this only be custom rules? If so, the bundled rules wouldn't be a fall back?

enhancements/provider-specific-rule-packaging/README.md
jmle: We have inconsistencies with this matter, I think @dymurray wanted to unify it to dotnet?

enhancements/provider-specific-rule-packaging/README.md
jmle: This was originally for the hub to be able to provide the rules from the outside, I wasn't thinking about custom rules, but it's a good point, we need a mechanism for loading custom rules and the enhancement doesn't mention custom rules anywhere. I will update

jmle: Added info for custom rules

enhancements/provider-specific-rule-packaging/README.md
```diff
@@ -0,0 +1,612 @@
...
+migration scenarios, proprietary framework migrations, or other specialized analysis needs.
+
+**Custom Rules Mechanism:**
+- Providers load rules from two locations:
+  1. `/opt/rulesets/` - Default bundled rules (always loaded from the container image)
+  2. `/opt/custom-rulesets/` - Custom user-provided rules (optional, via volume mount)
+- Custom rules are loaded **in addition to** default rules, not as a replacement
+- Custom rules can be unstructured (do not need to follow the language-specific directory organization of default rules)
+- Override behavior: The existing rule loading logic for handling duplicate rule IDs is preserved
+
```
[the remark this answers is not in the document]
jmle: I don't think we need to address this directly, as the current logic for precedence should suffice.

eemcmullan (approved):

djzager (approved):
ACK. went ahead and created the tracking issue in ide repo. Thank you for doing this.

PR #291, opened 2026-05-26
Update multi-stage transform enhancement for embedded kustomize

stillalearner:
Reflects that crane#369 is now completed: kustomize is embedded directly into the crane binary via sigs.k8s.io/kustomize/api/krusty. No runtime dependency on kubectl or oc for transform/apply operations.

- Updated Q&A, goals, non-goals, user stories, apply behavior, risks, epics table, drawbacks, and alternatives sections
- "Being evaluated" and "shells out to kubectl" language replaced with "embedded kustomize library" throughout
- kubectl kustomize references retained only where describing optional external inspection by users


aufi (approved):
LGTM

PR #295, opened 2026-06-17
Enhancement: Agentic Platform Controller

djzager:
## Summary

Introduces seven CRDs under `konveyor.io` and six controllers in a new `konveyor/agentic-controller` repository for managing AI agent workloads on Kubernetes.

**MVP scope**: SkillCard, SkillCollection, LLMProvider, Agent, AgentRun — five CRDs, four controllers enabling single-agent execution end to end.

**Post-MVP**: AgentPlaybook and AgentPlaybookRun for multi-phase orchestration with session continuity and stage-based handoff.

## Key Design Decisions

| Decision | Resolution |
|---|---|
| Orchestration | Custom controllers, no Tekton for MVP |
| Execution layer | Agent Sandbox (kubernetes-sigs/agent-sandbox), hard dependency |
| Skill packaging | skillimage Go library, three source types (OCI, git, inline) |
| Agent vs AgentRun | Agent = capability menu, AgentRun = specific selections |
| In-cluster registry | Zot for dev/vanilla k8s, OpenShift built-in auto-detect |
| Base image | Layered hierarchy with runtime-agnostic harness entrypoint |
| API group | `konveyor.io` |
| Repo | `konveyor/agentic-controller`, Go controller-runtime |

## Related

- Agent-driven migration (IDE side): `/enhancements/kai/agent-driven-migration`
- CRD architecture ADR: `dymurray/tackle2-ui` agent branch
- [skillimage](https://github.com/redhat-et/skillimage)
- [Agent Sandbox](https://github.com/kubernetes-sigs/agent-sandbox)

Status: `provisional`

## Summary by CodeRabbit

* **Documentation**
  * Added a new documentation page for the Agentic Platform Controller, describing its Kubernetes-native design, supported CRDs/controllers, and how skills are resolved and mounted for Agent Sandbox execution.
  * Documented the workspace persistence model, readiness/discovery and orchestration behavior, plus CRD examples and expected status flows.
  * Included security considerations, phased rollout and upgrade assumptions, and a complete unit, envtest, and end-to-end testing plan.

### DIFF
--- a/enhancements/agentic-platform-controller/README.md
+++ b/enhancements/agentic-platform-controller/README.md
@@ -1,0 +1,1138 @@
+---
+title: agentic-platform-controller
+authors:
+  - "@djzager"
+reviewers:
+  - TBD
+  - "@dymurray"
+approvers:
+  - TBD
+creation-date: 2026-06-17
+last-updated: 2026-06-22
+status: provisional
+see-also:
+  - "/enhancements/kai/agent-driven-migration/README.md"
+  - "/enhancements/agentic-platform-agent-images/README.md"
+  - "https://github.com/redhat-et/skillimage"
+  - "https://github.com/kubernetes-sigs/agent-sandbox"
+  - "https://agentskills.io"
+replaces: []
+superseded-by: []
+---
+
+# Agentic Platform Controller
+
+A Kubernetes-native controller for managing AI agent workloads. Introduces
+seven CRDs under the `konveyor.io` API group and six controllers in a new
+`konveyor/agentic-controller` repository.
+
+Agents are templates that declare available skills, LLM providers, a
+container image, a prompt, and typed parameters. AgentRuns are invocations
+that supply concrete values — model selections, parameter values,
+instructions, and additional env/envFrom passed through to the Sandbox.
+The controller validates, resolves skills to ImageVolumes, and creates
+Agent Sandbox workloads. It does not interpret parameter values or call
+any external API.
+
+Agent workspaces are ephemeral — git is the persistence layer. The
+harness (in the base image) clones a repo, the agent works, the harness
+commits and pushes. Everything durable lives in git.
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] User-facing documentation is created
+
+## Open Questions
+
+1. **Model role names**: Resolved. Model roles follow a small known
+   convention (`primary`, `efficient`, optional `planner`). Free-form
+   roles are allowed but unmapped by the harness — passed through as
+   env vars.
+
+2. **SkillCard content hashing**: Resolved. Content-addressable tags
+   by default. The controller hashes inline/git-sourced skill content
+   and uses SHA-based OCI tags to avoid rebuilding unchanged skills.
+
+3. **LLMProvider verification frequency**: Resolved. Verification on
+   create/update only. Periodic health checks are deferred.
+
+4. **Agent Sandbox API stability**: Resolved. Agent Sandbox is a hard
+   dependency. Pin the version and maintain a CI contract test. Bare
+   Pods are not a fallback.
+
+5. **Incremental push frequency**: Resolved. The harness pushes (git
+   push) after each successful commit batch. This is an event-based
+   strategy. The push frequency detail is owned by the
+   [agent-base-image-composition](/enhancements/agentic-platform-agent-images/README.md)
+   enhancement.
+
+## Summary
+
+The Konveyor agentic platform defines Kubernetes CRDs for composing and
+executing AI agent workloads. The design follows the Tekton Task/TaskRun
+pattern: an Agent declares what is available and what inputs it needs;
+an AgentRun supplies concrete values and triggers execution.
+
+The controller is domain-agnostic. It does not call Hub, Backstage, or
+any inventory system. The creator of the AgentRun (UI, CLI, Backstage
+plugin, CI pipeline) resolves application metadata and supplies it as
+parameter values and env/envFrom on the AgentRun. The controller passes
+these through without interpretation.
+
+Skills are sourced from OCI registries, git repositories, or inline
+content and resolved to OCI artifacts by the SkillCard controller.
+All skills are mounted at `/opt/skills/{name}/` via ImageVolumes —
+baked-in and mounted skills coexist in the same directory.
+
+Workspaces are ephemeral (EmptyDir). The harness clones, the agent
+works, the harness commits and pushes. No PVCs survive between runs.
+
+This enhancement covers the controller. Agent container image
+composition is covered by the
+[agent-base-image-composition](/enhancements/agentic-platform-agent-images/README.md)
+enhancement. IDE-side integration is covered by the
+[agent-driven-migration](/enhancements/kai/agent-driven-migration/README.md)
+enhancement.
+
+## Motivation
+
+### Current State
+
+The Konveyor POC for AI-powered migration (tackle2-addon-kai) uses Hub's
+addon framework to launch agent containers. This couples agent execution
+to Hub's task lifecycle, requires Hub API changes for every new resource
+type, and makes the agent platform dependent on Hub.
+
+The POC also uses pallet to sync skills from git repos at container
+startup, adding network dependency and startup latency. Skills are not
+versioned as artifacts, making reproducibility and supply chain security
+difficult.
+
+Workspaces in the addon model are EmptyDir volumes. The only durable
+output is a git branch pushed before pod termination. Session context
+(what the agent did, why, what to do next) is lost when the pod dies.
+
+### Goals
+
+1. **Kubernetes-native agent management**: CRDs, standard RBAC, GitOps,
+   controller-runtime patterns.
+
+2. **Domain-agnostic CRDs**: The controller does not encode assumptions
+   about Hub, Backstage, git hosting, or any specific use case.
+   Parameter values and env/envFrom are opaque to the controller.
+
+3. **Tekton-style param model**: Agent declares typed params, AgentRun
+   supplies values. The controller validates and passes through.
+
+4. **Skill lifecycle management**: Resolve skills from OCI, git, or
+   inline to versioned OCI artifacts via skillimage.
+
+5. **LLM provider verification**: Validate that LLM endpoints are
+   reachable by launching the actual base image container.
+
+6. **Git-as-persistence**: Ephemeral workspaces, durable git branches.
+   No PVCs between runs.
+
+### Non-Goals
+
+1. **Agent container image composition**: Covered by the
+   [agent-base-image-composition](/enhancements/agentic-platform-agent-images/README.md)
+   enhancement.
+
+2. **Tekton integration**: Deferred. The architecture does not preclude
+   it.
+
+3. **Skill authoring tools**: Domain of the skillimage project.
+
+4. **Hub or inventory integration in the controller**: The controller
+   does not call any external API. The UI resolves application metadata
+   before creating the AgentRun.
+
+5. **Interactive / human-in-the-loop execution**: ACP-based interactive
+   sessions are a future enhancement.
+
+6. **Session handoff schema**: The schema for `.konveyor/session.json`
+   and `.konveyor/handoff.md` is owned by the
+   [agent-base-image-composition](/enhancements/agentic-platform-agent-images/README.md)
+   enhancement.
+
+## Proposal
+
+### Architecture Overview
+
+```text
+                    ┌─────────────────────────────────────────────┐
+                    │          Kubernetes API Server               │
+                    │                                             │
+                    │  SkillCard    SkillCollection   LLMProvider │
+                    │  Agent       AgentPlaybook                  │
+                    │  AgentRun    AgentPlaybookRun               │
+                    └──────────────┬──────────────────────────────┘
+                                   │
+                    ┌──────────────▼──────────────────────────────┐
+                    │      konveyor/agentic-controller            │
+                    │                                             │
+                    │  SkillCard controller    (OCI/git/inline)   │
+                    │  SkillCollection controller (child CRs)     │
+                    │  LLMProvider controller  (verification)     │
+                    │  Agent controller        (budget, readiness)│
+                    │  AgentRun controller     (validate, Sandbox)│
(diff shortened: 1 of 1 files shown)


enhancements/agentic-platform-controller/README.md
```diff
@@ -0,0 +1,1051 @@
...
+    - name: review
+      phase: Pending
+```
+
+**Controller behavior**: Create AgentRuns per stage on the shared
+branch. Each stage reads `.konveyor/handoff.md` from the previous.
+
+### Skill Mounting
+
+All skills live at `/opt/skills/{skillName}/SKILL.md`. Built-in
```
savitharaghunathan, line 561: Would we support runtime discovery of skills? and from an ux perspective does a user pick the skill before kicking off AgentRun task?

Also, do we need to provide the ability to the user where they can create a skill based on their internal migration doc. they only bring the doc they trust, we create a migration skill for them and use that?

savitharaghunathan: I do notice that skill authoring is a non-goal. But, if we dont provide a mechanism to convert a migration doc to be a skill, we are assuming user possess skill creation knowledge. It feels like the rules creation problem to me.

enhancements/agentic-platform-controller/README.md
```diff
@@ -0,0 +1,1051 @@
...
+autonomous (no human-in-the-loop during execution). Explicitly
+rejected Agent Sandbox and ACP for their use case.
+
+**Opportunity:** If we build k8s-native agent execution well,
+fullsend could be a consumer when they add Kubernetes support
+(listed on their roadmap as "Later").
+
+### Positioning
+
+All three projects are Red Hat-adjacent, solving adjacent problems:
```
savitharaghunathan: nit: (vendor neutral) Red Hat-adjacent sounds we have only looked at Red Hat projects and nothing else.

savitharaghunathan (commented):
Thanks for this PR, @djzager. I have a few comments around ux and skills and I am happy to explain more of my thought process if needed.

JonahSussman (commented):
Like the enhancement, a few thoughts:

- I would like to see soemthing about observability here so that we can keep track of agent runs while they are executing.
- Skill/image compatibility is checked at runtime. So for example, a Maven skill on a Python image would fail. Perhaps a `requires` field on `SkillCard`, something like `requires: [maven, jdk]`, would be interesting to explore. We can also accept the runtime failure mode as well.

enhancements/agentic-platform-controller/README.md
```diff
@@ -0,0 +1,1051 @@
...
+
+1. **Model role names**: The AgentRun selects models with roles (e.g.
+   `primary`, `efficient`). Should recognized roles be fixed by
+   convention or free-form?
+
+2. **SkillCard content hashing**: When a SkillCard uses inline content,
+   should the controller use content-addressable tags to avoid
+   rebuilding unchanged skills?
+
+3. **LLMProvider verification frequency**: Should verification happen
```
savitharaghunathan: create/update sounds reasonable to me. I think if we support persistent agent sandbox pods it makes sense to verify the provider. IIUC, we are not doing anything persistent, right?

enhancements/agentic-platform-controller/README.md
```diff
@@ -0,0 +1,1051 @@
...
+   only on create/update, or periodically?
+
+4. **Agent Sandbox API stability**: Agent Sandbox is v1beta1. What is
+   the migration cost if the API changes before our v1?
+
+5. **Incremental push frequency**: How often should the harness push
```
savitharaghunathan: Can you clarify what 'push' refers to? git push or update hub with status?

enhancements/agentic-platform-controller/README.md
```diff
@@ -0,0 +1,1051 @@
...
+- **2026-06-22**: Revised: Tekton-style param model, controller
+  decoupled from Hub, env/envFrom passthrough, skill mounting
+  simplified to single /opt/skills/ directory
+
+## Drawbacks
+
+1. **New Go component**: Adds controller-runtime to the project's
+   technology surface.
+
+2. **Agent Sandbox dependency**: v1beta1 SIG project. Fallback to
```
fabianvf: Bare Pods are positioned two ways that don't quite reconcile: here (and in the Security section's "Agent Sandbox API changes" mitigation) as the fallback if Agent Sandbox's API shifts, but in **Alternatives** bare Pods are *rejected* precisely because Sandbox provides warm pools + network isolation. If those features are load-bearing enough to reject bare Pods, the fallback isn't equivalent; if they aren't, the hard dependency is harder to justify.

Could you pin this down — either (a) note the POC uses only a thin slice of Sandbox so a bare-Pod + NetworkPolicy fallback is genuinely viable (a one-paragraph sketch would settle it), or (b) commit to Sandbox as a hard dependency and replace "fallback is feasible" with the real mitigation (pinned version + a CI contract test)?

enhancements/agentic-platform-controller/README.md
```diff
@@ -0,0 +1,1051 @@
...
+DB).
+
+### Security, Risks, and Mitigations
+
+**Risk: Agent containers execute arbitrary code.**
+- *Mitigation*: Agent Sandbox provides network isolation. Non-root
+  execution. RBAC controls AgentRun creation. OpenShell can layer
+  on top for L7 egress policy (future).
+
+**Risk: Credentials in the Sandbox.**
```
fabianvf, line 1090: With no human-in-the-loop in the POC (now an explicit non-goal) and `git-write-creds` mounted via `envFrom` for the whole run, a misbehaving or prompt-injected agent holds live push credentials + egress for the entire session. "Output lands on a branch" bounds the *merge*, not what the agent does mid-run.

The Landscape section sketches a strong future story (OpenShell L7 egress, Kagenti AuthBridge/SPIFFE), but none of it is in the POC. As it stands the only model is direct-push to the source repo's `target_branch` — so the write credential effectively grants branch-push to the source for the whole run. Could the POC state a minimal credential-scoping contract: least-privilege per-repo tokens, short TTL, and ideally a fork-based push (agent pushes to a fork, not the source) to bound blast radius?

enhancements/agentic-platform-controller/README.md
```diff
@@ -0,0 +1,1051 @@
...
+
+**Rejected**: Git/inline sources need resolution to OCI artifacts.
+CRDs provide visibility, RBAC, status.
+
+## Infrastructure Needed
+
+- **New repository**: `konveyor/agentic-controller`
+- **Container image builds**: CI for controller + base images
+- **OCI registry access**: Push to `quay.io/konveyor/`
+- **Test cluster**: Agent Sandbox + ImageVolumes enabled
```
fabianvf, line 1507: ImageVolumes and `agents.x-k8s.io/v1beta1` both imply a minimum Kubernetes/OpenShift version (ImageVolume was feature-gated for several releases before going default-on). That's a hard prerequisite, not just a test-cluster detail. Worth a short "Platform Requirements" line: minimum k8s version (where ImageVolume is GA/default-on), minimum OpenShift version, and the Agent Sandbox version you're pinning to.

enhancements/agentic-platform-controller/README.md
```diff
@@ -0,0 +1,1051 @@
...
+      description: Git URL of the application source
+    - name: source_branch
+      type: string
+      default: main
+    - name: target_branch
+      type: string
+      description: Branch to push results to
+
+status:
+  contextBudget:
```
fabianvf: Two clarifications would help: (1) Is this budget a *gate* on Agent readiness or advisory? The controller-behavior note reads like a gate but doesn't say. (2) The count is `rulesContentTokens` only — it excludes `skill`-type cards, the `prompt`, and per-run `instructions`, and it's measured against the *smallest* model's window rather than the selected `primary`, so "0.75%" likely understates real utilization. If it's meant to gate, consider counting all always-loaded content against the run's primary model; if advisory, label it so.

enhancements/agentic-platform-controller/README.md
```diff
@@ -0,0 +1,1051 @@
...
+5. Push to the target repo
+6. Write `/.konveyor/results.json` (pod-local, read by controller)
+
+**Cross-stage continuity** in AgentPlaybooks: all stages share the
+same target branch. Each stage reads the previous stage's committed
+`.konveyor/handoff.md`.
+
+**Parallel agents**: different branches, same source. No contention.
+
+**Session context on the branch** (`.konveyor/` directory):
```
fabianvf, line 279: For AgentPlaybooks, `.konveyor/handoff.md` + `session.json` are the structured handoff channel between stages (alongside the code commits themselves) — stage N+1's correctness depends on stage N writing them in an agreed shape. `handoff.md` as free-form prose is fine, but `session.json` (model, tokens, timing, tool calls) probably wants a defined schema so the orchestrator and every agent image agree on it. Could you specify it here, or point to where it's owned (the #296 harness)?

fabianvf (commented):
Read through the v3 revision — I like the domain-agnostic / Tekton-style direction and git-as-persistence a lot. A handful of questions from the read-through, mostly around the POC security model plus a couple of internal-consistency points. All provisional-stage and non-blocking — feel free to punt any to implementation.

enhancements/agentic-platform-controller/README.md
```diff
@@ -0,0 +1,1051 @@
...
+commits and pushes. Everything durable lives in git.
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] User-facing documentation is created
+
+## Open Questions
```
fabianvf, line 48: For the `provisional → implementable` step it'd help to land positions on these rather than leave them open. Suggested defaults to react to: **OQ1** small known role convention (`primary`/`efficient`/optional `planner`), free-form allowed but unmapped; **OQ2** content-addressable tags by default (dedup + reproducibility, avoids rebuild churn); **OQ3** create/update-only for POC (a Pod-per-provider-per-interval health check is heavy — note the cost); **OQ4** pin + CI contract test, tied to the bare-Pods decision; **OQ5** event-based push (after each successful commit/tool-batch) over time-based — Drawback #5 already names the lost-work risk, and the flip side is that pushing too often makes noisy branch history, so the cadence is a real tradeoff worth stating.

enhancements/agentic-platform-controller/README.md
```diff
@@ -0,0 +1,1051 @@
...
+Trade-offs to evaluate include: k8s API conventions vs Hub
+conventions, privilege escalation via Hub's SA, and cross-plane
+data joining (agent resources in etcd, application data in Hub's
+DB).
+
+### Security, Risks, and Mitigations
+
+**Risk: Agent containers execute arbitrary code.**
+- *Mitigation*: Agent Sandbox provides network isolation. Non-root
+  execution. RBAC controls AgentRun creation. OpenShell can layer
```
fabianvf: OpenShell appears here and three more times in Landscape/Positioning as foundational shared infra ("Agent Sandbox and OpenShell are shared infrastructure at the bottom"), but it's never introduced or linked. A one-line intro + link on this first mention would help readers who don't know it.

enhancements/agentic-platform-controller/README.md
savitharaghunathan: this will be captured as an issue when @djzager creates the repo. This is resolved

savitharaghunathan (approved):

fabianvf (approved):
LGTM

enhancements/agentic-platform-controller/README.md
savitharaghunathan: tracked here https://github.com/konveyor/agentic-controller/issues/32

enhancements/agentic-platform-controller/README.md
```diff
@@ -0,0 +1,1138 @@
...
+`session/load` replays full conversation history on connect, so
+the UI can join a running session at any time without missing
+messages. For unattended runs, `GOOSE_MODE=auto` disables
+permission requests.
+
+When adding agent runtimes that do not implement ACP-over-HTTP
+(OpenCode, Claude, etc.), the harness adds a stdio-to-HTTP bridge.
+The external interface is identical regardless of runtime.
+
+See [ADR 0002: ACP Transport and Agent Observability](docs/adr/0002-acp-transport-and-observability.md)
```
jwmatthews: FYI, this is a broken link.

djzager: fixed

enhancements/agentic-platform-controller/README.md
```diff
@@ -0,0 +1,1471 @@
...
+    - type: Succeeded
+      status: "Unknown"
+      reason: Running
+```
+
+**Controller behavior**:
+
+1. Validate params match Agent's declarations
+2. Resolve skills → OCI image refs (from SkillCard status)
+3. Resolve gateway → credential Secret, set `KONVEYOR_LLM_ENDPOINT`, `KONVEYOR_LLM_MODEL`, `KONVEYOR_LLM_API_KEY`
```
ibolton336: One gap in the new env contract: `KONVEYOR_LLM_ENDPOINT/MODEL/API_KEY` drops the provider type. Pre-OpenShell the endpoint is the raw provider API, so the runtime still has to know which dialect to speak — the goose harness hard-requires `KONVEYOR_MODEL_PRIMARY_PROVIDER` to set `GOOSE_PROVIDER` and pick the right key var (`ANTHROPIC_API_KEY` vs `OPENAI_API_KEY` vs …). Today that value is inferred from the LLMProvider CR name (`anthropic`, `aws-bedrock`), which `anthropic-sonnet`-style Gateway names break.

Should Gateway grow a `spec.provider` field (runtime provider id) injected as `KONVEYOR_LLM_PROVIDER`? It's exactly the pre-OpenShell shim field — it deletes cleanly when real gateways land and everything speaks one dialect.

djzager: added

jortel (approved):

enhancements/agentic-platform-controller/README.md
```diff
@@ -0,0 +1,1482 @@
...
+  name: java-migration-agent-standalone
+spec:
+  image: quay.io/konveyor/agent-java-goose:latest
+  prompt: "You are a Java migration specialist..."
+  gateways:
+    - ref: anthropic-sonnet
+  skillCollections:
+    - ref: java-migration-skills
+  params:
+    - name: source_url
```
jortel, line 916: How is the kind-of repository determined: git or svn?

How are credentials specified?

PR #298, opened 2026-07-15
Added enhancement doc for SC conversion

stillalearner:
Added Enhancement Doc for adding support for Intra Cluster Storage Class Conversion feature in Crane.

## Summary by CodeRabbit

* **Documentation**
  * Updated the `storageclass-conversion` guide with a simplified, workflow-focused approach for intra-cluster StorageClass conversion.
  * Added a release checklist, open decisions, and clearer user stories for Deployment and StatefulSet use cases.
  * Described the revised `transfer-pvc` same-cluster workflow, including how workload PVC references are updated (including `volumeClaimTemplates`) and related cleanup considerations.
  * Refined the test plan, upgrade/downgrade strategy, and documented security/risk mitigations, limitations, drawbacks/alternatives, and infrastructure requirements.


enhancements/crane-2.1/storageclass-conversion/README.md
```diff
@@ -0,0 +1,335 @@
...
+
+3. Should old PVCs be cleaned up automatically or left for the user?
+    * **Decision:** Old PVCs are labeled (`crane.konveyor.io/migrated-to`) but never deleted. The user verifies data integrity and deletes manually.
+
+## Summary
+
+Crane currently has no mechanism for converting PVCs from one StorageClass to another within the same cluster. The only existing option is the `--dest-storage-class` flag on `crane transfer-pvc`, which is designed for cross-cluster migration — not intra-cluster StorageClass changes.
+
+This enhancement adds a new **`crane convert-storage`** command that provides MTC-equivalent `StorageConversionPlan` functionality as a CLI tool. The command creates a new PVC with the target StorageClass, transfers data via rsync (reusing the pvc-transfer library), automatically swaps workload references (Deployments, StatefulSets, DaemonSets, CronJobs, Jobs, ReplicaSets), and preserves the old PVC for rollback safety.
```
aufi: This is a nice enhancement, however I'd like challenge the way how it is solved by adding a dedicated command `crane convert-storage`.

## Main problem from my side
Convert PVC storageClass with single command introduces an adhoc command doing too many things IMO (copy data and patch resources using the converted PVC).

### Story for user
Migrate data (always) with `transfer-pvc` command, if it is from one cluster to another cluster, or on same cluster from one storage class to another, that depends on the command flags.

## Suggestion
- split into smaller reusable steps for v0.11 release
- single command storageClass conversion flow can come later with orchestration

## How (for v0.11)
1. update/extend transfer-pvc command to support this PVC data copy within same cluster and different storageClass
2. resources update/patching volumes mount is not supported by crane yet, we need either document manual workload changes, provide scripts or add some patching capabilities to crane? (export/transform/apply possible, but not handy)

aufi: Output from discussion:
- preferably not introduce new command
- extend tranfer-pvc command with storageClass PVC conversion within the same cluster
- solve patching resources using the PVC in a transform plugin (part of export/transform/apply flow), so the output manifest will be applied to patch resources that used the PVC
- at least for now confirmed approach we're taking with crane does not have live resources patching capabilities.

enhancements/crane-2.1/storageclass-conversion/README.md
```diff
@@ -0,0 +1,223 @@
...
+#### Simple case: Deployment with one PVC
+
+```bash
+# 1. Quiesce workload
+kubectl scale deploy webapp --replicas=0 -n myapp
+
+# 2. Transfer data to new PVC with new StorageClass
+crane transfer-pvc \
+  --source-context mycluster --destination-context mycluster \
+  --pvc-name "mysql-data:mysql-data-new" \
```
aufi, line 117: I'm not sure about the `--pvc-name "source-name:target-name"` syntax, but this might be updated later according to others opinion.

enhancements/crane-2.1/storageclass-conversion/README.md
```diff
@@ -0,0 +1,223 @@
...
+* [#659](https://github.com/migtools/crane/issues/659) — Handle StatefulSet volumeClaimTemplate conversion
+* [#660](https://github.com/migtools/crane/issues/660) — Support non-admin users for StorageClass conversion
+* [#661](https://github.com/migtools/crane/issues/661) — E2E test coverage for StorageClass conversion
+
+## Drawbacks
+
+* **Multiple manual steps:** The user must run `transfer-pvc` per PVC, then the export/transform/apply pipeline, then `kubectl apply`. This is by design (composability and reviewability) but requires more user effort than a single-command approach.
+* **StatefulSet requires manual delete+recreate:** Since `kubectl apply` cannot change immutable `volumeClaimTemplates`, the user must delete and recreate the StatefulSet manually. The generated manifest shows the correct target state.
+* **No automatic quiesce:** The user must manually scale down workloads before transfer to prevent data loss from writes during rsync.
+* **`pvc-rename-map` does not update StorageClass in templates:** For StatefulSets, new replicas scaled up after conversion would use the old StorageClass from the template unless a future `storage-class-map` flag is added.
```
aufi, line 228: Could we add it as part of this task?

aufi (approved):
Left few minor comments, overall looks good to me. Thanks for the enhancement Sachin!

enhancements/crane-2.1/storageclass-conversion/README.md
stillalearner: ya we can discuss on this separately i think, this is an already existing flag.

enhancements/crane-2.1/storageclass-conversion/README.md
stillalearner: this is actually only for Statefulsets workload updation. 

Statefulsets anyway require manual delete plus recreated. So i don tthink this flag is much of value here.

We can just mention in the docs, that for statefulsets while recreating, update the SC value to the new one in the YAML

aufi: Makes sense, thanks for clarification.

rromannissen (approved):
LGTM!

enhancements/crane-2.1/storageclass-conversion/README.md
```diff
@@ -0,0 +1,223 @@
...
+Since Kubernetes PVC `spec.storageClassName` is immutable after creation, converting requires: creating a new PVC with the target StorageClass, copying data, and updating workload references — a multi-step process that is error-prone when done manually.
+
+### Goals
+
+* **No new commands:** Extend existing `transfer-pvc` and leverage existing `pvc-rename-map` in transform
+* **Composable pipeline:** Data transfer and manifest generation are separate steps that the user composes
+* **Non-destructive:** Crane generates updated manifests on disk — the user reviews and applies
+* **No live workload patching:** Crane does not modify running workloads directly
+* **Data integrity:** Transfer data via rsync with optional checksum verification
+* **OCP and K8s:** Work on both OpenShift (Route endpoint) and vanilla Kubernetes (Ingress endpoint)
```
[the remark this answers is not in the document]
stillalearner: not needed.

enhancements/crane-2.1/storageclass-conversion/README.md
```diff
@@ -0,0 +1,223 @@
...
+#### Part 2: Workload manifest updates via existing `pvc-rename-map`
+
+The KubernetesPlugin in crane-lib already supports `pvc-rename-map` which generates JSONPatch operations to rename PVC references. It covers all workload types:
+
+* Deployments, DaemonSets, ReplicaSets, ReplicationControllers, Jobs, CronJobs, Pods — volume claim name references
+* StatefulSets — both pod spec volumes and `volumeClaimTemplates` names
+
+No plugin changes needed. The user passes the rename mapping via `--optional-flags` (or the future parametrized stage flag when the plugins enhancement lands).
+
+Note: `pvc-rename-map` renames the volumeClaimTemplate name but does not update its `storageClassName`. This is acceptable because the new PVC already has the correct StorageClass from `transfer-pvc`. However, for StatefulSets, new replicas scaled up in the future would use the old StorageClass from the template. A `storage-class-map` optional flag could be added to KubernetesPlugin in a future iteration to address this.
```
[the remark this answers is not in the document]
stillalearner: yes it will be mentioned in docs.

enhancements/crane-2.1/storageclass-conversion/README.md
```diff
@@ -0,0 +1,223 @@
...
+2. **Fix cert secret naming for same-namespace** — when source and destination are in the same namespace, copy the server's TLS cert secret under the name the client expects (different PVC names produce different secret names)
+3. **Split pod labels for same-namespace** — add a `role` label to distinguish server and client pods, so the log reader finds the correct pod
+4. **Dual garbage collection** — clean up both server-side and client-side resources when using split labels
+
+#### Part 2: Workload manifest updates via existing `pvc-rename-map`
+
+The KubernetesPlugin in crane-lib already supports `pvc-rename-map` which generates JSONPatch operations to rename PVC references. It covers all workload types:
+
+* Deployments, DaemonSets, ReplicaSets, ReplicationControllers, Jobs, CronJobs, Pods — volume claim name references
+* StatefulSets — both pod spec volumes and `volumeClaimTemplates` names
```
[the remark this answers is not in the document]
stillalearner: not needed.

istein1:
## Implementation suggestion: `storageClassName` in StatefulSet `volumeClaimTemplates`

The enhancement acknowledges that `pvc-rename-map` renames the template name but does not update its `storageClassName`, deferring a `storage-class-map` flag to a future iteration.

I'd suggest bringing this into scope for this enhancement. Without it, the StatefulSet conversion story is incomplete in a way that can silently surprise users: after completing the full migration flow, if the user scales up the StatefulSet, new replicas provision PVCs with the **old** StorageClass. The user believes the migration is done, but the cluster drifts back to the storage class they were migrating away from.

Two options to consider:

1. **Add a `storage-class-map` optional flag to KubernetesPlugin** alongside the existing `pvc-rename-map` — this would generate a JSONPatch to update `storageClassName` in `volumeClaimTemplates`. This keeps the composable pipeline model and is a small addition to the plugin.

2. **At minimum, add a post-pipeline validation warning** — after `crane apply` generates the output manifests, warn the user if any `volumeClaimTemplate` still references a StorageClass that was the source of a conversion. This doesn't fix the gap but makes it visible.

Without one of these, a user following the documented StatefulSet workflow would end up with a partially migrated StatefulSet that re-introduces the old StorageClass on scale-up.

istein1:
## Documentation suggestion: prominent warnings for destructive steps

Two steps in the documented workflows carry significant data-loss risk if performed incorrectly. Suggesting the docs team call these out with explicit warning boxes in the user-facing documentation:

### 1. StatefulSet delete must use `--cascade=orphan`

The workflow correctly shows `kubectl delete sts redis --cascade=orphan`, but if a user omits `--cascade=orphan`, a plain `kubectl delete sts` deletes the pods **and** their PVCs — destroying the data that was just migrated. This should have a prominent warning callout (e.g., an admonition/alert box) explaining:
- Why `--cascade=orphan` is required (preserves PVCs and pods)
- What happens if omitted (PVC deletion, data loss)

### 2. Workload must be quiesced before transfer

The workflow shows `kubectl scale deploy webapp --replicas=0` as step 1, but there's no warning about what happens if skipped. The docs should include a data-loss warning explaining that running `transfer-pvc` while the workload is still writing to the PVC can result in inconsistent data on the destination PVC, since rsync is not a point-in-time snapshot.

stillalearner:
> ## Implementation suggestion: `storageClassName` in StatefulSet `volumeClaimTemplates`
> The enhancement acknowledges that `pvc-rename-map` renames the template name but does not update its `storageClassName`, deferring a `storage-class-map` flag to a future iteration.
> 
> I'd suggest bringing this into scope for this enhancement. Without it, the StatefulSet conversion story is incomplete in a way that can silently surprise users: after completing the full migration flow, if the user scales up the StatefulSet, new replicas provision PVCs with the **old** StorageClass. The user believes the migration is done, but the cluster drifts back to the storage class they were migrating away from.
> 
> Two options to consider:
> 
> 1. **Add a `storage-class-map` optional flag to KubernetesPlugin** alongside the existing `pvc-rename-map` — this would generate a JSONPatch to update `storageClassName` in `volumeClaimTemplates`. This keeps the composable pipeline model and is a small addition to the plugin.
> 2. **At minimum, add a post-pipeline validation warning** — after `crane apply` generates the output manifests, warn the user if any `volumeClaimTemplate` still references a StorageClass that was the source of a conversion. This doesn't fix the gap but makes it visible.
> 
> Without one of these, a user following the documented StatefulSet workflow would end up with a partially migrated StatefulSet that re-introduces the old StorageClass on scale-up.


As discussed with @aufi earlier, 

This updation of Storageclass can be just mentioned in the Docs of storageclass conversion.
The extra flag is not needed at all , for just the statefulset updation.

istein1:
Thanks @stillalearner . Approving. 
Just please follow up on: "This updation of Storageclass can be just mentioned in the Docs of storageclass conversion."

istein1 (approved):

PR #300, opened 2026-07-21
Add crane BuildConfig to ShipWright conversion

aufi:
## Summary

  - New enhancement proposal for crane-2.1: a crane transform plugin that converts OpenShift BuildConfig resources to Shipwright Build CRs
  - Replaces the existing crane convert PoC approach with a GitOps-friendly, offline plugin integrated into crane's export → transform → apply pipeline
  - Covers Docker and S2I strategy mapping, source/output field conversion, and unsupported field warnings
  - Depends on the plugin API extension (NewResources in PluginResponse) as a prerequisite

## Motivation

  BuildConfig is one of the most complex OpenShift-specific resources to migrate. This proposal formalizes the conversion path using crane's standard transformation architecture rather than the direct API-based crane convert approach.

## Scope

  - 4-phase implementation plan: plugin API extension → BuildConfig plugin → crane workflow updates → documentation
  - Field mapping tables for strategies, sources, outputs, and known gaps
  - End-to-end conversion example (BuildConfig YAML → Shipwright Build YAML)

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive proposal document for a new crane transform plugin to convert OpenShift BuildConfig resources into Shipwright Build resources.
  * Documents deterministic/offline transformation behavior, including whiteout of source BuildConfigs and generation of required prerequisite resources.
  * Covers field mappings for build inputs/outputs, layered image reference resolution, handling of unsupported features, failure behavior, example conversion, upgrade considerations, and a test plan.

### DIFF
--- a/enhancements/crane-2.1/buildconfig-to-shipwright/README.md
+++ b/enhancements/crane-2.1/buildconfig-to-shipwright/README.md
@@ -37,11 +37,11 @@ OpenShift `BuildConfig` (`build.openshift.io/v1`) is a platform-specific CI/CD r
 
 2. **Shipwright strategy version pinning:** Should the plugin hardcode ClusterBuildStrategy names (`buildah`, `source-to-image`) or make them configurable via flags? The existing PoC hardcodes them, but cluster-specific ClusterBuildStrategy names may vary.
 
-3. **BuildRun generation:** Should the plugin also generate a BuildRun CR (the Shipwright equivalent of triggering a build), or leave that to the user? Recommendation: leave it on user, out of scope for this version.
+3. **BuildRun generation:** Should the plugin also generate a BuildRun CR (the Shipwright equivalent of triggering a build), or leave that to the user? Recommendation: generate a BuildRun CR referencing the Build and ServiceAccount.
 
 ## Summary
 
-Organizations migrating to newer versions of OpenShift to vanilla Kubernetesmight need to convert their `BuildConfig` resources to a portable, cloud-native alternative. Shipwright is a CNCF Sandbox project that provides a Kubernetes-native build framework and is the natural successor for OpenShift's build system.
+Organizations migrating to newer versions of OpenShift to vanilla Kubernetes might need to convert their `BuildConfig` resources to a portable, cloud-native alternative. Shipwright is a CNCF Sandbox project that provides a Kubernetes-native build framework and is the natural successor for OpenShift's build system.
 
 This enhancement adds a new crane transform plugin that performs offline, deterministic conversion of BuildConfig resources to Shipwright Build CRs. The plugin integrates into crane's multi-stage transformation pipeline, producing reviewable YAML artifacts with full whiteout and patch trail — no live cluster connectivity is required during transformation.
 
@@ -116,15 +116,15 @@ The existing PoC uses a direct API approach: it queries the live source cluster,
 
 #### Plugin System Prerequisite
 
-This plugin requires the crane plugin API extension that enables generating new resources (adding `NewResources []unstructured.Unstructured` to `PluginResponse`). This extension is 100% backward compatible via the `omitempty` JSON tag and is detailed in a [separate plan](https://github.com/aufi/move-crane/blob/main/drafts/plugin-update-new-resource-plan.md). Implementation touches:
+This plugin requires the crane plugin API extension that enables generating new resources (adding `NewResources []unstructured.Unstructured` to `PluginResponse`). This extension is backward compatible on the JSON wire format via the `omitempty` tag (existing plugins that omit `NewResources` continue to work unchanged). Note that adding the field to the Go struct will break any downstream code using unkeyed `PluginResponse{...}` literals — those must be updated to use keyed fields. Details in a [separate plan](https://github.com/aufi/move-crane/blob/main/drafts/plugin-update-new-resource-plan.md). Implementation touches:
 
 - **crane-lib:** `transform/plugin.go` (PluginResponse), `transform/runner.go` (RunnerResponse, Runner.Run)
-- **crane:** `Orchestrator` (artifact creation for new resources) and potentialy `Writer` (plugin dirs update).
+- **crane:** `Orchestrator` (artifact creation for new resources) and potentially `Writer` (plugin dirs update).
 
 
 #### Integration with Crane Workflow
 
-```
+```text
 crane export -n myapp
     ↓
     export/resources/
@@ -157,9 +157,10 @@ The plugin processes each resource in the stage input:
 
 1. **Filter:** Skip non-BuildConfig resources (return empty response)
 2. **Whiteout** the original BuildConfig (mark for deletion via `IsWhiteOut: true`)
-3. **Generate** a new Shipwright Build CR with mapped fields
+3. **Generate** a new Shipwright Build CR with mapped fields (unsupported strategies fail the conversion with a clear error and warning)
 4. **Optionally generate** a ServiceAccount if pull/push secrets are referenced
-5. Return the new resource(s) via `NewResources` in PluginResponse
+5. **Optionally generate** a BuildRun CR referencing the Build and the ServiceAccount (via `BuildRunSpec.serviceAccount`)
+6. Return the new resource(s) via `NewResources` in PluginResponse
 
 #### Field Mapping
 
@@ -177,7 +178,7 @@ The plugin processes each resource in the stage input:
 | BuildConfig Source | Shipwright Source | Notes |
 |-------------------|------------------|-------|
 | `git.uri` + `git.ref` | `source.type: Git`, `source.git.url` + `revision` | Direct mapping |
-| `git.httpProxy/httpsProxy` | Env vars `HTTP_PROXY`, `HTTPS_PROXY` | Injected as build env |
+| `git.httpProxy/httpsProxy` | _(not directly supported)_ | Shipwright's Git clone runs in a separate container; proxy must be configured via `GIT_CONTAINER_TEMPLATE` at the cluster level. Warning emitted. |
 | `sourceSecret` | `source.git.cloneSecret` | Direct mapping |
 | `contextDir` | `source.contextDir` | Direct mapping |
 | `binary` | `source.type: Local` | Requires manual `shp build upload` after apply |
@@ -277,7 +278,7 @@ spec:
 
 #### Plugin Repository Structure (draft)
 
-```
+```text
 crane-plugin-buildconfig-to-shipwright/
 ├── buildconfig/converter.go         # BuildConfig → Shipwright Build mapping
 ├── buildconfig/converter_test.go    # Unit tests for each strategy/source type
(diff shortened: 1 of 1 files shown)


enhancements/crane-2.1/buildconfig-to-shipwright/README.md
```diff
@@ -0,0 +1,378 @@
...
+#### Field Mapping
+
+**Strategy mapping:**
+
+| BuildConfig Strategy | Shipwright ClusterBuildStrategy | Notes |
+|---------------------|-------------------------------|-------|
+| `dockerStrategy` | `buildah` | Dockerfile path, build args, base image mapped as paramValues |
+| `sourceStrategy` (S2I) | `source-to-image` | Builder image mapped as paramValue, env vars copied directly |
+| `customStrategy` | _(no conversion)_ | Warning emitted, resource passed through |
+| `jenkinsPipelineStrategy` | _(out of scope)_ | Warning emitted, resource passed through |
```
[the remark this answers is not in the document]
aufi: updated the flow steps above, closing.

enhancements/crane-2.1/buildconfig-to-shipwright/README.md
```diff
@@ -0,0 +1,378 @@
...
+- `--block-registries` — comma-separated blocked registries
+- `--default-build-strategy` — override default ClusterBuildStrategy name
+
+### Security, Risks, and Mitigations
+
+| Risk | Impact | Mitigation |
+|------|--------|------------|
+| **Shipwright not installed on target cluster** | Build CRs fail on `kubectl apply` | Validate command should check it, document Shipwright installation in user guide |
+| **ImageStreamTag references can't be resolved offline** | Output image URL incomplete | Use exported ImageStream data; fallback to placeholder with `crane.konveyor.io/unresolved-ref` annotation |
+| **BuildConfig uses unsupported features** | Incomplete conversion | Emit warnings per unsupported field; annotate output CR with `crane.konveyor.io/warnings` listing each gap |
```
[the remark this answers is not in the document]
aufi: updated with failing in such case (not found registry)

enhancements/crane-2.1/buildconfig-to-shipwright/README.md
```diff
@@ -0,0 +1,379 @@
...
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] User-facing documentation is created
+
+## Open Questions
+
+1. **ImageStream resolution strategy:** When a BuildConfig references an ImageStreamTag for its base image or output, the plugin needs to resolve this to a concrete registry URL. Options:
```
rromannissen: Can we have the three of them? I'm thinking of implementing a layered fallthrough approach (a → b → c).

1. Step 1: Check co-exported YAML (Option a). If the export directory contains the matching ImageStream file, parse its status.dockerImageRepository or spec.tags to extract the exact registry URL. This is the cleanest, zero-config approach for local discovery.

2. Step 2: Check explicit mapping flag (Option b). If a user passes a --registry-mapping flag or config file, override Step 1. This is critical for cross-cluster migrations where the destination cluster uses an entirely different registry (e.g., moving from an old OCP internal registry to Quay.io or AWS ECR).

3. Step 3: Fallback to internal registry pattern (Option c). If neither applies, fall back to constructing the internal OpenShift registry string (image-registry.openshift-image-registry.svc:5000/<namespace>/<name>:<tag>).

enhancements/crane-2.1/buildconfig-to-shipwright/README.md
```diff
@@ -0,0 +1,379 @@
...
+   - (a) Read from co-exported ImageStream YAML in the same export directory (preferred)
+   - (b) Accept a mapping file as plugin flag
+   - (c) Use a default OpenShift internal registry URL pattern as fallback
+
+2. **Shipwright strategy version pinning:** Should the plugin hardcode ClusterBuildStrategy names (`buildah`, `source-to-image`) or make them configurable via flags? The existing PoC hardcodes them, but cluster-specific ClusterBuildStrategy names may vary.
```
rromannissen: I think having flags is the safest choice, as cluster administrators can customize strategies, and vanilla Kubernetes environments might not use the same strategy naming conventions as Red Hat's downstream operator (Builds for Red Hat OpenShift). We could default to standard upstream names (buildah for Docker strategy, source-to-image for S2I) but have flags to allow custom overrides.

enhancements/crane-2.1/buildconfig-to-shipwright/README.md
```diff
@@ -0,0 +1,379 @@
...
+
+3. **BuildRun generation:** Should the plugin also generate a BuildRun CR (the Shipwright equivalent of triggering a build), or leave that to the user? Recommendation: generate a BuildRun CR referencing the Build and ServiceAccount.
```
rromannissen: Creating the BuildRun instance would trigger a build, which could lead to undesired effects. I think it would be dangerous to do that, especially at scale. It should be up to the Ops Engineer running the migration to decide when to trigger a build, which is not necessarily tied up to the migration itself, as the lifecycle of the application images (the BuildRuns) can be independent from the resource definition (creating the Build instance) itself and managed by an external CI/CD system. I would avoid that and stick to only creating Builds.

rromannissen (changes requested):
Overall, looks good to me. I've provided my take on the open questions, and I would expect those to be incorporated in the design once the discussion is finished. Besides that, the rest looks great!

enhancements/crane-2.1/buildconfig-to-shipwright/README.md
aufi: Makes sense, updated.

enhancements/crane-2.1/buildconfig-to-shipwright/README.md
aufi: Makes sense, we'll allow override these strategies with optional flag. Updated.

enhancements/crane-2.1/buildconfig-to-shipwright/README.md
aufi: Thanks for feedback, shortly, we can. Updated.

istein1:
Well-structured enhancement proposal!
 
 **Outstanding Issues**

  1. Conversion logic contradicts the strategy mapping table (correctness)

  Steps 2-3 in "Conversion Logic" still read as sequential:

  ▎ 2. Whiteout the original BuildConfig
  ▎ 3. Generate a new Shipwright Build CR (unsupported strategies fail...)

  But the strategy table says customStrategy and jenkinsPipelineStrategy are "passed
  through" with a warning. If whiteout happens unconditionally at step 2, unsupported
  BuildConfigs get deleted with no replacement. The flow should be: check strategy support
  first, then whiteout only on successful conversion. The author said this was addressed,
  but the current text at HEAD still shows the same ordering.

  2. ImageStream fallback contradicts risk mitigation (consistency)

  Open Question 1, option 3 still says:

  ▎ Fallback — construct internal OpenShift registry URL 
  ▎ (image-registry.openshift-image-registry.svc:5000/...) with a warning

  But the risk table says:

  ▎ if resolution fails, fail the conversion with a clear error rather than emitting a 
  ▎ placeholder registry URL

  These directly contradict each other. CodeRabbit flagged this too (on the latest
  commit). Pick one — I'd recommend failing the conversion, consistent with the risk
  mitigation.

  3. ServiceAccount generation has no consumer (design gap)

  Step 4: "Optionally generate a ServiceAccount if pull/push secrets are referenced."
  Shipwright's serviceAccount field lives on BuildRunSpec, not on Build. Since the
  proposal explicitly doesn't generate BuildRuns (Open Question 3), this ServiceAccount
  would be emitted into NewResources but never referenced by anything in the output.
  Either drop it, or document that it's scaffolded for the user to wire into their own
  BuildRun definitions.

  **Suggestions (non-blocking)**

  4. --default-build-strategy covers only one strategy. There are two distinct mappings
  (Docker→buildah, S2I→source-to-image). A single flag can't override both. Consider
  --docker-build-strategy and --s2i-build-strategy, or a mapping format like
  --strategy-mapping docker=my-buildah,s2i=my-s2i.

  5. --registry-mapping format unspecified. What's the input format? A CLI flag with
  old=new pairs? A file path to a YAML/JSON map? This matters for usability with
  large-scale migrations involving many registries.

  6. Plugin stdin contract vs. --export-dir. Open Question 1 option 2 requires reading
  sibling files from the export directory, but plugins receive one resource at a time via
  stdin. This breaks the clean stdin/stdout plugin contract. Worth calling out explicitly
  as a design constraint — it means this plugin needs filesystem access beyond what the
  standard plugin interface provides.

  7. spec.resources (resource limits) not addressed. BuildConfig can specify CPU/memory
  limits for builds. Shipwright supports resource overrides. This should at least appear
  in the unsupported fields table if intentionally omitted.

  8. Binary source mapping is incomplete. The table notes binary → Local "requires manual
  shp build upload after apply." This effectively means binary builds aren't fully
  automated. Worth highlighting more prominently — maybe in Drawbacks or known gaps.

  9. Conversion example: runtime-stage-from param. Verify this is the correct Shipwright
  buildah parameter for overriding the base image from dockerStrategy.from.
  runtime-stage-from is typically for the final stage in multi-stage Dockerfiles, not a
  general FROM override.

  10. last-updated metadata still shows 2026-07-21 despite 3 follow-up commits through
  July 24.

aufi:
Thanks for feedback!

> **Outstanding Issues**

Those were mostly caused by non-updated enhancement body after open questions were decided, updated now.
 
> **Suggestions (non-blocking)**
> 
> 4. --default-build-strategy covers only one strategy. There are two distinct mappings
>    (Docker→buildah, S2I→source-to-image). A single flag can't override both. Consider
>    --docker-build-strategy and --s2i-build-strategy, or a mapping format like
>    --strategy-mapping docker=my-buildah,s2i=my-s2i.

:+1: Added.

> 5. --registry-mapping format unspecified. What's the input format? A CLI flag with
>    old=new pairs? A file path to a YAML/JSON map? This matters for usability with
>    large-scale migrations involving many registries.

Good point - the exact format (inline pairs vs. file path) will be defined during implementation in the plugin repo. For large-scale migrations a file-based input is likely.

> 6. Plugin stdin contract vs. --export-dir. Open Question 1 option 2 requires reading
>    sibling files from the export directory, but plugins receive one resource at a time via
>    stdin. This breaks the clean stdin/stdout plugin contract. Worth calling out explicitly
>    as a design constraint — it means this plugin needs filesystem access beyond what the
>    standard plugin interface provides.

Good for for clarification, plugin cannot read other files, relevant resources from export (or just their existence) needs to be passed via optiona flags to plugin, updated. (this might be more clarified during implementation)

> 7. spec.resources (resource limits) not addressed. BuildConfig can specify CPU/memory
>    limits for builds. Shipwright supports resource overrides. This should at least appear
>    in the unsupported fields table if intentionally omitted.

For now, adding it to unsupported, however it might be updated along with implemetation and testing and clarification of typical usecases.

> 8. Binary source mapping is incomplete. The table notes binary → Local "requires manual
>    shp build upload after apply." This effectively means binary builds aren't fully
>    automated. Worth highlighting more prominently — maybe in Drawbacks or known gaps.

IIUC this is already documented in the source mapping table. Binary builds are inherently manual - no tooling can fully automate them offline.

> 9. Conversion example: runtime-stage-from param. Verify this is the correct Shipwright
>    buildah parameter for overriding the base image from dockerStrategy.from.
>    runtime-stage-from is typically for the final stage in multi-stage Dockerfiles, not a
>    general FROM override.

Thanks, updated the field name.

> 10. last-updated metadata still shows 2026-07-21 despite 3 follow-up commits through
>     July 24.

Updated.

enhancements/crane-2.1/buildconfig-to-shipwright/README.md
```diff
@@ -0,0 +1,378 @@
...
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] User-facing documentation is created
+
+## Open Questions
+
+1. **ImageStream resolution strategy:** When a BuildConfig references an ImageStreamTag for its base image or output, the plugin needs to resolve this to a concrete registry URL. Decision: use a layered fallthrough approach (1 → 2 → 3):
+   1. **Explicit mapping** — if user provides a `--registry-mapping` flag, use it (highest priority, critical for cross-cluster migrations)
+   2. **Co-exported ImageStream YAML** — read from sibling files in the export directory (requires `--export-dir` flag since plugins receive one resource at a time via stdin)
```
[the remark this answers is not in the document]
aufi: re-check with current PR state, it should have been removed

istein1 (approved):

enhancements/crane-2.1/buildconfig-to-shipwright/README.md
psrvere: Agree the plugin should never trigger builds. 

But could we still emit a template BuildRun that can't be applied accidentally. 
It can be written as BuildRun_<name>.yaml.template. kubectl apply -f <dir> only picks up .yaml/.yml/.json, so the file is inert until the user deliberately renames it. 

A build only executes when the BuildRun object is created in the cluster. This could also be gated behind an opt-in flag like --generate-buildrun-template.

Why do this? several BuildConfig fields have no home on Build and only map to BuildRunSpec:

- spec.serviceAccount → BuildRunSpec.serviceAccount — also the only consumer for the ServiceAccount the plugin generates; without a BuildRun it's emitted but referenced by nothing
- spec.resources → BuildRun resource overrides (currently listed as unsupported)
- spec.completionDeadlineSeconds → BuildRunSpec.timeout

With "leave it to the user," these values are silently dropped and must be reconstructed by hand. A template preserves them in reviewable YAML, while when to run the build stays entirely with the user or their CI/CD system.

